PR #1, opened 2024-02-15
Adds custom numeric_limits template for IEC data types

MartinMelikMerkumians:
As a preparation, CIEC_LREAL's cast to TForteDFloat has been made explicit, to stop the compiler to make various casts in order to fulfill certain tests. As an additional cleanup, the dangerous function tmpnam in CSV_WRITER test has been replaced with the recommended function mkstemp


azoitl:
looks good to me.

PR #4, opened 2024-02-16
Refactor BSD socket interface receive functions

diplfranzhoepfinger:
- Handle different error cases in receiveDataFromTCP and receiveDataFromUDP
- Log specific error messages for disconnection scenarios
- Return 0 when the connection is closed by the peer
Bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=582098


diplfranzhoepfinger:
#18

PR #6, opened 2024-02-18
Fix modbus build for Windows target

kumajaya:


kumajaya:
Sorry, incorrect base branch.

PR #11, opened 2024-02-18
Fix when compiling using uClibc-ng libc

kumajaya:
Resolved issue when linking to uClibc-ng libc, gcc could not find references to `__stack_chk_guard', `__stack_chk_fail', and `_dl_auxvt'

### DIFF
--- a/src/arch/posix/main.cpp
+++ b/src/arch/posix/main.cpp
@@ -24,6 +24,37 @@
 #include "../../modules/PLCnext/plcNextDeviceStatus.h"
 #endif
 
+#ifdef __UCLIBC__
+extern "C" {
+  void * __stack_chk_guard = 0;
+
+  void __stack_chk_guard_setup(void) {
+    unsigned char *p;
+    p = (unsigned char *) &__stack_chk_guard;
+
+    p[0] = 0;
+    p[1] = 0;
+    p[2] = '\n';
+    p[3] = 255;	// XXX: Random this
+  }
+
+  /*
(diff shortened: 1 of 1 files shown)


kumajaya:
References:
- https://github.com/crosstool-ng/crosstool-ng/issues/681#issuecomment-444005824
- https://stackoverflow.com/questions/27290086/gcc-canaries-undefined-reference-to-stack-chk-guard

Code references:
- https://github.com/eistec/contiki/blob/mulle-master/cpu/arm/k60/ssp.c
- https://github.com/wbx-github/uclibc-ng/blob/master/ldso/include/ldso.h

Still not sure if this is a uClibc bug or an incorrect build flag from my side. With -fstack-protector-all enabled, gcc will try to find libssp and libssp_nonshared which are not available in my version of the toolchain from https://toolchains.bootlin.com/

jwalt:
I don't think this is the right approach. This is clearly a compiler/libc issue that should be handled externally through cusom compiler/linker options. The proper solution would be to use an external `libssp` (as described in one of your references) and add it to your link options when using uclibc. In general it is always error-prone to use a different libc than it was built for. If your actual goal is static linking of FORTE, take a look at the upcoming https://github.com/eclipse-4diac/4diac-fbe which solves this properly (using musl libc). It is still work in progress, I recommend waiting for the first binary release.

kumajaya:
No problems so far using the bleeding-edge musl libc from https://toolchains.bootlin.com/ I'm closing this and leaving it as a reference only.

PR #18, opened 2024-02-22
Refactor BSD socket interface receive functions

franz-hoepfinger-4diac:
- Handle different error cases in receiveDataFromTCP and receiveDataFromUDP
- Log specific error messages for disconnection scenarios
- Return 0 when the connection is closed by the peer

### DIFF
--- a/src/arch/bsdsocketinterf.cpp
+++ b/src/arch/bsdsocketinterf.cpp
@@ -137,34 +137,6 @@ int CBSDSocketInterface::sendDataOnTCP(TSocketDescriptor paSockD, const char* pa
   return nRetVal;
 }
 
-int CBSDSocketInterface::handleError(int nRetVal, int err, const char* msg) {
-  // recv only sets errno if res is <= 0
-  if(nRetVal <= 0) {
-    switch(errno){
-      case EWOULDBLOCK:
-      case ENOENT: //caused by vfs
-        //connected = true;
-        break;
-      case ENOTCONN:
-      case EPIPE:
-      case ECONNRESET:
-      case ECONNREFUSED:
-      case ECONNABORTED:
-        //connected = false;
-        DEVLOG_ERROR("CBSDSocketInterface::receiveDataFrom%s  recv() Disconnected: nRetVal: %d, ERR: %d %s\n", msg, nRetVal, errno, strerror(errno));
-        return 0; //Connection closed by peer
-      default:
-        DEVLOG_ERROR("CBSDSocketInterface::receiveDataFrom%s recv() Unexpected: nRetVal: %d, ERR: %d %s\n", msg, nRetVal, errno, strerror(errno));
-        //connected = true;
-        break;
-    }
-  }
-  return nRetVal;
-}
-
-
-
-
 int CBSDSocketInterface::receiveDataFromTCP(TSocketDescriptor paSockD, char* paData,
     unsigned int paBufSize){
   int nRetVal;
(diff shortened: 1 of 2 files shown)


MartinMelikMerkumians (approved):

src/arch/bsdsocketinterf.cpp
```diff
@@ -147,6 +147,27 @@ int CBSDSocketInterface::receiveDataFromTCP(TSocketDescriptor paSockD, char* paD
...
+      case ECONNRESET:
+      case ECONNREFUSED:
+      case ECONNABORTED:
+        DEVLOG_ERROR("CBSDSocketInterface::receiveDataFromTCP recv() Disconnected: nRetVal: %d, ERR: %d %s\n", nRetVal, errno, strerror(errno));
+        return 0; //Connection closed by peer
+      default:
+        DEVLOG_ERROR("CBSDSocketInterface::receiveDataFromTCP recv() Unexpected: nRetVal: %d, ERR: %d %s\n", nRetVal, errno, strerror(errno));
+        //TODO: this is a "connected = true" case, for the moment we leave it open.
+        break;
+    }
```
azoitl: Can you extract that code into a method. Below we have a near 100% copy of that code. This makes it very hard to maintain.

azoitl (changes requested):

src/arch/bsdsocketinterf.cpp
diplfranzhoepfinger: Yes. Good idea.

diplfranzhoepfinger: done.

franz-hoepfinger-4diac: done

src/arch/bsdsocketinterf.h
```diff
@@ -33,6 +33,10 @@ class CBSDSocketInterface{
     static int sendDataOnUDP(TSocketDescriptor paSockD, TUDPDestAddr *paDestAddr, char* paData, unsigned int paSize);
     static int receiveDataFromUDP(TSocketDescriptor paSockD, char* paData, unsigned int paBufSize);
 
+
+    static int handleError(int nRetVal, int err, const char* msg);
```
azoitl: I don't think this method should be public please make it private

src/arch/bsdsocketinterf.cpp
```diff
@@ -137,6 +137,34 @@ int CBSDSocketInterface::sendDataOnTCP(TSocketDescriptor paSockD, const char* pa
   return nRetVal;
 }
 
+int CBSDSocketInterface::handleError(int nRetVal, int err, const char* msg) {
```
azoitl: As this is a private helper function I find it typically nicer to have them on the bottom of the file.

azoitl (changes requested):

azoitl (approved):

diplfranzhoepfinger:
there came in some Remarks from David: 

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/lwip.html#bsd-sockets-api

diplfranzhoepfinger:
espressif/esp-idf#13133

azoitl:
> [espressif/esp-idf#13133](https://github.com/espressif/esp-idf/issues/13133)

I'm now curious why we shouldn't clearly close the connection. From 4diac IDE side we definitely do that. The only problem that could arrise is that there is a timeout in 4diac IDE if the device is not responding with xxxms to close the socket on our side. But then we assume the device has died or the connection is broken (e.g., cable disconnected).  As other implementations are fine with how we are interacting with 4diac FORTE I still assume it is an esp32 problem.

diplfranzhoepfinger:
#18 was merged and to me it seems it works very good. 

we lost the Error Messages here, due to the Move from Gerrit to Github:

Bug: [https://bugs.eclipse.org/bugs/show_bug.cgi?id=582098](https://bugs.eclipse.org/bugs/show_bug.cgi?id=582098)

diplfranzhoepfinger:
this was the Error: 


it is solved now.

diplfranzhoepfinger:
i must add here another Log, so we see how the Behaviour is on a Close of Eclipse 4diac IDE. 

otherwise these knowledge gets lost.

diplfranzhoepfinger:
i must add here another Log, so we see how the Behaviour is on a Close of Eclipse 4diac IDE. 

otherwise these knowledge gets lost.

PR #21, opened 2024-02-24
Update project name in XML file

franz-hoepfinger-4diac:
Changed the project name in the XML file from "org.eclipse.4diac.forte" to "4diac-forte".

### DIFF
--- a/.project
+++ b/.project
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>org.eclipse.4diac.forte</name>
+	<name>4diac-forte</name>
 	<comment></comment>
 	<projects>
 	</projects>


azoitl:
Personally I would like to keep the old project name. @MartinMelikMerkumians @moniwe @bwiesmayr what do you think?

franz-hoepfinger-4diac:
> Personally I would like to keep the old project name.

in eclipse i experienced Problems when Folder Name and Project Name is not same.

azoitl:
> > Personally I would like to keep the old project name.
> 
> in eclipse i experienced Problems when Folder Name and Project Name is not same.

I have several examples on my machine which prove the opposite. So your problems need to come from something else.

moniwe:
in my Eclipse everything seems to be fine. Which problems did you have?

diplfranzhoepfinger:
> in my Eclipse everything seems to be fine. Which problems did you have?

at espressif eclipse plugin the hex files was lost. 

because the build script took the wrong one (path vs. Project name). 

must look after where i documented this issue.

moniwe:
i guess most 4diac contributor/commiter do not use this plugin? At least this explains why we do not have this problem. For Java developers this difference in project name and folder name is quite common i think, at least i am used to it from where i work. Changeing it would mean to have a different naming for IDE and RTE within Eclipse wouldn't it? Maybe your build script could change the folder?

diplfranzhoepfinger:
yes, this turns out to be a pure "comfort" issue now. not more. 

i will make 2 Screenshots how it looks in Eclipse.

diplfranzhoepfinger:
it looks like this with non-Changed Name, 

checkout from [https://github.com/eclipse-4diac](https://github.com/eclipse-4diac)

and imported:



my Proposal leads to this: 



so just cosmetics now.

azoitl:
You can also rename your directory to your needs. I for example still have the old name `org.eclipse.4diac.forte` for my 4diac FORTE repo. The reason mainly is that I just changed the links for origin and upstream in my git config.

MartinMelikMerkumians:
IMHO, keeping `org.eclipse.4diac.forte` is preferable for me, as it also points to the project as a whole.

MartinMelikMerkumians (commented):
I would keep the original, as it points to eclipse and the project as a whole

MartinMelikMerkumians:
Thanks for your PR, but we keep the original name

PR #22, opened 2024-02-25
MQTT: Restore config file parser for authentication support

kumajaya:


azoitl:
Does this work also for multiple clients connections?

kumajaya:
Yes, I've tested it. From MQTTClientConfigParser.h:

"The parser looks for an endpoint, and only after it matches the one passed as argument, it starts storing the following information, and it will keep reading after another endpoint other the end of file is found"

OPC UA should have the same behavior, I will confirm it.

src/com/mqtt_paho/MQTTClient.cpp
```diff
@@ -186,6 +187,18 @@ int CMQTTClient::initClient() {
   mClientConnectionOptions.onSuccess = onMqttConnectionSucceed;
   mClientConnectionOptions.onFailure = onMqttConnectionFailed;
   mClientConnectionOptions.context = this;
+
+  if ("" != gMqttClientConfigFile) { //file was provided
+    CMQTTClientConfigFileParser::MQTTConfigFromFile result = CMQTTClientConfigFileParser::MQTTConfigFromFile(mUsername, mPassword);
```
azoitl: do you really need user name and password as fields of the Client class. You are immediately forwarding them to the client connection options. So wouldn't it be nicer to have them as local variables?

azoitl:
> Yes, I've tested it. From MQTTClientConfigParser.h:
> 
> "The parser looks for an endpoint, and only after it matches the one passed as argument, it starts storing the following information, and it will keep reading after another endpoint other the end of file is found"
> 
> OPC UA should have the same behavior, I will confirm it.

Cool. If you want to update the docs they are now also on github: https://github.com/eclipse-4diac/4diac-documentation

azoitl (approved):

PR #25, opened 2024-02-28
Fix compile errors for VS Build Tools 2019

kumajaya:


src/com/mqtt_paho/MQTTClient.cpp
```diff
@@ -188,9 +188,9 @@ int CMQTTClient::initClient() {
   mClientConnectionOptions.onFailure = onMqttConnectionFailed;
   mClientConnectionOptions.context = this;
 
+  std::string username;
```
kumajaya, line 191: I have no idea why MSVC needs a change like this to make MQTT authentication work.

src/com/modbus/modbusclientconnection.cpp
```diff
@@ -14,13 +14,16 @@
 #include "devlog.h"
 #include "modbuspoll.h"
 #include <forte_thread.h>
-#include <unistd.h> // for sleep
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
 #ifndef WIN32
+#include <unistd.h> // for sleep
```
azoitl: I think the better solution for this file is not to fix that here but use `CThread::sleepThread(` instead of sleep. Because the `CThread::sleepThread(` we do the hardware/OS abstraction allready.

src/core/datatypes/forte_wstring.h
```diff
@@ -52,13 +52,22 @@ class CIEC_WSTRING final : public CIEC_ANY_STRING {
       TForteWChar value = static_cast<TForteWChar>(paValue);
       auto &converter = std::use_facet<std::codecvt<char16_t, char, std::mbstate_t>>(std::locale());
       std::mbstate_t mb{};
+#ifdef _MSC_VER
```
azoitl: I don't think in this case it makes sense to have two versions. I would prefer to only have the one that compiles for all systems.

src/com/mqtt_paho/MQTTClient.cpp
azoitl: this is really strange.

azoitl (approved):

PR #26, opened 2024-02-28
Fix concurent modification of timerlist

### LINKED ISSUE #24: E_CYCLE crashes with assertation on windows.
The problem is a corrupt vector iterator in void CTimerHandler::processTimedFBList(). The problem definitly occurs if only one E_CYCLE is used. This results in the runtime assert:
assert_cycle-FORTE

The following code fixes this:

```
void CTimerHandler::processTimedFBList() {
  bool fired;
  do {
    fired = false;
    // We keep a list of sorted entries (oldest at the beginning) so we only have to check the first entry in the list:
    auto it = mTimedFBList.begin();
    if (it != mTimedFBList.end() && it->mTimeOut <= mForteTime) {
      STimedFBListEntry entry = *it; // buffer entry
      it = mTimedFBList.erase(it);
      triggerTimedFB(entry);  // This adds the entry again if not a one shot.
      fired = true; // Check the next entry.
    }
  } while (fired);
}
```

azoitl:
Periodic entries in the timer list, wich are readded after triggering into the list resulted in a concurent modifcation. With this change the list is handled correctly and als copying of entries is reduced.

Fixes #24

### DIFF
--- a/src/arch/timerha.cpp
+++ b/src/arch/timerha.cpp
@@ -1,6 +1,8 @@
 /*******************************************************************************
- * Copyright (c) 2005 - 2015 ACIN, Profactor GmbH, fortiss GmbH
- *               2020 Johannes Kepler University Linz
+ * Copyright (c) 2005, 2024 ACIN, Profactor GmbH, fortiss GmbH,
+ *                          Johannes Kepler University Linz,
+ *                          OFFIS e.V.
+ *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -22,6 +24,7 @@
 #include <functional>
 
 DEFINE_HANDLER(CTimerHandler)
+
 
 CTimerHandler::CTimerHandler(CDeviceExecution& paDeviceExecution) : CExternalEventHandler(paDeviceExecution),
     mForteTime(0) {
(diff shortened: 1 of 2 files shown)


src/arch/timerha.cpp
```diff
@@ -82,13 +83,11 @@ void CTimerHandler::nextTick() {
...
 
 void CTimerHandler::processTimedFBList() {
-  for (auto it = mTimedFBList.begin(); it != mTimedFBList.end();) {
-    if (it->mTimeOut > mForteTime) {
-      break;  
-    }
-    STimedFBListEntry entry = *it; // buffer entry
-    it = mTimedFBList.erase(it);
-    triggerTimedFB(entry);
+  auto last = std::lower_bound(mTimedFBList.begin(), mTimedFBList.end(), mForteTime);
```
jwalt: This is not equivalent. It leaves values that are equal to the target value in the list, whereas before equal values would be removed. The solution is to use `std::upper_bound`.

kumajaya:
With std::upper_bound, I got "call to object of class type 'std::less<void>': no matching call operator found" error on VS Build Tools 2022

azoitl:
> With std::upper_bound, I got "call to object of class type 'std::less': no matching call operator found" error on VS Build Tools 2022

can you say which operator is missing? would be important to fix this.

kumajaya:
> > With std::upper_bound, I got "call to object of class type 'std::less': no matching call operator found" error on VS Build Tools 2022
> 
> can you say which operator is missing? would be important to fix this.

Reapplied this patch and rebuild but the error did not appear. I'll try a clean build and back to confirm.

kumajaya:
> > With std::upper_bound, I got "call to object of class type 'std::less': no matching call operator found" error on VS Build Tools 2022
> 
> can you say which operator is missing? would be important to fix this.

Sorry, FALSE INFO. No errors at all.

azoitl:
> > > With std::upper_bound, I got "call to object of class type 'std::less': no matching call operator found" error on VS Build Tools 2022
> > 
> > 
> > can you say which operator is missing? would be important to fix this.
> 
> Sorry, FALSE INFO. No errors at all.

No problem. First of all thx for testing. This is a critical patch. The more eyes the better.

src/arch/timerha.cpp
azoitl: Thx for pointing this out. Changed that.

jwalt (approved):

MartinMelikMerkumians (approved):

PR #28, opened 2024-02-29
array_dynamic: Fix compile errors for VS Build Tools 2022

kumajaya:


azoitl:
According to https://en.cppreference.com/w/cpp/header/memory the missing functions destroy and destroy_n should be in the memory include. Have you tried that. 

Also please note that we are currently not on C++20 but C++17 as some of our platforms not yet have a C++20 compiler.

azoitl (approved):

PR #31, opened 2024-03-02
Port to Zephyr OS: Modular IO support incl. ESP32 example

dok-net:
As per the title, this PR completes the Zephyr OS port by adding the FB and supporting code to implement custom boards with modular IO.
An example for the ESP32 Ethernet Kit is included. This can be build in conjunction with https://github.com/dok-net/4diac.forte-esp32-sample.git


diplfranzhoepfinger:
I am very interested in this. 

especially i also made a IO Mapping for ESP32, 

but still WIP, and stored here for Access: 

[https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte](https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte)

latest Branch is 

https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte/-/tree/feature/TC-SC

diplfranzhoepfinger:
I am very interested in this. 

especially i also made a IO Mapping for ESP32, 

but still WIP, and stored here for Access: 

[https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte](https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte)

latest Branch is 

https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte/-/tree/feature/TC-SC

diplfranzhoepfinger:
most of ESP32 IO is here: 

[https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte/-/tree/feature/TC-SC/src/modules/commonIO](https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte/-/tree/feature/TC-SC/src/modules/commonIO)

it rely on [https://components.espressif.com/components/espressif/button](https://components.espressif.com/components/espressif/button)

an other edependencies like 

[https://components.espressif.com/components/espressif/led_strip](https://components.espressif.com/components/espressif/led_strip)

https://components.espressif.com/components/espressif/led_indicator

diplfranzhoepfinger:
most of ESP32 IO is here: 

[https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte/-/tree/feature/TC-SC/src/modules/commonIO](https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/4diac-forte/-/tree/feature/TC-SC/src/modules/commonIO)

it rely on [https://components.espressif.com/components/espressif/button](https://components.espressif.com/components/espressif/button)

an other edependencies like 

[https://components.espressif.com/components/espressif/led_strip](https://components.espressif.com/components/espressif/led_strip)

https://components.espressif.com/components/espressif/led_indicator

dok-net:
@diplfranzhoepfinger Should be of great interest to anyone extending the ESP32 support. Do you agree that in the context of this PR, I don't have to add anything? ESP32 isn't my primary target, I use it for fun at home and verification of the implementation for portability. If you feel like deploying Zephyr OS in your project - the learning curve is steep. Keep up with their releases. I would look at PRs for my example project as long as it doesn't take me too long to review. I'm spending a lot of unpaid spare time on this as it stands already :-)

src/arch/bsdsocketinterf.cpp
```diff
@@ -17,6 +17,14 @@
 #include "devlog.h"
 #include <string.h>
 
+#ifdef __ZEPHYR__
```
azoitl: The design of 4diac FORTE's bsdsocketinerface is such that these kind of includes and configurations should be done in the os specific includes and not here. 

Therefore please move this to `sockhand.h`in the zephyr directory. A good example is the `sockhand.h`of freeRTOS.

src/arch/bsdsocketinterf.cpp
```diff
@@ -175,7 +183,33 @@ CBSDSocketInterface::TSocketDescriptor CBSDSocketInterface::openUDPSendPort(char
           DEVLOG_WARNING("CBSDSocketInterface: setsockopt(IP_MULTICAST_IF) failed: %s\n", strerror(errno));
       }
     }
-#endif
+#else
+    if (paMCInterface) {
```
azoitl, line 179: I would prefer to have that somewhere else. If that is not possible at all please move it at least into a function so that the code stays readable.

src/arch/bsdsocketinterf.cpp
```diff
@@ -239,7 +273,34 @@ CBSDSocketInterface::TSocketDescriptor CBSDSocketInterface::openUDPReceivePort(c
           //if this fails we may have given a non multicasting addr. For now we accept this. May need to be changed in the future.
           DEVLOG_WARNING("CBSDSocketInterface: setsockopt(IP_ADD_MEMBERSHIP) failed: %s\n", strerror(errno));
         }
-#endif
+#else
+        // setting up multicast group
```
azoitl: I would prefer to have that somewhere else. If that is not possible at all please move it at least into a function so that the code stays readable.

src/modules/zephyr/types/ZephyrIO_fbt.h
```diff
@@ -0,0 +1,95 @@
+/*************************************************************************
```
azoitl, line 1: Please add here the correct copyright header.

src/modules/zephyr/types/ZephyrIO_fbt.cpp
```diff
@@ -0,0 +1,159 @@
+/*************************************************************************
```
azoitl, line 1: Please add here the correct copyright header.

src/modules/zephyr/types/Esp32EthernetKitIO_fbt.h
```diff
@@ -0,0 +1,109 @@
+/*************************************************************************
```
azoitl, line 1: Please add here the correct copyright header.

src/modules/zephyr/types/Esp32EthernetKitIO_fbt.cpp
```diff
@@ -0,0 +1,246 @@
+/*************************************************************************
```
azoitl, line 1: Please add here the correct copyright header.

src/modules/zephyr/TypeLibrary/io/zephyr/Esp32EthernetKitIO.fbt
azoitl, line 1: Please no function block sources in 4diac FORTE. WE have them either for our typelib in 4diac IDE or in an example project in 4diac Examples.

src/modules/zephyr/TypeLibrary/io/zephyr/ZephyrIO.fbt
azoitl, line 1: Please no function block sources in 4diac FORTE. WE have them either for our typelib in 4diac IDE or in an example project in 4diac Examples.

src/arch/threadbase.h
```diff
@@ -89,7 +89,7 @@ namespace forte {
 
         /*! \brief set the alive flag for this flag
          */
-        void setAlive(bool paVal) {
+        virtual void setAlive(bool paVal) {
```
azoitl: I'm not sure if I'm happy with that change. This method is on purpose not virtual.

azoitl (changes requested):

src/arch/threadbase.h
dok-net: Can you please advice on how else it's possible to break out of `runLoop`? I didn't come up with another solution and it seemed a general problem for any arch, anyway.

src/modules/zephyr/TypeLibrary/io/zephyr/ZephyrIO.fbt
dok-net: Please advice on what is the preferred way, I have not found an IDE typelib in Github, examples doesn't seem right to me, and using an own repo while the implementations of the FBs are an integral part of FORTE seems weird to me.

src/modules/zephyr/TypeLibrary/io/zephyr/Esp32EthernetKitIO.fbt
dok-net: Please advice on what is the preferred way, I have not found an IDE typelib in Github, examples doesn't seem right to me, and using an own repo while the implementations of the FBs are an integral part of FORTE seems weird to me.

src/modules/zephyr/TypeLibrary/io/zephyr/ZephyrIO.fbt
MartinMelikMerkumians: The 4diac IDE, which contains the typelib, is still on Gerrit
https://git.eclipse.org/r/plugins/gitiles/4diac/org.eclipse.4diac.ide

src/arch/threadbase.h
azoitl: I had to check a second time, because I thought we have a generic solution. But unfortunately when we did the last redesign of our thread infra we decided against a generic solution. AFAIR the reason was that the different use cases are to specific each time. On solution you can see in our CEventChainExecutionThread class. But after your feedback I'm not to happy about it anymore.

One thing that came to my mind is to introduce a new private virtual method `onEnd()` which threads could override to perform closing specific things. This method could be called from the end() method.

I think private is important because only CThreadBase should be allowed to call this method.

src/modules/zephyr/TypeLibrary/io/zephyr/ZephyrIO.fbt
dok-net: Please see https://git.eclipse.org/r/c/4diac/org.eclipse.4diac.ide/+/206613
Once merged, I will drop the TypeLibrary subdirectory from this PR immediately.

src/arch/threadbase.h
dok-net: @azoitl I respect your considerations. For the time being, may we keep my solution and revert/fix it once the greater refactoring takes place? Thank you for your effort.

src/arch/bsdsocketinterf.cpp
dok-net: There are #ifdefs for many archs in that file by now. Short of a full refactoring, I've done as suggested and designed a local linkage (anon namespace) function - looks pretty readable to me now, I hope it's fine with you as well. Thank you for the input.

src/arch/bsdsocketinterf.cpp
dok-net: Done.

src/arch/bsdsocketinterf.cpp
dok-net: Done.

src/arch/threadbase.h
azoitl: @dok-net as this is one of the most curcial components in 4diac FORTE I would not like to have this intermediate fix in it. However I can offer to do the proposed refactoring upfront your PR.

src/arch/bsdsocketinterf.cpp
azoitl: Yes there are. We are working on 4diac FORTE for nearly 20 years. There is still stuff in it that is not as good as it could be. Also we learned a lot in the last 20 years. For us `#ifdef`are now technical depth making maintenance harder. So if we can avoid it in new code we try as hard as possible to do so.

src/arch/threadbase.h
dok-net: @azoitl Yes, please, by all means, go ahead with that.

src/modules/zephyr/TypeLibrary/io/zephyr/ZephyrIO.fbt
dok-net: Removed the TypeLibrary subdirectory for Zephyr FBs from FORTE now.

src/arch/threadbase.h
azoitl: @dok-net done in PR #41. BTW thx for finding this problem. With the new api this is much cleaner and made CEventChainExecutionThread more correct.

kaartee: The modular IO core doesn't seem to correctly support asynchronous input... It keeps emitting two IND events at the ID where only one would be justified. That is for cyclical updates triggered from `runLoop()`. I am out of my depth how to fix this without modifying the core implementation, which is quite complicated at that.
I am referring to the IOHandleADC compiled for async ADC... the runLoop() causes a check of equal, which return false, in order to trigger a get. Once the async measurement started by get finishes, work_callback triggers another get. Which functions exactly I am using, is all visible in this PR. Unfortunately, this sequence of execution results in ID's emitting of TWO IND events, instead of one. Yes, it reads twice. But once is just perfunctory to even start the measurement. Is the modular IO not designed to support async input, that has to be actively triggered? ADC must be started, as opposed to GPIO interrupts...

azoitl: I think the problem of your solution is that you use the default `checkForInputChanges();`  will update and fire IND events for all registered FBs. This is maybe not what you want. So calling this default method from you runLoop is maybe not the right thing. The other option is that you store your ADC handles in a seperate list and manage them seperatly.

kaartee: I fixed it. Now REQ for ADC results in both CNF and IND, whereas cyclical sampling from the device configuration FB result in IND, only. This allows the app to connect to only IND in most cases, alleviating the need to deal with multiple events for the same measurement.

azoitl: Interesting approach

kaartee: Did the same for GPIO now - IND for any change of value, be it via IRQ or FB REQ event.

src/arch/bsdsocketinterf.cpp
```diff
@@ -142,14 +142,47 @@ int CBSDSocketInterface::receiveDataFromTCP(TSocketDescriptor paSockD, char* paD
...
 
   if(nRetVal == -1){
     DEVLOG_ERROR("CBSDSocketInterface: TCP-Socket recv() failed: %s\n", strerror(errno));
   }
   return handleError(nRetVal, "TCP");
 }
 
+#ifdef __ZEPHYR__
+namespace {
+  void setupMulticastGroup(char* paIPAddr, const char* paMCInterface) {
```
azoitl: You could move this to the sockhand.h / new cpp of zephyr which would reduce the complexity hiere in the generic socketinerface

azoitl:
Unfortunately your last commit was done with a different committer name and email then the one you have in the Eclipse account records. Can you please correct that and force push the fix.

dok-net:
> Unfortunately your last commit was done with a different committer name and email then the one you have in the Eclipse account records. Can you please correct that and force push the fix.

I noticed as much over the day :-) Done already. Thanks for mentioning it anyway.

src/arch/bsdsocketinterf.cpp
dok-net: Done.

src/arch/threadbase.h
kaartee: @azoitl This requested change was resolved, please re-review and close it accordingly, if you are OK with it.

azoitl (approved):

azoitl:
@dok-net github does not allow me to rebase your comitts. I assume squashing and merging is fine.

kaartee:
@azoitl You are right, rebase no longer is an option. I see that "This branch has no conflicts with base branch", therefore any type of permitted merge should be risk-free. Please go ahead! And feel free to announce the Zephyr port, just make sure to use my Dirk O. Kaar <dok@dok-net.net> identity, not the KT Elektronik one. The copyrights are OK, it's just that there is no desire, nor any point, in annoucing KT Elektronik's involvement.

azoitl:
> @azoitl You are right, rebase no longer is an option. I see that "This branch has no conflicts with base branch", therefore any type of permitted merge should be risk-free. Please go ahead! And feel free to announce the Zephyr port, just make sure to use my Dirk O. Kaar [dok@dok-net.net](mailto:dok@dok-net.net) identity, not the KT Elektronik one. The copyrights are OK, it's just that there is no desire, nor any point, in annoucing KT Elektronik's involvement.

Thx. Will prepare a Eclipse 4diac web-pagenews post as PR on the website repo and tag you for review later today.

PR #38, opened 2024-03-07
Fix bugs related to MSVC and build for Windows

mhansal:


src/core/datatypes/forte_array_dynamic.h
```diff
@@ -26,6 +26,7 @@
 #pragma once
 
 #include "forte_array.h"
+#include <memory>
```
azoitl: This was already fixed in an other PR

azoitl (changes requested):

kumajaya:
Build successfully with GCC 13.2.0 aarch64-musl, GCC 13.2.0 mingw-w64, and VS Build Tools 2022 amd64.

src/core/datatypes/forte_any_string.cpp
```diff
@@ -192,7 +192,7 @@ int CIEC_ANY_STRING::dollarEscapeChar(char *paValue, char paSymbol, unsigned int
       }
       break;
     default:
-      if(!isprint(paSymbol) && paTypeID == CIEC_ANY::e_STRING) {
+      if(!isprint((unsigned char)paSymbol) && paTypeID == CIEC_ANY::e_STRING) {
```
MartinMelikMerkumians: could you please change this to a C++ style cast (static_cast)

src/core/datatypes/forte_string_fixed.h
```diff
@@ -69,18 +70,26 @@ class CIEC_STRING_FIXED final : public CIEC_STRING {
...
 
-    CIEC_STRING_FIXED &operator=(CIEC_STRING_FIXED &&paValue) {
+    template<size_t maxLengthSrc>
+    CIEC_STRING_FIXED &operator=(CIEC_STRING_FIXED<maxLengthSrc> &&paValue) {
       getStorageMutable() = std::move(paValue.getStorageMutable());
+      if constexpr (maxLengthSrc > maxLength) {
+        getStorageMutable().resize(std::min<std::size_t>(this->length(), maxLength));
+      }
       return *this;
     }
```
MartinMelikMerkumians, line 82: Am I correct that this is necessary for MSVC but does not add new types of operator=?

tests/core/datatypes/CIEC_LTIME_test.cpp
```diff
@@ -588,10 +588,10 @@ BOOST_AUTO_TEST_CASE(parse_missing_s_in_microseconds_literal)
 {
   CIEC_LTIME time;
 
-  BOOST_CHECK_EQUAL(time.fromString("LT#-1u"), -1);
+  BOOST_CHECK_EQUAL(time.fromString("LT#-1u"), -1_LTIME);
   BOOST_CHECK_EQUAL(time.getInMicroSeconds(), 0);
 
-  BOOST_CHECK_EQUAL(time.fromString("LT#-1ut"), -1);
+  BOOST_CHECK_EQUAL(time.fromString("LT#-1ut"), -1_LTIME);
```
MartinMelikMerkumians: @azoitl shouldn't those tests fail? I mean neither -1u nor -1ut are valid time literals

tests/modules/utils/GEN_CSV_WRITER_tester.cpp
```diff
@@ -31,8 +31,12 @@ class GEN_CSV_WRITER_TestFixture : public CFBTestFixtureBase {
       setOutputData({&mQO, &mSTATUS});
       CFBTestFixtureBase::setup("CSV_WRITER_2");
 
+#ifdef WIN32
+      char filename[] = "./GEN_CSV_WRITER.tmp";
+#else
       char filename[] = "/tmp/fileXXXXXX";
       BOOST_CHECK(mkstemp(filename));
+#endif
```
MartinMelikMerkumians: Perhaps @azoitl can shed a light why mk(s)temp was used for this test. If it is necessary/wanted, then Windows could use mktemp for this test

MartinMelikMerkumians (changes requested):

src/core/datatypes/forte_string_fixed.h
mhansal: The commit included an improvement, changing the definition order is enough for now. I removed the improvements from this PR, maybe we can merge it later.

tests/modules/utils/GEN_CSV_WRITER_tester.cpp
mhansal: Using mkstemp is strange anyways, because it opens the file and returns a file descriptor. mktemp should be enough, also on Windows, even if it is deprecated.

mhansal: After more research, mktemp is probably no good solution. @azoitl Would using a fixed file path in the working directory be a valid solution?

tests/core/datatypes/CIEC_LTIME_test.cpp
azoitl: @MartinMelikMerkumians fromstring is returing an int with -1 one on failure or the number of used bytes on success. 

@mhansal why have you changed that to a time literal?

tests/modules/utils/GEN_CSV_WRITER_tester.cpp
azoitl: @mhansal yes this sounds for tests a much better aproach

tests/core/datatypes/CIEC_LTIME_test.cpp
MartinMelikMerkumians: Oh, I see. Yeah, the the comparison values are wrong. I thought I disabled the implicit conversion to builtin types from LTIME

tests/modules/utils/GEN_CSV_WRITER_tester.cpp
mhansal: Fixed.

tests/core/datatypes/CIEC_LTIME_test.cpp
mhansal: Fixed.

src/core/utils/ifSpecBuilder.cpp
```diff
@@ -76,12 +76,14 @@ std::size_t CStringIdListSpecBuilder::calcStorageSize() const {
   return cmStaticList ? 0 : mDynamicList.size() * sizeof(CStringDictionary::TStringId);
 }
 
-std::tuple<const CStringDictionary::TStringId*, TForteUInt8> CStringIdListSpecBuilder::build(CMixedStorage &paStorage) const {
+template<typename TSize>
```
MartinMelikMerkumians: Hi.
Not sure if `TSize` (Type Size, which is also a bit strange as T indicates a type definition, which you don't do here) is a good name here, as its not really a size but some sort of index or numeric identifier (`TPortID`, `TEventID`).
I personally would go with a generic `Type` or `IndexType`

MartinMelikMerkumians: @azoitl what is your opinion here?

src/core/utils/ifSpecBuilder.h
```diff
@@ -171,7 +171,9 @@ class CStringIdListSpecBuilder {
      * @return required dynamic data size
      */
     std::size_t calcStorageSize() const;
-    std::tuple<const CStringDictionary::TStringId*, TForteUInt8> build(CMixedStorage &paStorage) const;
+
+    template<typename TSize>
```
MartinMelikMerkumians: same comment as above, I would go with `Type `or `IndexType`

src/core/utils/mixedStorage.cpp
```diff
@@ -19,6 +19,7 @@ using namespace forte::core::util;
 
 void* CMixedStorage::write(const void *paBegin, std::size_t paSize) {
   auto pBegin = reinterpret_cast<const char*>(paBegin);
+  const std::size_t offset = mData.size();
```
MartinMelikMerkumians: Is it really necessary to specify std namespace here for size_t?

src/core/utils/ifSpecBuilder.cpp
azoitl: @MartinMelikMerkumians I agree IndexType would be better

src/core/utils/ifSpecBuilder.h
azoitl: I'll vote for IndexType as written above.

azoitl (changes requested):

src/core/utils/mixedStorage.cpp
kumajaya: It all comes from the initial of implementation.

MartinMelikMerkumians (approved):

azoitl (approved):

PR #41, opened 2024-03-10
Introduced onAliveChanged to CThreadBase

azoitl:
This method can be used by children of CThreadBase to perform any child specific actions when the alive state of this thread changes (e.g., release a mutex or semaphore).

### DIFF
--- a/src/arch/threadbase.h
+++ b/src/arch/threadbase.h
@@ -90,7 +90,11 @@ namespace forte {
         /*! \brief set the alive flag for this flag
          */
         void setAlive(bool paVal) {
+          bool oldAlive = mAlive;
           mAlive = paVal;
+          if(oldAlive != mAlive){
+            onAliveChanged(mAlive);
+          }
         }
 
         /*! \brief Helper method to run the thread.
@@ -128,6 +132,15 @@ namespace forte {
          * @return handle to the newly created thread
          */
         virtual TThreadHandleType createThread(long paStackSize) = 0;
+
+
+        /*! \brief Call back allowing children to perform child specific actions on alive state changes
+         *
+         * @param paNewValue the new value of the alive flag
+         */
+        virtual void onAliveChanged(bool paNewValue) {
+          (void)paNewValue;  // inhibit compiler warning
+        }
 
         //! Semaphore for implementing a generic join functionality. For a stable functionality this mutex must be locked during thread creation.
         CSemaphore mJoinSem;
--- a/src/core/ecet.cpp
+++ b/src/core/ecet.cpp
@@ -34,6 +34,12 @@ void CEventChainExecutionThread::run(){
   }
 }
 
+void CEventChainExecutionThread::onAliveChanged(bool paNewValue) {
+  if(!paNewValue){
+    resumeSelfSuspend();
+  }
+}
+
 void CEventChainExecutionThread::mainRun(){
   if(externalEventOccured()){
     transferExternalEvents();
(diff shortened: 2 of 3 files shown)


kaartee:
Looks good to me, w.r.t. https://github.com/eclipse-4diac/4diac-forte/pull/31

MartinMelikMerkumians (approved):

kaartee:
@azoitl Oops, turns out that it doesn't look perfectly good to me when I actually try to build my PR based off this change.
As I call:
```cpp
protected:
  void onAliveChanged(bool paNewValue) override {
    forte::core::io::IODeviceController::onAliveChanged(paNewValue);
    mUpdateSema.inc();
  }
```
The accessibility "private" in this PR gets in the way.
What do you think, should an override invoke the base class implementation, or not?

azoitl:
> @azoitl Oops, turns out that it doesn't look perfectly good to me when I actually try to build my PR based off this change. As I call:
> 
> ```c++
> protected:
>   void onAliveChanged(bool paNewValue) override {
>     forte::core::io::IODeviceController::onAliveChanged(paNewValue);
>     mUpdateSema.inc();
>   }
> ```
> 
> The accessibility "private" in this PR gets in the way. What do you think, should an override invoke the base class implementation, or not?

The reason for making it private is that no child can call it. With that on child should also not invoke the base class implementation.

kaartee:
> The reason for making it private is that no child can call it. With that on child should also not invoke the base class implementation.

No offense, but I had to ask if it's totally on purpose. Fine, I remove the call.

kaartee:
On second thought, I am beginning to have some concerns.
It appears to be generally frowned upon to provide private virtuals due to the level of confusion this sows.
In the particular case, `resumeSelfSuspend()` is dependent on the base implementation of `onAliveChange()` getting called, therefore any override of `onAliveChanged()` according to your specification alters the function of `CEventChainExecutionThread::changeExecutionState()`, doesn't it?

At the level of using `forte::core::io::IODeviceController::onAliveChanged()` I don't think anyone knows about this effect...?

azoitl:
> No offense, but I had to ask if it's totally on purpose. Fine, I remove the call.
I'm not offended. This is one of our most critical classes any pair of eyes and test helping us ensuring correct behavior is extremely helpfull.

> On second thought, I am beginning to have some concerns. It appears to be generally frowned upon to provide private virtuals due to the level of confusion this sows. 

I learned about this concept in one of the books from Herb Suttner. We are by now using it in several places and so far found it nearly always more helpful to have it. 

>In the particular case, `resumeSelfSuspend()` is dependent on the base implementation of `onAliveChange()` getting called, therefore any override of `onAliveChanged()` according to your specification alters the function of `CEventChainExecutionThread::changeExecutionState()`, doesn't it?

`resumeSelfSuspend()` is used not only in `onAliveChanged()` but also for other things. I agree that if a child class of ECET is not correctly overriding `onAliveChange()` ECET can be broken. On the other hand now that we have onAliveChange ECET is more correct as anyone calling `end()` without calling `changeExecutionState()` will get the correct behavior. Also potential childs of ECET would mostly be either something related with simulation (e.g., FMU) or testing. In both cases the new setup allows for more correctness and flexibility. Therefore I'm willing to take the risk.

kaartee:
@azoitl You lost me :-) Perhaps it boils down to what "child class of ECET is not correctly overriding onAliveChange()" means.
What I'm really trying to find out here is whether my PR is correct or not. Right now, the override of `onAliveChange()` gives a semaphore, but no longer  does `resumeSelfSuspend()`. I don't know it the thread running `runLoop()` could get suspended until `setAlive()` triggers its resumption. In other words, must I call `resumeSelfSuspend()` directly in the override of `onAliveChange()` now, or rather not?

azoitl:
Ah sorry for me ECET is always the short form of `CEventChainExecutionThread`. `resuemeSelfSusbpend() is a method of `CEventChainExecutionThread`. As you are not inheriting from this class this is not bothering you.

PR #44, opened 2024-03-13
Remove gcc11 dependency

cochicde:
As for https://stackoverflow.com/questions/63616709/incomplete-type-stdvariant-used-in-nested-name-specifier, the problem comes from a possible unclear specification. There is also shown a workaround of casting to the `parent::variant` class so the visit can work properly.

If this patch  doesn't create any problem and it only adds a `static_cast` I think it is worth having it to expand the compiler list which work with forte.

### DIFF
--- a/buildsupport/check_compiler_features.cmake
+++ b/buildsupport/check_compiler_features.cmake
@@ -24,13 +24,6 @@ function(TEST_NEEDED_FORTE_COMPILER_FEATURES)
     message(WARNING "`nullptr` not supported, replacing it with 0 via precprocessor")
     forte_add_definition("-Dnullptr=0")
   endif()
-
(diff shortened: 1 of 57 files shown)


azoitl:
@mx990 and @MartinMelikMerkumians as the base for this PR is mostly from your side I kindly ask to have a look. What I already see is that we still have a lower bound on GCC versions as we are using C++17 features that are not available below a certain GCC version.

mx990:
This PR looks reasonable to me. I believe we should then require GCC starting with version 9, which is the mimumum version to offer stable C++17 support, according to https://gcc.gnu.org/projects/cxx-status.html#cxx17.

MartinMelikMerkumians:
We can also add a CMake check for a certain C++ standard. As far as I know, this just wouldn't work with VS2018 but not sure if that version even supports C++17

MartinMelikMerkumians (approved):

PR #49, opened 2024-03-16
[convert] add missing data types

moniwe:


azoitl:
We started a work that makes the dt2dt obsolete: A configurable `F_MOVE`.

With that you just use the `F_MOVE` block and configure the data type to be used in its property sheet. In the IDE nightly this is already available. However 4diac FORTE code and deployment is still missing. Should come in the next weeks.

PR #52, opened 2024-03-20
Add Monitoring support for OPC UA Device Management

m-meingast:
Implement addWatch, readWatches, and removeWatches methods.

### DIFF
--- a/src/stdfblib/ita/OPCUA_MGR.cpp
+++ b/src/stdfblib/ita/OPCUA_MGR.cpp
@@ -186,6 +186,13 @@ char OPCUA_MGR::smAddWatchArgDescription[] = "Fully qualified name of FB Port";
 char OPCUA_MGR::smAddWatchAttrDisplayName[] = "Add Watch";
 char OPCUA_MGR::smAddWatchAttrDescription[] = "Add Watch";
 
+/* Read Watches */
+char OPCUA_MGR::smReadWatchesMethodName[] = "readWatches";
+char OPCUA_MGR::smReadWatchesOutArgName[] = "Response";
+char OPCUA_MGR::smReadWatchesOutArgDescription[] = "Read Response";
+char OPCUA_MGR::smReadWatchesDisplayName[] = "Read Watches";
+char OPCUA_MGR::smReadWatchesDescription[] = "Read Watches";
+
 #endif // FORTE_SUPPORT_MONITORING
 
 /* Initialize UA Status Codes */
@@ -246,6 +253,9 @@ EMGMResponse OPCUA_MGR::createIEC61499MgmtObject(UA_Server* paServer) {
   if (addKillDeviceMethod(paServer) != EMGMResponse::Ready) return eRetVal;
   if (addKillResourceMethod(paServer) != EMGMResponse::Ready) return eRetVal;
   if (addDeleteResourceMethod(paServer) != EMGMResponse::Ready) return eRetVal;
+#ifdef FORTE_SUPPORT_MONITORING
+  if (addReadWatchesMethod(paServer) != EMGMResponse::Ready) return eRetVal;
+#endif // FORTE_SUPPORT_MONITORING
   return addMgmtObjectInstance();
 }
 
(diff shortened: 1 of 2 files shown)


src/stdfblib/ita/OPCUA_MGR.cpp
```diff
@@ -949,6 +981,101 @@ UA_StatusCode OPCUA_MGR::onDeleteConnection(UA_Server*,
...
+
+UA_StatusCode OPCUA_MGR::onRemoveWatch(UA_Server*,
+  const UA_NodeId*, void*,
+  const UA_NodeId*, void* methodContext,
+  const UA_NodeId*, void* objectContext,
+  size_t, const UA_Variant* input,
+  size_t, UA_Variant*) {
+  if (methodContext != nullptr) {
+    EMGMResponse eRetVal = EMGMResponse::UnsupportedType;
+    std::string destination;
```
MartinMelikMerkumians: Could these two lines be combined into one?
`std::string destination(getInputValue(*static_cast<UA_String*>(input[0].data));)` or
`std::string destination = getInputValue(*static_cast<UA_String*>(input[0].data));`

By passing the value to the ctor, the string storage would be initialized in the appropriate size directly, and you could make `std::string destination` even `const`.

While outside this review, I would also recommend making the first parameter of the `parseDestinationName` also `const`, as the string is never modified

m-meingast: Fixed in new commit

src/stdfblib/ita/OPCUA_MGR.cpp
```diff
@@ -1016,7 +1124,7 @@ void OPCUA_MGR::setMGMCommand(EMGMCommandType paCMD, CStringDictionary::TStringI
     mCommand.mSecondParam.pushBack(CStringDictionary::getInstance().insert(paSecondParam));
   }
   if (paAdditionalParams != nullptr) {
-    mCommand.mAdditionalParams.fromString(paAdditionalParams);
+    mCommand.mAdditionalParams = CIEC_STRING(paAdditionalParams);
```
kumajaya, line 1127: This will trigger deprecated warning "Use ctor with explicit length parameter".

src/stdfblib/ita/OPCUA_MGR.cpp
```diff
@@ -1037,7 +1145,7 @@ void OPCUA_MGR::setMGMCommand(EMGMCommandType paCMD, CStringDictionary::TStringI
     }
   }
   if (paAdditionalParams != nullptr) {
-    mCommand.mAdditionalParams.fromString(paAdditionalParams);
+    mCommand.mAdditionalParams = CIEC_STRING(paAdditionalParams);
```
kumajaya, line 1148: This will trigger deprecated warning "Use ctor with explicit length parameter".

src/stdfblib/ita/OPCUA_MGR.cpp
MartinMelikMerkumians: True, please add the length of the string. If it's guaranteed to be a nul terminated c-string you can use strlen

azoitl: I don't think this is worth the effort.  In the next PR planned by @m-meingast `mAdditionalParams` will be turned into a `std:string` as it should be.  

But as this is a bigger change independent of this one I suggested to him to leave this PR as is.

@MartinMelikMerkumians  is this ok for you?

kumajaya: My question is how to enable OPC UA Device Management and Monitoring? Is it enabled by default?

azoitl: In 4diac IDE it will be available as soon as you build your 4diac FORTE with OPC UA. In 4diac IDE you need to set the device profile of your device to OPC UA xxx.

MartinMelikMerkumians: @azoitl fine for me

kumajaya: > In 4diac IDE it will be available as soon as you build your 4diac FORTE with OPC UA. In 4diac IDE you need to set the device profile of your device to OPC UA xxx.

Using milestone IDE here and no OPC UA device profile beside HOLOBLOC, FBDK2, and DynamicTypeLoad. I always build FORTE with OPC UA enabled.

azoitl: OPC UA deployment is currently only in the nightly build 4diac IDE. Later this week we will merge develop to freeze which will update the milestone 4diac IDE version.

PR #56, opened 2024-03-21
Support EQ and NE for array and struct

mx990:


src/core/datatypes/forte_array_common.h
```diff
@@ -29,7 +29,24 @@
  */
 template<typename T>
 class CIEC_ARRAY_COMMON : public CIEC_ARRAY {
-public:
+    friend bool operator==(const CIEC_ARRAY_COMMON &paLeft, const CIEC_ARRAY_COMMON &paRight) {
+      if (paLeft.getLowerBound() != paRight.getLowerBound() || paLeft.getUpperBound() != paRight.getUpperBound()) {
+        return false;
+      }
+
+      for (intmax_t i = paLeft.getLowerBound(), end = paRight.getUpperBound(); i <= end; ++i) {
```
MartinMelikMerkumians: I know that the bounds of left and right are checked on equality before, but is there a special reason to use paLeft.getLowerBound() and then paRight.getUpperBound()?
If there is no reason, I would personally prefer to use bounds of either left or right, as people might wonder if there is a special reason

src/core/datatypes/forte_array.h
```diff
@@ -27,7 +27,24 @@
 /** \brief A common supertype for all CIEC_ARRAY variants, providing the minimal interface an array must provide
  */
 class CIEC_ARRAY : public CIEC_ANY_DERIVED {
-public:
+    friend bool operator==(const CIEC_ARRAY &paLeft, const CIEC_ARRAY &paRight) {
+      if (paLeft.getLowerBound() != paRight.getLowerBound() || paLeft.getUpperBound() != paRight.getUpperBound()) {
+        return false;
+      }
+
+      for (intmax_t i = paLeft.getLowerBound(), end = paRight.getUpperBound(); i <= end; ++i) {
```
MartinMelikMerkumians: I know that the bounds of left and right are checked on equality before, but is there a special reason to use paLeft.getLowerBound() and then paRight.getUpperBound()?
If there is no reason, I would personally prefer to use bounds of either left or right, as people might wonder if there is a special reason

MartinMelikMerkumians (changes requested):

src/core/datatypes/forte_array_common.h
mx990: Ah, makes sense. I changed it.

src/core/datatypes/forte_array.h
mx990: Ah, makes sense. I changed it.

MartinMelikMerkumians (approved):

PR #60, opened 2024-03-23
win32: Use the standard C++ time functions

kumajaya:
This fix pre Windows 8/Server 2012 compatibility issue since the GetSystemTimePreciseAsFileTime function is not available in Windows 7

### DIFF
--- a/src/arch/win32/forte_architecture_time.cpp
+++ b/src/arch/win32/forte_architecture_time.cpp
@@ -15,24 +15,12 @@
  *  Martin Jobst - add high-resolution realtime clock
  *  Ketut Kumajaya - Use the standard C++ time functions, fix pre Windows 8/
  *                   Server 2012 compatibility issue
+ *                 - Move the standard C++ time functions as a common time
+ *                   implementation
  *******************************************************************************/
-
-#include <chrono>
-
-using namespace std::chrono;
 
 #include "forte_architecture_time.h"
 #include "forte_constants.h"
-
-#ifndef FORTE_FAKE_TIME
-uint_fast64_t getNanoSecondsMonotonic() {
-  return static_cast<uint_fast64_t>(duration_cast<nanoseconds>(steady_clock::now().time_since_epoch()).count());
-}
-
-uint_fast64_t getNanoSecondsRealtime() {
-  return static_cast<uint_fast64_t>(duration_cast<nanoseconds>(system_clock::now().time_since_epoch()).count());
-}
-#endif
 
 time_t forte_timegm(struct tm *pa_tm) {
   return _mkgmtime(pa_tm);
--- a/src/arch/win32/forte_architecture_time.cpp
+++ b/src/arch/win32/forte_architecture_time.cpp
@@ -2,6 +2,7 @@
  * Copyright (c) 2019 TU Wien/ACIN
  *               2022 Primetals Technologies Austria GmbH
  *               2023 Martin Erich Jobst
+ *               2024 Samator Indo Gas
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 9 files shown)


kumajaya:
@azoitl Sir, I would like to switch the authorship of this commit to the original author because almost none of the codes come from me but of course I can't. What is the best way to handle a situation like this? Is it enough with a link to the original commit?

azoitl:
@kumajaya it is definitely not possible that you partly copy code from other people to submit as Eclipse 4diac PR. This violates the copy-right of the original authors. Only with the permission of the original authors this is possible. 

@jwalt I had in the back of my mind that your build infrastructure is helping in this situation, am I right?

kumajaya:
> @kumajaya it is definitely not possible that you partly copy code from other people to submit as Eclipse 4diac PR. This violates the copy-right of the original authors. Only with the permission of the original authors this is possible.
> 
> @jwalt I had in the back of my mind that your build infrastructure is helping in this situation, am I right?

I just sent an email to the original author asking for permission to reuse his work.

jwalt:
@kumajaya after some research, I see a better solution that should be more portable and future-proof. Can you revise your patch to use C++11 std::chrono instead of Windows API functions?

https://stackoverflow.com/questions/48211688/how-accurate-is-stdchrono should give you initial pointers. I have checked, and on MinGW that should automatically be compatible at least down to Windows XP. MSVC windows support probably depends on which version of MSVC you use. In all cases, you should get true nanosecond resolution.

kumajaya:
> @kumajaya after some research, I see a better solution that should be more portable and future-proof. Can you revise your patch to use C++11 std::chrono instead of Windows API functions?
> 
> https://stackoverflow.com/questions/48211688/how-accurate-is-stdchrono should give you initial pointers. I have checked, and on MinGW that should automatically be compatible at least down to Windows XP. MSVC windows support probably depends on which version of MSVC you use. In all cases, you should get true nanosecond resolution.

I just modified FORTE main.cpp and add:
```
    auto timeNow = system_clock::now();
    std::cout << duration_cast<nanoseconds>(timeNow.time_since_epoch()).count() << " system_clock::now()\n";
    std::cout << getNanoSecondsRealtime() << " getNanoSecondsRealtime()\n";
    timeNow = system_clock::now();
    std::cout << duration_cast<nanoseconds>(timeNow.time_since_epoch()).count() << " system_clock::now()\n";
    std::cout << getNanoSecondsRealtime() << " getNanoSecondsRealtime()\n";
```
then got output like this:
```
1711711433355238900 system_clock::now()
1711711433356667500 getNanoSecondsRealtime()
1711711433358095000 system_clock::now()
1711711433359376200 getNanoSecondsRealtime()
```
Actually I don't care much about precision but prefer more portable functions.

@mx990 What is your opinion here since the initial codes came from you?

azoitl:
@kumajaya I would assume that both deliver the same time. If you take the delta of the times you can see that it is always app. 1.4ms. Which is reasonable as you are doing couts in between.  I guess a better test would be a loop where you call both (`system_clock::now()` and `getNanoSecondsRelatime()`) and then do the output for both. maybe even the delta.

But I agree with you and @jwalt that the more portable solution seems to be the better approach.

kumajaya:
> @kumajaya I would assume that both deliver the same time. If you take the delta of the times you can see that it is always app. 1.4ms. Which is reasonable as you are doing couts in between. I guess a better test would be a loop where you call both (`system_clock::now()` and `getNanoSecondsRelatime()`) and then do the output for both. maybe even the delta.
> 
> But I agree with you and @jwalt that the more portable solution seems to be the better approach.
@azoitl Sir, I do like this:
```
    const int num_measurements = 11;
    uint_fast64_t c_measurements[num_measurements];
    uint_fast64_t w_measurements[num_measurements];
    for (int i = 0; i < num_measurements; ++i) {
      c_measurements[i] = (uint_fast64_t) duration_cast<nanoseconds>(system_clock::now().time_since_epoch()).count();
      w_measurements[i] = getNanoSecondsRealtime();
    }

    for (int i = 0; i < (num_measurements - 1); ++i) {
      int duration = c_measurements[i + 1] - c_measurements[i];
      std::cout << c_measurements[i] << ":" << duration << " system_clock::now()\n";
      duration = w_measurements[i + 1] - w_measurements[i];
      std::cout << w_measurements[i] << ":" << duration << " getNanoSecondsRealtime()\n";
    }
```
and the output:
```
1711717095620364800:100 system_clock::now()
1711717095620364800:100 getNanoSecondsRealtime()
1711717095620364900:0 system_clock::now()
1711717095620364900:0 getNanoSecondsRealtime()
1711717095620364900:100 system_clock::now()
1711717095620364900:100 getNanoSecondsRealtime()
1711717095620365000:0 system_clock::now()
1711717095620365000:100 getNanoSecondsRealtime()
1711717095620365000:100 system_clock::now()
1711717095620365100:0 getNanoSecondsRealtime()
1711717095620365100:0 system_clock::now()
1711717095620365100:100 getNanoSecondsRealtime()
1711717095620365100:100 system_clock::now()
1711717095620365200:0 getNanoSecondsRealtime()
1711717095620365200:0 system_clock::now()
1711717095620365200:100 getNanoSecondsRealtime()
1711717095620365200:100 system_clock::now()
1711717095620365300:0 getNanoSecondsRealtime()
1711717095620365300:0 system_clock::now()
1711717095620365300:100 getNanoSecondsRealtime()
```

azoitl:
Yes they yield definitely the same result.

kumajaya:
> Yes they yield definitely the same result.

@azoitl @jwalt Based on https://learn.microsoft.com/en-us/cpp/standard-library/steady-clock-struct?view=msvc-170 I also check the monotonic version and got:

```
306559593626300:100 steady_clock::now()
306559593626300:100 getNanoSecondsMonotonic()
306559593626400:100 steady_clock::now()
306559593626400:100 getNanoSecondsMonotonic()
306559593626500:0 steady_clock::now()
306559593626500:100 getNanoSecondsMonotonic()
306559593626500:100 steady_clock::now()
306559593626600:100 getNanoSecondsMonotonic()
306559593626600:100 steady_clock::now()
306559593626700:0 getNanoSecondsMonotonic()
306559593626700:100 steady_clock::now()
306559593626700:100 getNanoSecondsMonotonic()
306559593626800:0 steady_clock::now()
306559593626800:100 getNanoSecondsMonotonic()
306559593626800:100 steady_clock::now()
306559593626900:0 getNanoSecondsMonotonic()
306559593626900:100 steady_clock::now()
306559593626900:100 getNanoSecondsMonotonic()
306559593627000:0 steady_clock::now()
306559593627000:100 getNanoSecondsMonotonic()
```

MartinMelikMerkumians:
Time values look good to me

azoitl:
If I see this right we should use the standard C++ time functions for at least Windows and posix arch am I right?

mx990:
Yes, I also agree that we should switch to `std::chrono` clocks on Windows and Posix architectures.

I have found the following additional resources:

> On Windows, steady_clock wraps the QueryPerformanceCounter function.

from the official documentation at https://learn.microsoft.com/en-us/cpp/standard-library/steady-clock-struct

It seems there was also some internal discussion about using less-precise clocks for performance reasons in https://github.com/microsoft/STL/issues/624. Fortunately for us, this was rejected. In the beginning, it also states:
> Currently std::chrono::system_clock::now() is GetSystemTimePreciseAsFileTime() on Vista+

So it seems we are safe on Windows for the time being.

On Linux, I found that at least the gcc `libstdc++` also simply wraps `clock_gettime` for `std::chrono::system_clock` and `steady_clock`, see [chrono.cc#L59](https://github.com/gcc-mirror/gcc/blob/28dca4be504dda29f55eafe958cdf299ec89b94e/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L59) and [chrono.cc#L87](https://github.com/gcc-mirror/gcc/blob/28dca4be504dda29f55eafe958cdf299ec89b94e/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L87), respectively.

src/arch/win32/forte_architecture_time.cpp
mx990, line 1: I would suggest making this the common implementation under `src/arch` and adapting both `src/arch/posix/CMakeLists.txt` and `src/arch/win32/CMakeLists.txt` to use the common implementation.

mx990 (changes requested):

src/arch/win32/forte_architecture_time.cpp
kumajaya: I have made a separate commit as per your suggestion.

mx990 (approved):

PR #65, opened 2024-03-31
Update OPC DA

kumajaya:
* Don't do any unnecessary preventions for string conversion problems
* Remove items and groups depend on FB config
* Code refactoring from char* to std::string
* Proper memory usage and deallocation

### DIFF
--- a/src/com/opc/opcconnectionimpl.cpp
+++ b/src/com/opc/opcconnectionimpl.cpp
@@ -236,8 +236,7 @@ int COpcConnectionImpl::sendItemData(const char *paGroupName, const char *paItem
   if(it != mOpcItems.end()){
     items = it->second;
     for(size_t i = 0; i < items.size(); i++){
-      std::string itemName = WS2S(items[i]->getName());
-      if(0 == strcmp(itemName.c_str(), paItemName)){
+      if(WS2S(items[i]->getName()) == paItemName){
         try{
           items[i]->writeSync(paVar);
         }
@@ -266,14 +265,12 @@ void COpcConnectionImpl::OnDataChange(COPCGroup & paGroup, COPCItemDataMap & paC
   while(pos){
     OPCItemData *data = paChanges.GetNextValue(pos);
     if (data) {
-      // Possible unpredictable string behavior in a multi-threaded program
-      std::string itemName = WS2S(data->item()->getName());
-      itemList.push_back(new SOpcItemData(itemName.c_str(), (Variant) data->vDataValue));
+      itemList.push_back(new SOpcItemData(WS2S(data->item()->getName()).c_str(), (Variant) data->vDataValue));
     }
   }
 
   // Possible unpredictable string behavior in a multi-threaded program
-  std::string s_groupName = WS2S(paGroup.getName());
+  const std::string s_groupName(WS2S(paGroup.getName()));
   const char *c_groupName = s_groupName.c_str();
 
   long long position = 0;
--- a/src/com/opc/opcconnectionimpl.cpp
+++ b/src/com/opc/opcconnectionimpl.cpp
@@ -10,6 +10,7 @@
  *   Filip Andren, Alois Zoitl - initial API and implementation and/or initial documentation
  *   Tibalt Zhao - add the list of items instead of add item one by one
  *   Ketut Kumajaya - switch to OPCClientToolKit with 64bit support
+ *                  - Code refactoring from char* to std::string
  *******************************************************************************/
 #include "opcconnectionimpl.h"
 #include "../../arch/devlog.h"
@@ -26,9 +27,10 @@
 #define S2WS(x) COPCHost::S2WS(x)
 #define WS2S(x) COPCHost::WS2S(x)
 
+using namespace std::string_literals;
+
 COpcConnectionImpl::COpcConnectionImpl(const char *paHost, const char *paServerName, COpcConnection* paOpcConn) :
-    mOpcConn(paOpcConn), mOpcHost(0), mOpcServer(0),mConnected(0), mHost(paHost), mServerName(paServerName),
-    mGroupName(0), mReqUpdateRate(0), mRealUpdateRate(0), mDeadBand(0) {
+    mOpcConn(paOpcConn), mOpcHost(0), mOpcServer(0),mConnected(0), mHost(paHost), mServerName(paServerName) {
 }
 
 COpcConnectionImpl::~COpcConnectionImpl(){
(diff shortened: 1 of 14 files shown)


src/com/opc/opcconnectionimpl.cpp
```diff
@@ -236,8 +236,7 @@ int COpcConnectionImpl::sendItemData(const char *paGroupName, const char *paItem
   if(it != mOpcItems.end()){
     items = it->second;
     for(size_t i = 0; i < items.size(); i++){
-      std::string itemName = WS2S(items[i]->getName());
-      if(0 == strcmp(itemName.c_str(), paItemName)){
+      if(0 == strcmp(WS2S(items[i]->getName()).c_str(), paItemName)){
```
azoitl: as you are comparing a std::string with a char * you can just use:
`if(WS2S(items[i]->getName()) == paItemName) {` 

I guess it would also make sense to consider change paItemName to a std:string, but I would see this in a separate PR.

src/com/opc/opcconnectionimpl.cpp
```diff
@@ -266,14 +265,12 @@ void COpcConnectionImpl::OnDataChange(COPCGroup & paGroup, COPCItemDataMap & paC
...
-      // Possible unpredictable string behavior in a multi-threaded program
-      std::string itemName = WS2S(data->item()->getName());
-      itemList.push_back(new SOpcItemData(itemName.c_str(), (Variant) data->vDataValue));
+      itemList.push_back(new SOpcItemData(WS2S(data->item()->getName()).c_str(), (Variant) data->vDataValue));
     }
   }
 
   // Possible unpredictable string behavior in a multi-threaded program
-  std::string s_groupName = WS2S(paGroup.getName());
+  const std::string s_groupName = static_cast<const std::string>(WS2S(paGroup.getName()));
```
azoitl: why do you need this cast?

azoitl (changes requested):

src/com/opc/opcconnectionimpl.cpp
kumajaya: Yes, I just restore the previous initial update. Save it as a temporary std::string not needed.

src/com/opc/opcconnectionimpl.cpp
kumajaya: Just to make sure the group name really is an immutable constant. Previously I had a string conversion problem, somehow groups and items name get corrupted so the OPC items could not be read.

MartinMelikMerkumians: So you need a copy of the mutable value? If that is the case I would make the copy explicit and not hide it via a cast which probably exactly does this in the background.
I think this should work by passing the value to the ctor
const std::string groupName(WS2S(paGroup.getName()));

Please also remove the s_ from the variables name

azoitl: while I was thinking about your patch I was pondering what is the task of the WS2S? You are calling it a lot. Would it make sense to store the result of this function in the data structures?

kumajaya: @MartinMelikMerkumians I have to retain the name because there is another variable with the same name.
@azoitl It's a helper function from OPCClientToolKit library to convert std::wstring to std::string:
```
    static std::string WS2S(const std::wstring &wstr)
    {
        return std::wstring_convert<std::codecvt_utf8<wchar_t>>().to_bytes(wstr);
    }

    static std::wstring S2WS(const std::string &str)
    {
        return std::wstring_convert<std::codecvt_utf8<wchar_t>>().from_bytes(str);
    }
```

src/com/opc/opcconnectionimpl.cpp
kumajaya: @azoitl Sir, what exactly are the advantages of switching from char* to std::string?

MartinMelikMerkumians: @kumajaya afaik c_str() generates a C-String representation which is then used for strcmp, while using the comparison operator should directly compare to the internal storage representation of std::string without creating the intermediate C-String representation.

azoitl: Furthermore std::string is memory save and the source is mostly also std::string. Converting back and forth is unnecessary and potential dangerous overhead.

kumajaya: Thank you for the explanation. Currently I'm working on code refactoring from char* to std::string. More c_sting() conversion on DEVLOG but core logic string operation much simple.

azoitl: DEVLOG is definitly an issue that we also need to look into. But if it makes the rest of to OPC code better I would go for that.

kumajaya: > DEVLOG is definitly an issue that we also need to look into. But if it makes the rest of to OPC code better I would go for that.

Sir, may I just push the changes here as another commit?

azoitl: I guess an own PR would be nicer or?

kumajaya: > I guess an own PR would be nicer or?

Actually yes, but it's depend on this PR.

azoitl: Ok then it makes more sense that you push it on top of this PR and I will not squash the changes.

kumajaya:
@azoitl @MartinMelikMerkumians
I hope my understanding to const std::string& is correct:
* No copy.
* No allocation in the heap.
* Quick behavior.
* Inform developers that the variable is not altered.

Reference: https://grapeprogrammer.com/const_string_reference_cpp/

azoitl:
> @azoitl @MartinMelikMerkumians I hope my understanding to const std::string& is correct:
> 
>     * No copy.
> 
>     * No allocation in the heap.
> 
>     * Quick behavior.
> 
>     * Inform developers that the variable is not altered.
> 
> 
> Reference: https://grapeprogrammer.com/const_string_reference_cpp/

Yes this is my understanding of const references to std::string.

There is an exception to this rule: When C++ compiler could apply the move semantics to transfer the string's content into a newly created object then a pass by value is more efficient. Am I right @MartinMelikMerkumians ?

src/com/opc/opcconnectionimpl.cpp
```diff
@@ -270,73 +265,52 @@ void COpcConnectionImpl::OnDataChange(COPCGroup & paGroup, COPCItemDataMap & paC
...
-          free(groupName);
         } else {
           retGroup = (*it)->mOpcGroupRead;
         }
       } else{
         if(!((*it)->mWriteGroupAdded)){
-          char *groupName = (char*) malloc(strlen(paGroupName) + 6 + 1);
-          strcpy(groupName, paGroupName);
-          strcat(groupName, "_write");
+          const std::string groupName(paGroupName + std::string("_write"));
```
azoitl: you can use here the new C++ literal syntax: `"_write"s`:
https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s

kumajaya: > you can use here the new C++ literal syntax: `"_write"s`: https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s

It just worked 😄 Now I'm thinking to update this PR as a single commit since the last commit almost invalidate all previous commits.

azoitl: > It just worked 😄 Now I'm thinking to update this PR as a single commit since the last commit almost invalidate all previous commits.

That is not needed. Github allows me to easily squash all the changes into one single commit.

kumajaya: > > It just worked 😄 Now I'm thinking to update this PR as a single commit since the last commit almost invalidate all previous commits.
> 
> That is not needed. Github allows me to easily squash all the changes into one single commit.

OK then. I found more dead code here and need a little bit clean up.

kumajaya: Finish. There is still one problem left (from the beginning), the destructor doesn't always work properly. Killing FORTE process (CTRL + C) doesn't always successfully remove OPC items and groups but clearing the device (or re-deploy) from the IDE always works fine to clear everything and disconnect from the OPC server.

azoitl (approved):

PR #66, opened 2024-04-01
Build OPC-UA Support in Github Workflow

mhansal:


.github/workflows/cmake-multi-platform.yml
```diff
@@ -65,6 +65,14 @@ jobs:
       if: ${{ matrix.os == 'windows-latest' }}
       run: nuget install boost -Version 1.84.0
 
+    - name: Install OPC-UA
+      run: |
+        git clone https://github.com/open62541/open62541 -b v1.3.4
```
azoitl, line 70: this should definitely be a newer version, see for example the recent update in 4diac Build Environment: https://github.com/eclipse-4diac/4diac-fbe/commit/424637726f2a924cac643ae824fe174bd03565e5

azoitl (changes requested):

.github/workflows/cmake-multi-platform.yml
MartinMelikMerkumians: I think it would make sense to extract the build procedures of other libraries into separate yml files/functions as the workflow gets cluttered already with all the libraries.
But I must admiti didn't take a look on how to do that

.github/workflows/cmake-multi-platform.yml
```diff
@@ -90,7 +102,7 @@ jobs:
         -DFORTE_MODULE_SIGNALPROCESSING=ON
         -DFORTE_MODULE_UTILS=ON
         -DFORTE_TESTS=ON
-        ${{ matrix.os == 'windows-latest' && format('-DFORTE_TESTS_INC_DIRS={0}/boost.1.84.0/lib/native/include', github.workspace) || '' }}
```
MartinMelikMerkumians, line 94: I would put this into he matrix settings to remove the if here

MartinMelikMerkumians: Not sure if this is possible in this situation though

cochicde:
closing as per https://github.com/eclipse-4diac/4diac-forte/pull/228#issuecomment-2452165133

PR #72, opened 2024-04-07
OPC DA: Fix disconnection issue on exit

kumajaya:
Items deletion not properly working on kill and there are always commands left in the queue

### DIFF
--- a/src/com/opc/opceventhandler.cpp
+++ b/src/com/opc/opceventhandler.cpp
@@ -12,6 +12,7 @@
  *   ys guo - Fix opc module compilation errors and deadlock bug
  *   Tibalt Zhao - use stl vector
  *   Ketut Kumajaya - Clear command in queue on exit
+ *                  - Fix disconnection issue on exit
  *******************************************************************************/
 #include "opceventhandler.h"
 #include "../core/devexec.h"
@@ -21,22 +22,37 @@
 
 DEFINE_HANDLER(COpcEventHandler);
 
+CSyncObject COpcEventHandler::mSync;
+forte::arch::CSemaphore COpcEventHandler::mStateSemaphore;
+bool COpcEventHandler::mIsSemaphoreEmpty = true;
+
 COpcEventHandler::TCallbackDescriptor COpcEventHandler::mCallbackDescCount = 0;
 
 COpcEventHandler::COpcEventHandler(CDeviceExecution& paDeviceExecution) : CExternalEventHandler(paDeviceExecution)  {
-  this->start();
-  // Sleep to allow new thread to start
-  CThread::sleepThread(100);
+  if (!isAlive()) {
+    this->start();
+    // Sleep to allow new thread to start
+    CThread::sleepThread(100);
+  }
 }
 
 COpcEventHandler::~COpcEventHandler(){
-  this->end();
+  if (!mCommandQueue.isEmpty()) {
+    resumeSelfSuspend(); //wake-up and execute all commands in queue if exist
+  }
 
-  clearCommandQueue(); // delete command in queue if exist
+  if (isAlive()) {
+    setAlive(false);
+    resumeSelfSuspend();
+    end();
+  }
+
+  clearCommandQueue(); //re-check and delete all commands in queue if still exist
 }
 
 void COpcEventHandler::clearCommandQueue(){
   while(!mCommandQueue.isEmpty()) {
+    // should never have come in here
     ICmd* nextCommand = getNextCommand();
     if(nextCommand != nullptr) {
       DEVLOG_ERROR("erase from command queue[%s]\n", nextCommand->getCommandName());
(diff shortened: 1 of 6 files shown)


kumajaya:
@azoitl I'm still working on this commit in attempt to fix disconnection issue on exit.

kumajaya:
@azoitl Sir, with the last commit, disconnecting/unregister from OPC server on kill always working now. Tested as a client of Matrikon OPC server, Festo Didactic EzOPC, and Moxa AOPC server. No a single crash that make Moxa AOPC restart or EzOPC always listed the client still connected and create a new connection on the next connection. EzOPC usually harder than Matrikon but no problem for both now. This really a come back for OPC DA or maybe better then previous implementation. Please check it.

src/com/opc/opceventhandler.cpp
```diff
@@ -41,18 +66,18 @@ void COpcEventHandler::run(){
...
     while(isAlive()){
-      ICmd* nextCommand = getNextCommand();
-      if(nextCommand != nullptr) {
-        nextCommand->runCommand();
-        delete nextCommand;
+      if (!isAlive()) {
+        break;
       }
-
-      MSG msg;
```
kumajaya, line 50: @azoitl Sir, what exactly this MSG for? Is it related to OPC multithread?

azoitl: Unfortunately I don't know. I was not involved in the original development of our OPC DA infrastructure and never used it myself.

src/com/opc/opceventhandler.h
```diff
@@ -35,11 +38,14 @@ class COpcEventHandler : public CExternalEventHandler, private CThread{
 
     /* functions needed for the external event handler interface */
     void enableHandler() override {
-      start();
+      //nothing call this function
     }
 
     void disableHandler() override {
-      end();
+      if (!isAlive()) {
```
azoitl: in disableHandler at one point you should end the execution of the thread. How would you do that now? Maybe you would like to look at the latest changes in the EventChainExecutionThread.

kumajaya: I found disableHandler executed too early and left some commands in the queue when FORTE was killed, so deleting groups (and items) and disconnecting from the server were never get executed. Actually it's strange delayed the end of the thread fix disconnection issue on kill, I left end of the execution in the destructor.

src/com/opc/opceventhandler.cpp
```diff
@@ -19,16 +22,39 @@
...
+    resumeSelfSuspend();
+    end();
+  }
+
+  clearCommandQueue(); // delete command in queue if exist
+}
+
+void COpcEventHandler::clearCommandQueue(){
+  while(!mCommandQueue.isEmpty()) {
+    // should never have come in here
```
kumajaya, line 55: @azoitl Sir, with a "normal" disableHandler routine, there are always some commands left in the queue on kill. I write this clearCommandQueue actually to debug what command left there.

src/com/opc/opceventhandler.cpp
kumajaya: I deleted it and it was fine so far.

src/com/opc/opceventhandler.h
azoitl: I understand that. However I think you need to ensure that when the command queue is cleared that the thread ends. Maybe you do that and I missed it. However if the queue is empty the thread needs to be stopped. 

While I'm writing this I'm thinking if you can't just clear the queue? Do you need to wait for its execution?

kumajaya: The execution is important to ensure the client properly unregistered from the server. I have tried to just clear the queue and then remove groups and items from COpcConnection destructor but not always working properly, crash more often.

kumajaya: With the last change, basically I did not implement both enableHandler and disableHandler, start and end of the thread controlled by the constructor and the destructor. I use disableHandler only to wake up the thread to empty the command queue. The thread is always in suspend mode after the command queue empty until something wakes it up again.

kumajaya: @azoitl Sir, with the last change now it's really doesn't do anything in enableHandler and disableHandler but double check command queue in the destructor. Groups and items deletion and disconnection on kill still working as the previous change.

kumajaya: I found this is the only multithreaded process inside FORTE so it may need to be handled differently. Maybe we can only start the process in the constructor and stop it in the destructor and not in between via enableHandler/disableHandler.

azoitl: It is not totally in line with the behavior we expect from an external event handler. But I think for now this is a good solutions.

azoitl:
> @azoitl Sir, with the last commit, disconnecting/unregister from OPC server on kill always working now. Tested as a client of Matrikon OPC server, Festo Didactic EzOPC, and Moxa AOPC server. No a single crash that make Moxa AOPC restart or EzOPC always listed the client still connected and create a new connection on the next connection. EzOPC usually harder than Matrikon but no problem for both now. This really a come back for OPC DA or maybe better then previous implementation. Please check it.

This sounds really very good. Thinks for the hard work.

PR #78, opened 2024-04-11
win32: Fix deprecated warning

kumajaya:


src/arch/win32/win32socketinterf.h
```diff
@@ -35,6 +36,16 @@ class CWin32SocketInterface {
...
     CWin32SocketInterface(); //this function is not implemented as we don't want instances of this class
 
+    static LPWSTR fromCharString(const char *paValue) {
+      const size_t size = (strlen(paValue) + 1) * sizeof(wchar_t);
+      wchar_t *buf = new wchar_t[size];
+      buf[size] = L'\0';
+      MultiByteToWideChar(CP_UTF8, 0, paValue, -1, buf, (int)size);
+      const wchar_t *ret = (const wchar_t*)buf;
+      delete[] buf;
+      return (LPWSTR)ret;
```
MartinMelikMerkumians: I think you are returning an invalid pointer here. `ret` points to the same memory as `buf`, but `buf`'s memory is deleted in line 45

kumajaya: In my understanding, I have stored the buf contents to ret constant before deleting it so the returned value still has a valid conversion result.

azoitl: Unfortunately this is not the behavior. you stored in  line 44 in ret the pointer to the memory that you created. However in line 45 you delete that memory. therefore the pointer that you are returning is pointing to freed memory.

kumajaya: @MartinMelikMerkumians @azoitl Now I create a new std::wstring as a result buffer.

kumajaya: I added the following function for reverse conversion locally:

```
    static std::string fromWCharString(const wchar_t *paValue) {
      const int size = WideCharToMultiByte(CP_UTF8, 0, paValue, -1, nullptr, 0, nullptr, nullptr);
      char *buf = new char[size + 1];
      buf[size + 1] = '\0';
      WideCharToMultiByte(CP_UTF8, 0, paValue, -1, buf, size, nullptr, nullptr);
      std::string ret(buf);
      delete[] buf;
      return ret;
    }
```

No problem so far for:
```
DEVLOG_INFO("CWin32SocketInterface: Opening TCP-Server connection at: %s:%d\n", fromWCharString(fromCharString(paIPAddr).c_str()).c_str(), paPort);
```

kumajaya: Finally, a conversion without buffer: https://learn.microsoft.com/id-id/archive/blogs/larryosterman/converting-win32-api-results-to-stdwstring-or-stdstring

kumajaya:
I experimented further because OPC DA easily glitches if string conversion is problematic, there are a lot of WS2S and S2WS functions in it. A working string conversion tested with OPC DA:

```
inline std::wstring forte_stringToWstring(const std::string &paValue) {
  std::wstring ret;
  auto size = MultiByteToWideChar(CP_UTF8, 0, paValue.c_str(), -1, nullptr, 0);
  if (size == 0) {
    return ret;
  }
  ret.resize(size);
  size = MultiByteToWideChar(CP_UTF8, 0, paValue.c_str(), -1, &ret[0], size);
  if (size == 0) {
    ret.resize(0);
    return ret;
  }
  ret.resize(size - 1);
  return ret;
}

inline std::string forte_wstringToString(const std::wstring &paValue) {
  std::string ret;
  auto size = WideCharToMultiByte(CP_UTF8, 0, paValue.c_str(), -1, nullptr, 0, nullptr, nullptr);
  if (size == 0) {
    return ret;
  }
  ret.resize(size);
  size = WideCharToMultiByte(CP_UTF8, 0, paValue.c_str(), -1, &ret[0], size, nullptr, nullptr);
  if (size == 0) {
    ret.resize(0);
    return ret;
  }
  ret.resize(size - 1);
  return ret;
}
```

More portable version almost similar to string conversion in forte_wstring.h:
```
inline std::wstring forte_stringToWstring(const std::string &paVal) {
  auto &converter = std::use_facet<std::codecvt<wchar_t, char, std::mbstate_t>>(std::locale());
  std::mbstate_t mb{};
  std::wstring dst(paVal.size(), '\0');
  const char *from_next;
  wchar_t *to_next;
  converter.in(mb, paVal.data(), paVal.data() + paVal.length(), from_next,
                    dst.data(), dst.data() + dst.length(), to_next);
  // error checking skipped for brevity
  dst.resize(to_next - dst.data());
  return dst;
}

inline std::string forte_wstringToString(const std::wstring &paVal) {
  auto &converter = std::use_facet<std::codecvt<wchar_t, char, std::mbstate_t>>(std::locale());
  std::mbstate_t mb{};
  std::string dst(paVal.size() * converter.max_length(), '\0');
  const wchar_t *from_next;
  char *to_next;
  converter.out(mb, paVal.data(), paVal.data() + paVal.length(), from_next,
                    dst.data(), dst.data() + dst.length(), to_next);
  // error checking skipped for brevity
  dst.resize(to_next - dst.data());
  return dst;
}
```

src/arch/forte_stringFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+ * Contributors:
+ *  Ketut Kumajaya - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+#ifndef SRC_ARCH_FORTE_STRINGFUNCTIONS_H_
+#define SRC_ARCH_FORTE_STRINGFUNCTIONS_H_
+
+#include <string>
+#include <codecvt>
+
+inline const std::wstring forte_stringToWstring(const std::string &paVal) {
```
MartinMelikMerkumians: Returning const values has more drawbacks than benefits starting with C++11, as rvalue usage is impeded
rationale here: https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value

I would recommend to change this to a non-const return value.

src/arch/forte_stringFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+  const char *from_next;
+  wchar_t *to_next;
+  converter.out(mb, &paVal[0], &paVal[paVal.size()], from_next,
+                    &dst[0], &dst[dst.size()], to_next);
+  // error checking skipped for brevity
+  dst.resize(to_next - &dst[0]);
+  return dst;
+}
+
+inline const std::string forte_wstringToString(const std::wstring &paVal) {
```
MartinMelikMerkumians: Returning const values has more drawbacks than benefits starting with C++11, as rvalue usage is impeded
rationale here: https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value

I would recommend to change this to a non-const return value.

src/arch/forte_stringFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+  auto &converter = std::use_facet<std::codecvt<wchar_t, char, std::mbstate_t>>(std::locale());
+  std::mbstate_t mb{};
+  std::string dst(paVal.size() * converter.max_length(), '\0');
+  const wchar_t *from_next;
+  char *to_next;
+  converter.out(mb, &paVal[0], &paVal[paVal.size()], from_next,
+                    &dst[0], &dst[dst.size()], to_next);
```
MartinMelikMerkumians: I am not sure if passing the address of the wstring's index operator is correct here, as, as far as I know, there is no guarantee that two calls are pointing to a consecutive memory of glyphs.
From my point of view, you need to call data() on the wstring to get this consecutive memory.

I would also ask you to provide some test cases for the new functions, so in case of code changes automatic tests are in place already.

MartinMelikMerkumians (changes requested):

kumajaya:
Actually I couldn't reproduce the previous build error locally but found boost's unit test framework failed: "fatal error std::bad_cast: std::bad_cast" and sure the error came from std::use_facet. I'm not sure about musl locale support, it could be the problem comes from musl but the conversion function should not fail because musl is my main compiler for aarch64 targets, so I decided to switch to mbsrtowcs and wcsrtombs.

kumajaya:
> Actually I couldn't reproduce the previous build error locally but found boost's unit test framework failed: "fatal error std::bad_cast: std::bad_cast" and sure the error came from std::use_facet. I'm not sure about musl locale support, it could be the problem comes from musl but the conversion function should not fail because musl is my main compiler for aarch64 targets, so I decided to switch to mbsrtowcs and wcsrtombs.

Error from my side, musl is right. Restore back previous codes with fix.

kumajaya:
I tested this string/wstring conversion functions with OPC DA and no issue so far. I'm sure will be no problem for plain ASCII text but need special handle for Unicode per as my example in boost unit test. Please check it, I may be wrong and need some fix.

tests/arch/forte_stringFunctions_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+    const std::string testString = "4diac 4 \xe3\x81\xaf\xe4\xbb\x8a\xe3\x81\xbe\xe3\x81\xa7\xef\xbc\x81"; // u8"4diac 4 は今まで！"
+    const std::wstring resultString = L"4diac 4 \xe3\x81\xaf\xe4\xbb\x8a\xe3\x81\xbe\xe3\x81\xa7\xef\xbc\x81"; // L"4diac 4 は今まで！"
+#else
+    // Set the new and save the previous locale
+    // BUG: trigger std::runtime_error: locale::facet::_S_create_c_locale name not valid with musl
+    std::locale prev_loc = std::locale::global(std::locale("C.UTF-8"));
+    const std::string testString = u8"4diac 4 は今まで！"; // make sure to use UTF-8 text editor with Unicode support
+    const std::wstring resultString = L"4diac 4 は今まで！";
+#endif
+#if 0
```
MartinMelikMerkumians: correct me if I am wrong, but #if 0 means that this code will never be compiled, so it is essentially commented code, which should not be checked in

tests/arch/forte_stringFunctions_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+    std::cout << "wstringToString: " << testString.c_str() << std::endl;
+    std::cout << "wstringToString: " << forte_wstringToString(resultString).c_str() << std::endl << std::endl;
+#endif
+
+    BOOST_CHECK_EQUAL(testString.length(), forte_wstringToString(resultString).length());
```
MartinMelikMerkumians: as a suggestion, I nowadays only use BOOST_TEST, as it usually results in better output if a test fails, like getting the exact numbers or strings.
You can also check C-strings directly for equality (as compare) when using BOOST_TEST
https://www.boost.org/doc/libs/1_85_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/strings.html

tests/arch/forte_stringFunctions_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+    const std::string resultString = "4diac 4 \xe6\xb0\xb8\xe8\xbf\x9c\xef\xbc\x81"; // u8"4diac 4 永远！"
+#else
+    // Set the new and save the previous locale
+    // BUG: trigger std::runtime_error: locale::facet::_S_create_c_locale name not valid with musl
+    std::locale prev_loc = std::locale::global(std::locale("C.UTF-8"));
+    const std::wstring testString = L"4diac 4 永远！"; // make sure to use UTF-8 text editor with Unicode support
+    const std::string resultString = u8"4diac 4 永远！";
+#endif
+#if 0
+    std::cout << "stringToWstring: " << forte_wstringToString(testString).c_str() << std::endl;
```
MartinMelikMerkumians: see comments above

tests/arch/forte_stringFunctions_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+    std::cout << "stringToWstring: " << resultString.c_str() << std::endl << std::endl;
+#endif
+
+    BOOST_CHECK_EQUAL(testString.length(), forte_stringToWstring(resultString).length());
```
MartinMelikMerkumians: see comments above

MartinMelikMerkumians (changes requested):

tests/arch/forte_stringFunctions_test.cpp
kumajaya: It's for debugging when I need it. I deleted it.

tests/arch/forte_stringFunctions_test.cpp
kumajaya: I switched to BOOS_TEST.

tests/arch/forte_stringFunctions_test.cpp
kumajaya: I switched to BOOS_TEST.

MartinMelikMerkumians (approved):

azoitl (approved):

PR #79, opened 2024-04-12
modbus: Mute build warning

kumajaya:


src/com/modbus/modbuslayer.cpp
```diff
@@ -171,7 +171,7 @@ EComResponse CModbusComLayer::processInterrupt(){
     switch (mConnectionState){
       case e_Connected: {
         CIEC_ANY **apoRDs = mFb->getRDs();
-        unsigned int nrRDs = mFb->getNumRD();
+        unsigned int nrRDs = static_cast<unsigned int>(mFb->getNumRD());
```
azoitl: The correcter and in our current code style solution would be to change the data type of `nrRDs`, `dataIndex`, and `i` to `size_t`.

kumajaya: I just though the rest of the code is not capable to handle size_t.

azoitl: what I see the code would be correcter with size_t but maybe a bit more changes would be needed.

kumajaya: > what I see the code would be correcter with size_t but maybe a bit more changes would be needed.

I just mute the warning for now.

azoitl (approved):

PR #80, opened 2024-04-12
lua: Mute build warning

kumajaya:


azoitl:
While I'm very happy that you spend your time improving the code quality of 4diac FORTE I have some issues with this PR. The main reason is that the lua infrastructure is currently broken and needs a heavy rework. Especially the places with warnings should be changed bigger. Your patch would hide that places. Therefore I would prefer not to merge it.

kumajaya:
> While I'm very happy that you spend your time improving the code quality of 4diac FORTE I have some issues with this PR. The main reason is that the lua infrastructure is currently broken and needs a heavy rework. Especially the places with warnings should be changed bigger. Your patch would hide that places. Therefore I would prefer not to merge it.

You're absolutely right. Lua has a potential for FORTE but currently broken.

PR #91, opened 2024-04-19
Fix FORTE test with musl or uclibc

kumajaya:


kumajaya:
I build boost's unit test framework as below:
1. Run `./bootstrap.sh` to build b2 binary
2. Edit project-config.jam and set gcc configuration: `using gcc : aarch64 : /home/user/toolchain/aarch64--musl--bleeding-edge-2024.02-1/bin/aarch64-linux-g++ : <linkflags>-static-libgcc <linkflags>-static-libstdc++ ;`
3. Run `./b2 --with-test link=shared --prefix=/home/user/forte/install/aarch64-musl install`

Test result:
```
~$ ./forte_test
Running 1337 test cases...
ERROR: T#58996127983621: REPLACE called with P equal or less to 0!
ERROR: T#58996128605658: REPLACE called with P equal or less to 0!
ERROR: T#58996129181344: REPLACE called with P equal or less to 0!
ERROR: T#58996131306473: DELETE called with L equal or less than 0!
ERROR: T#58996131852010: DELETE called with L equal or less than 0!
ERROR: T#58996132428862: DELETE called with L equal or less than 0!
ERROR: T#58996132977825: DELETE called with P equal or less than 0!
ERROR: T#58996133535288: DELETE called with P equal or less than 0!
ERROR: T#58996134079621: DELETE called with P equal or less than 0!
ERROR: T#58996134627232: DELETE called with delete length exceeding the length of string!
ERROR: T#58996164373214: INSERT called with P exceeding input string length!
ERROR: T#58996164939788: INSERT called with P equal or lower than 0!
ERROR: T#58996165511603: INSERT called with P equal or lower than 0!
ERROR: T#58996166062918: INSERT called with P equal or lower than 0!
/home/user/forte/sources/org.eclipse.4diac.forte/tests/core/iec61131_functionstests.cpp(1948): error: in "IEC61131_functions/func_day_of_week": check static_cast<CIEC_ULINT::TValueType>(weekday) == 1 has failed [18446744073294306616 != 1]
unknown location(0): fatal error: in "internal_vars/sampleInteralVarList": memory access violation at address: 0x0: no mapping at fault address
/home/user/forte/sources/org.eclipse.4diac.forte/tests/core/internalvartests.cpp(86): last checkpoint: "sampleInteralVarList" test entry

*** 2 failures are detected in the test module "FORTE_TESTS"
```

kumajaya:
MSVC build test result:
```
Running 1333 test cases...
ERROR: T#60701405261700: REPLACE called with P equal or less to 0!
ERROR: T#60701405677700: REPLACE called with P equal or less to 0!
ERROR: T#60701405995700: REPLACE called with P equal or less to 0!
ERROR: T#60701406291100: DELETE called with L equal or less than 0!
ERROR: T#60701406514100: DELETE called with L equal or less than 0!
ERROR: T#60701406793400: DELETE called with L equal or less than 0!
ERROR: T#60701407072700: DELETE called with P equal or less than 0!
ERROR: T#60701407306900: DELETE called with P equal or less than 0!
ERROR: T#60701407507800: DELETE called with P equal or less than 0!
ERROR: T#60701407733000: DELETE called with delete length exceeding the length of string!
ERROR: T#60701408654600: INSERT called with P exceeding input string length!
ERROR: T#60701408887900: INSERT called with P equal or lower than 0!
ERROR: T#60701409083400: INSERT called with P equal or lower than 0!
ERROR: T#60701409264500: INSERT called with P equal or lower than 0!
unknown location(0): fatal error: in "internal_vars/sampleInteralVarList": memory access violation
C:\msys64\home\ketut\forte\sources\4diac-forte\tests\core\internalvartests.cpp(86): last checkpoint: "sampleInteralVarList" test entry

*** 1 failure is detected in the test module "FORTE_TESTS"
```

azoitl:
@jwalt can you please have a look on this PR as you have more experience with musl.

kumajaya:
Actually this is a copy of https://github.com/eclipse-4diac/4diac-forte/blob/2a67f64eb454d0424618c01cc340c20417d9fbdb/src/CMakeLists.txt#L299 to create a working fully static FORTE test binary with musl or uclibc. With musl or uclibc compiler, we have to manually add MUSL or UCLIBC define from cmake to enable it. The rest changes to make FORTE main binary for posix target respect FORTE_TEST_CODE_COVERAGE_ANALYSIS value.

jwalt:
There are two parts of this patch. The part where you put the coverage analysis behind a switch looks fine and sensible. 

What I dislike is the unconditional use of static linking with musl or uclibc. Defining MUSL or UCLIBC this way is  undocumented (no cmake cache option) and too obscure, even if this is already part of our cmake code.

Please add an

```
option(FORTE_STATIC_EXECUTABLE OFF "Force static linking of forte binary")
```

or something like that and then use that setting to enable these flags. While you are at it, it would be nice if you adjusted the other occurrence of these variables as well.

jwalt (changes requested):
There are two parts of this patch. The part where you put the coverage analysis behind a switch looks fine and sensible. 

What I dislike is the unconditional use of static linking with musl or uclibc. Defining MUSL or UCLIBC this way is  undocumented (no cmake cache option) and too obscure, even if this is already part of our cmake code.

Please add an

```
option(FORTE_STATIC_EXECUTABLE OFF "Force static linking of forte binary")
```

or something like that and then use that setting to enable these flags. While you are at it, it would be nice if you adjusted the other occurrence of these variables as well.

src/CMakeLists.txt
```diff
@@ -293,10 +293,14 @@ TARGET_LINK_LIBRARIES (FORTE_LITE ${LINK_LIBRARY})
 
 list(JOIN link_flags " " link_flags_string)
 
+if("${FORTE_ARCHITECTURE}" STREQUAL "Posix")
```
jwalt: This can be just `if (FORTE_ARCHITECTURE STREQUAL "Posix")`. But actually I think we don't need this guard at all. If people want to do this on win32 or other platforms, let them do so.

src/CMakeLists.txt
```diff
@@ -293,10 +293,14 @@ TARGET_LINK_LIBRARIES (FORTE_LITE ${LINK_LIBRARY})
...
+  set(FORTE_LINK_STATIC_LIBC_LIBSTDCPP OFF CACHE BOOL "Force FORTE linking statically against libc and lib stdc++")
```
jwalt: Since this enables the `-static` flag which affects all libraries, just call this `FORTE_LINK_STATIC`

jwalt (commented):
This looks very nice already. only minor comments remain. Thanks!

src/CMakeLists.txt
kumajaya: I chose no guard at all.

src/CMakeLists.txt
kumajaya: Agreed.

jwalt (approved):

azoitl (approved):

PR #95, opened 2024-04-22
Mute CIEC_STRING build warning

kumajaya:


azoitl:
I think I wrote about this already once: In the case of the OPCUA_MGR the warnings are showing us that here changes will be needed. The real problem that needs to be fixed and is on our todo list is that the `mAdditionalParams` should be changed from `CIEC_STRING` to `std::string`. Muting the warnings is just hiding a problem but not fixing it. Therefore I would prefer to not have it muted.

kumajaya:
> I think I wrote about this already once: In the case of the OPCUA_MGR the warnings are showing us that here changes will be needed. The real problem that needs to be fixed and is on our todo list is that the `mAdditionalParams` should be changed from `CIEC_STRING` to `std::string`. Muting the warnings is just hiding a problem but not fixing it. Therefore I would prefer to not have it muted.

Just remove the OPCUA_MGR part.

azoitl (approved):

PR #98, opened 2024-04-23
flush devlog after every message

ernstblechaPT:
The 4diac ide console also buffers output also for stderr, flush devlog messages after every message

### DIFF
--- a/src/arch/devlog.cpp
+++ b/src/arch/devlog.cpp
@@ -19,7 +19,6 @@
 # include "forte_printer.h"
 # include "../core/utils/criticalregion.h"
 # include "forte_architecture_time.h"
-# include <cstdio>
 # include <cstdlib>
 # include <cstdarg>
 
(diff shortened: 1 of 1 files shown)


ernstblechaPT:
@azoitl 
 - Should I move the fflush into a `#ifdef WIN` block?
 - Should I change the behavior so that only beginning from loglevel ERROR messages are sent to stderr?

azoitl:
> @azoitl
> 
>     * Should I move the fflush into a `#ifdef WIN` block?

As it is solving a win problem I think so

> 
>     * Should I change the behavior so that only beginning from loglevel ERROR messages are sent to stderr?

as you are on it yes please.

azoitl (approved):

ernstblechaPT:
for stacktraces the change was not needed, std::endl always flushes anyway
changed output from fprintf to iostream so output is only generated one way

azoitl (approved):

PR #100, opened 2024-04-25
DEVLOG: Option for human readable time

kumajaya:


src/arch/devlog.cpp
```diff
@@ -98,7 +100,13 @@ void logMessage(E_MsgLevel paLevel, const char *paMessage, ...) {
 
 void printLogMessage(E_MsgLevel paLevel, const char *paMessage) {
 #ifndef __ZEPHYR__
-  (paLevel == E_MsgLevel::Error ? std::cerr : std::cout) << scLogLevel[static_cast<int>(paLevel)] << ": " << getNanoSecondsMonotonic() << ": " << paMessage;
+#ifdef FORTE_LOGGER_TIME_IN_DT
```
azoitl: sorry to be so direct but this is unreadable code. Can you please create a method like `getTimeString` and have then two C++ files:
  - `plainTimeString.cpp` implementing `getTimeString` with  simple time string 
  - `readableTimeString.cpp` implementing `getTimeString` using func_NOW

in CMakeList.txt you can then add the right cpp file based on the CMake option. 

So the code here does not need another define.

azoitl (changes requested):

kumajaya:
@azoitl Sir, I'm not sure where to put getRealtimeString function prototype and I named it getRealtimeString because it is also possible to get a readable monotonic time string.

src/arch/readableTimeString.cpp
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Ketut Kumajaya - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+#include "../core/iec61131_functions.h"
+
+std::string getRealtimeString() {
+  std::string time(27, '\0'); // DT#1970-01-01-00:00:00.000
```
azoitl: please use the `getToStringBufferSize()` of `CIEC_DATE_AND_TIME` so that you can be save against any changes in the format.

src/arch/forte_architecture_time.h
```diff
@@ -76,4 +77,6 @@ uint_fast64_t getNanoSecondsRealtime();
 
 uint_fast64_t trackFakeForteTime();
 
+std::string getRealtimeString();
```
azoitl: please move this declaration into `devlog.cpp` currently we only need it there.

azoitl (changes requested):

azoitl:
> @azoitl Sir, I'm not sure where to put getRealtimeString function prototype and I named it getRealtimeString because it is also possible to get a readable monotonic time string.

As commented in the code review I think the best place is the devlog.cpp. Because currently we only need it there.

src/arch/forte_architecture_time.h
kumajaya: Done.

src/arch/readableTimeString.cpp
kumajaya: I'm so blind. Fixed.

src/arch/devlog.cpp
kumajaya: Done.

azoitl (approved):

src/arch/devlog.cpp
```diff
@@ -98,7 +100,7 @@ void logMessage(E_MsgLevel paLevel, const char *paMessage, ...) {
 
 void printLogMessage(E_MsgLevel paLevel, const char *paMessage) {
 #ifndef __ZEPHYR__
-  (paLevel == E_MsgLevel::Error ? std::cerr : std::cout) << scLogLevel[static_cast<int>(paLevel)] << ": " << getNanoSecondsMonotonic() << ": " << paMessage;
+  (paLevel == E_MsgLevel::Error ? std::cerr : std::cout) << scLogLevel[static_cast<int>(paLevel)] << ": " << getRealtimeString().c_str() << ": " << paMessage;
```
ernstblechaPT: Can you explain why you are explicitly calling .c_str() for the output? Concatinating the std::string object should work out of the box.

ernstblechaPT (changes requested):

src/arch/devlog.cpp
kumajaya: Thank you. Just fixed it.

ernstblechaPT (approved):

azoitl (approved):

PR #101, opened 2024-04-28
Fix gcc build warning

kumajaya:


src/core/datatypes/forte_string.cpp
```diff
@@ -68,7 +68,7 @@ int CIEC_STRING::fromString(const char *paValue) {
         i += 2;
         continue;
       } else {
-        char specialSymbol;
+        char specialSymbol = '\0';
```
azoitl: `\0` is dangerouse here. Because with that you end the string earlier then it should. Do we have a more neutral character?

src/core/datatypes/forte_string.cpp
```diff
@@ -68,7 +68,7 @@ int CIEC_STRING::fromString(const char *paValue) {
...
         switch (inputString[i + 1]) {
```
azoitl: as you are cleaning up here. Should we move the switch case into an own method? this would make the code much more readable.

azoitl (changes requested):

src/core/datatypes/forte_string.cpp
MartinMelikMerkumians: I wouldn't potentially return a valid string. The method should terminate and indicate an error. After all the string was not valid

MartinMelikMerkumians (changes requested):

azoitl (approved):

src/core/datatypes/forte_string.cpp
```diff
@@ -69,37 +104,9 @@ int CIEC_STRING::fromString(const char *paValue) {
...
-          case 't':
-            specialSymbol = '\t';
-            break;
-          case '$':
-            specialSymbol = '$';
-            break;
-          default:
-            DEVLOG_ERROR("Unknown escape symbol %c encountered", inputString[i + 1]);
+        if (getSpecialSymbol(inputString[i + 1], specialSymbol) != -1) {
+          unescapedString += specialSymbol; // append escaped char
```
MartinMelikMerkumians, line 108: As mentioned before, shouldn't we abort with an error here? @azoitl 
Now we are just not adding the invalid escaped symbol, thereby silently ignoring an error, right?

MartinMelikMerkumians (changes requested):

src/core/datatypes/forte_string.cpp
azoitl: True I missed that. However, giving it a second thought I'm not perfectly sure what should happen here. I had a look in IEC 61131-3 and there nothing is stated about invalid $ escapings. Does anyone know what other languages are doing. Should we just ignore the $ then. I think this is what happens in C or?

MartinMelikMerkumians: The compiler gives you an error what I have seen on StackOverflow

kumajaya: The previous logic possibly add uninitialized specialSymbol char that trigger compiler warning. I'm also not sure what should happen here.

MartinMelikMerkumians: Don't get me wrong, it is an improvement. I am. Just not sure if we should create a valid string in such a case

azoitl: @kumajaya first of all thx for your patience with us. With your work you are definitely improving 4diac FORTE. But as you are poking on a rather important component and as you are changing it, we think it would be good that we get it right this time.

kumajaya: @azoitl No problem, sir! I found this great project and started loving it from the beginning because it seemed to suit my needs. Just a little help from my side, I always take compiler warnings seriously and try to understand why they occur but of course I always need good guidance from people who understand more here.

MartinMelikMerkumians: @kumajaya it's good you fix compiler warnings, it is important and I fixed a bunch myself, but not always high on our priorities ;)

azoitl:
As there is no activity on this PR for quite some time and the 4diac FORTE code base has significantly change I'm closing this PR. @kumajaya if you want to bring this fixes to the latest code base feel free to open a new updated PR.

PR #102, opened 2024-04-28
forte_malloc: Check the size argument does not exceed PTRDIFF_MAX

kumajaya:
Suppress GCC warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807

Marked as RESOLVED WONTFIX: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85783

### DIFF
--- a/src/arch/genfortealloc.h
+++ b/src/arch/genfortealloc.h
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2010 - 2013 ACIN, nxtcontrol GmbH, fortiss GmbH
+ *               2024 Samator Indo Gas
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -9,11 +10,13 @@
  * Contributors:
  *   Alois Zoitl, Stanislav Meduna, Monika Wenger
  *    - initial API and implementation and/or initial documentation
+ *   Ketut Kumajaya - check the size argument does not exceed PTRDIFF_MAX
  *******************************************************************************/
 #ifndef GENFORTEALLOC_H_
 #define GENFORTEALLOC_H_
 
 #include <stdlib.h>
+#include <cstdint>
 
 
 inline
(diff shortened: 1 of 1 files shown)


src/arch/genfortealloc.h
```diff
@@ -14,6 +14,7 @@
 #define GENFORTEALLOC_H_
 
 #include <stdlib.h>
+#include <stdint.h>
```
MartinMelikMerkumians: I think we should use cstdint instead of stdint.h (see https://stackoverflow.com/questions/13642827/cstdint-vs-stdint-h)

Most likely you need to add the std namespace specifier before PTRDIFF_MAX (so std::PTRDIFF_MAX)

MartinMelikMerkumians: You should also add your copyright @kumajaya as you add a significant portion of code, compared to the existing one

MartinMelikMerkumians (changes requested):

src/arch/genfortealloc.h
```diff
@@ -23,7 +26,15 @@ void forte_free(void *paData){
 
 inline
 void *forte_malloc(size_t paSize){
+#if defined(__GNUC__) && !defined(__clang__)
+  // BUG RESOLVED WONTFIX: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85783
+  if (paSize <= PTRDIFF_MAX) {
```
MartinMelikMerkumians, line 31: Does `std::PTRDIFF_MAX` not work? If it does work with `std::` please add it, as assuming that std symbols are imported is not recommended according to StackOverflow

MartinMelikMerkumians (changes requested):

src/arch/genfortealloc.h
kumajaya: Not work. Defined in header cstdint https://en.cppreference.com/w/cpp/types/climits

MartinMelikMerkumians (approved):

src/arch/genfortealloc.h
MartinMelikMerkumians: Thanks for checking

azoitl (approved):

PR #106, opened 2024-05-03
win32: Forte as a Windows service

kumajaya:
Hijack main entry point and replaced by wmain for Windows service entry. Keep all files header description including the license statement. Not sure about the license compatibility to Eclise's EPL.

Based on: https://github.com/microsoftarchive/msdn-code-gallery-microsoft/tree/master/OneCodeTeam/A%20basic%20Windows%20service%20in%20C%2B%2B%20(CppWindowsService)%202

FORTE as a Windows service current status:
* Run as a Windows service or as a usual console application
* Self service install and remove via terminal console
* As a service, accept command line arguments via registry key
* As a service, stdout and stderr redirected to forte.log file and can be disable
* We can build with MSVC or MINGW, default unicode system call per as Microsoft recomendation https://learn.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings


kumajaya:
@azoitl With this changes, main.cpp looks messy now. How if I keep SampleService.cpp as a separate file and then create some dummies function call (endForte and _main) in test/arch/win32/* to make forte test build happy? Or how to tell forte test build to exclude some files?

kumajaya:
Ready for review, no more crashes listed in Windows Logs. Need an opinion on the dummies function in forte_boost_tester.cpp, is there a better solution?

kumajaya:
> Ready for review, no more crashes listed in Windows Logs. Need an opinion on the dummies function in forte_boost_tester.cpp, is there a better solution?

Just found the solution. I will push update later today.

kumajaya:
Finish. Change trackable per commit.

kumajaya:
Run service on LocalSystem account to make it have a permission to write log file, for compatibility to Windows 7. Use unicode version of freopen to reduce string conversion.

kumajaya:
Change the logging redirection using a more portable and simple way by manipulating rdbuf.

azoitl:
as this is by now outdated and we haven't sorted out the license problem I would close it for now.

PR #107, opened 2024-05-04
readableTimeString: Exclude the last NULL char

kumajaya:
I found this bug with directed stdout/stderr to file. Notepad++ shown excess of NULL

### DIFF
--- a/src/arch/readableTimeString.cpp
+++ b/src/arch/readableTimeString.cpp
@@ -14,7 +14,7 @@
 
 std::string getRealtimeString() {
   CIEC_DATE_AND_TIME dt = func_NOW();
-  std::string str(dt.getToStringBufferSize(), '\0');
+  std::string str(dt.getToStringBufferSize() - 1, '\0'); // -1 for exclude the last \0
   dt.toString(str.data(), str.size());
   return str;
 }


src/arch/readableTimeString.cpp
```diff
@@ -14,7 +14,7 @@
 
 std::string getRealtimeString() {
   CIEC_DATE_AND_TIME dt = func_NOW();
-  std::string str(dt.getToStringBufferSize(), '\0');
+  std::string str(dt.getToStringBufferSize() - 1, '\0');
   dt.toString(str.data(), str.size());
```
MartinMelikMerkumians, line 18: Perhaps I am wrong, but as I read this https://cplusplus.com/reference/string/string/data/ you shouldn't write on data in the first place

I propose to change the code to
```
char dateCString[sizeof("DT#1970-01-01-00:00:00.000")];
func_NOW().toString(dateCString, sizeof(dateCString));
std::string dateString(dateCString, sizeof(dateCString));
return dateString;
```

MartinMelikMerkumians (changes requested):

src/arch/readableTimeString.cpp
MartinMelikMerkumians: In a follow up step, the fixed toString buffer sizes, accessed by getToStringBufferSize should be put into a template, similar to numeric_limits or added as a constexpr static class variable, so that it can be used in constexpr use cases like here.
FYI @azoitl

kumajaya: > Perhaps I am wrong, but as I read this https://cplusplus.com/reference/string/string/data/ you shouldn't write on data in the first place
> 
> I propose to change the code to
> 
> ```
> char dateCString[sizeof("DT#1970-01-01-00:00:00.000")];
> func_NOW().toString(dateCString, sizeof(dateCString));
> std::string dateString(dateCString, sizeof(dateCString));
> return dateString;
> ```

How about if I write it:
```
  CIEC_DATE_AND_TIME dt = func_NOW();
  size_t length = dt.getToStringBufferSize() - 1;
  std::vector<char> str(length);
  dt.toString(str.data(), str.size());
  return std::string(str.data(), str.size());
```
or:
```
  CIEC_DATE_AND_TIME dt = func_NOW();
  size_t length = dt.getToStringBufferSize() - 1;
  std::vector<char> str(length);
  dt.toString(&str[0], str.size());
  return std::string(&str[0], str.size());
```
or maybe better:
```
  CIEC_DATE_AND_TIME dt = func_NOW();
  size_t length = dt.getToStringBufferSize() - 1;
  std::vector<char> str(length);
  dt.toString(&str[0], str.size());
  return std::string(str.data(), str.size());
```

I think the original code is valid since C++17: https://en.cppreference.com/w/cpp/string/basic_string/data

kumajaya:
Change getRealtimeString() to getRealtimeString().c_string() in devlog.cpp will remove the excess '\0' but in my opinion the source of problem from here.

src/arch/readableTimeString.cpp
MartinMelikMerkumians: Thanks for the link, you are right, my resource was outdated it seems, so your original version is fine

MartinMelikMerkumians (approved):

azoitl (approved):

PR #111, opened 2024-05-13
OPC DA: Fix memory leaks

kumajaya:
Not using an advance profiler but FORTE with OPC DA module only enabled for read 4 items Moxa OPC server, memory consumption on Task Manager slowly increase every time a new OPC data arrived. For about 16 hours, memory consumption increase from < 3MB to > 5MB. This issue only appears to occurs on Windows 7, no problem on Windows 11.

### DIFF
--- a/src/com/opc/opccomlayer.cpp
+++ b/src/com/opc/opccomlayer.cpp
@@ -38,9 +38,13 @@ COpcComLayer::~COpcComLayer(){
   for(auto &it :mFBInputVars){
     delete it;
   }
+  //clear vector and remove it from memory
+  TOpcProcessVarList().swap(mFBInputVars);
   for(auto &it :mFBOutputVars){
     delete it;
   }
+  //clear vector and remove it from memory
+  TOpcProcessVarList().swap(mFBOutputVars);
 }
 
 EComResponse COpcComLayer::sendData(void *paData, unsigned int paSize){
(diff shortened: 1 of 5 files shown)


src/com/opc/opccomlayer.cpp
```diff
@@ -36,11 +36,19 @@ COpcComLayer::COpcComLayer(CComLayer* paUpperLayer, CBaseCommFB* paComFB) :
 COpcComLayer::~COpcComLayer(){
   DEVLOG_DEBUG("delete OpcComLayer: [0x%lX][%s]\n", this, mOpcGroupName.c_str());
   for(auto &it :mFBInputVars){
-    delete it;
+    if(nullptr != it){
```
azoitl: according to the specification delete shall correctly handle nullptr. Therefore the nullptr check is not needed.

src/com/opc/opcconnection.cpp
```diff
@@ -52,8 +52,34 @@ int COpcConnection::removeGroupCallbackDesc(const std::string& paGroupName){
     if(mOpcGroupMapList[i]->mGroupName == paGroupName){
       mEventHandler->removeComCallback(mOpcGroupMapList[i]->mCallbackDesc);
       DEVLOG_INFO("erase from m_lOpcGroupMapList %d[%s]\n", i,paGroupName.c_str());
-      delete mOpcGroupMapList[i];
+      //remove group's read item list first
+      for(auto &readItem : mOpcGroupMapList[i]->mReadItemsList){
+        if(nullptr != readItem){
+          delete readItem;
+          readItem = nullptr;
```
azoitl: as you are clearing the list with the swap below you don't need to set the item to nullptr. This applies to all other code below as well.

azoitl (changes requested):

azoitl (approved):

azoitl (approved):

PR #113, opened 2024-05-18
Cleanup of raw pointers in CNodePairInfo

cochicde:
Just removing some raw pointers and using `unique_ptr` or objects directly.

It makes memory handling safer and keep destructors empty.

The major change is in `class CNodePairInfo`. The rest is refactoring the code to make it work

### DIFF
--- a/src/com/opc_ua/opcua_action_info.h
+++ b/src/com/opc_ua/opcua_action_info.h
@@ -46,11 +46,26 @@ class CActionInfo {
     /**
      * 2-Tuple to reference a specific node in an OPC UA server composed of the node ID and browsepath of the node
      */
-    struct CNodePairInfo {
+    class CNodePairInfo {
+      public:
         CNodePairInfo(UA_NodeId *paNodeId, const std::string &paBrowsePath) :
-            mNodeId(paNodeId), mBrowsePath(paBrowsePath) {
+            mNodeId(paNodeId, UA_NodeId_delete), mBrowsePath(paBrowsePath) {
         }
-        UA_NodeId *mNodeId;
+
+        UA_NodeId *getNodeId() const {
+          return mNodeId.get();
+        }
+
+        void setNodeId(UA_NodeId * paNodeId){
+          mNodeId.reset(paNodeId);
+        }
+
+        const std::string& getBrowsePath() const {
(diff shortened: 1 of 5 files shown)


src/com/opc_ua/opcua_action_info.h
```diff
@@ -46,11 +46,26 @@ class CActionInfo {
...
         CNodePairInfo(UA_NodeId *paNodeId, const std::string &paBrowsePath) :
-            mNodeId(paNodeId), mBrowsePath(paBrowsePath) {
+            mNodeId(paNodeId, UA_NodeId_delete), mBrowsePath(paBrowsePath) {
+        }
+
+        UA_NodeId *getNodeId() const {
+          return mNodeId.get();
+        }
+
+        void setNodeId(UA_NodeId * paNodeId){
```
m-meingast: There is a blank space between the asterisk and the parameter name.
```suggestion
        void setNodeId(UA_NodeId *paNodeId){
```

src/com/opc_ua/opcua_local_handler.cpp
```diff
@@ -906,9 +908,9 @@ UA_StatusCode COPC_UA_Local_Handler::initializeCreateNode(CActionInfo &paActionI
   //The main process is done in the execution
   UA_StatusCode retVal = UA_STATUSCODE_GOOD;
 
-  CSinglyLinkedList<CActionInfo::CNodePairInfo*>::Iterator it = paActionInfo.getNodePairInfo().back();
+  auto& instance = paActionInfo.getNodePairInfo().back();
```
m-meingast: Maybe we should use a more explicit variable name instead of instance, e.g. nodePair?

m-meingast (approved):
Looks good to me.

src/com/opc_ua/opcua_action_info.h
cochicde: changed

src/com/opc_ua/opcua_local_handler.cpp
cochicde: you're right. Changed

m-meingast (approved):

azoitl (approved):

PR #114, opened 2024-05-25
Add test to ctf tracer and helper classses 

cochicde:
This PR is to have a base for testing the ctf traces. It has classes representing all possible events being traced, a way of reading the trace files, and a minimal example of comparison of a small FB network.

The test is very simple, but the rest is what's needed for further improvement of the trace module and for more complex analysis of traces.

### DIFF
--- a/tests/core/trace/EventMessage.cpp
+++ b/tests/core/trace/EventMessage.cpp
@@ -6,6 +6,10 @@
 // EventMessage methods //
 // ******************** //
 
+EventMessage::EventMessage(const std::string& paEventType, std::unique_ptr<AbstractPayload> paPayload, int64_t paTimestamp) : 
+  mEventType(paEventType), mPayload(std::move(paPayload)), mTimestamp(paTimestamp){
+}
+
 EventMessage::EventMessage(const bt_message* message)  {
     
   // Borrow the event message's event and its class
@@ -58,64 +62,73 @@ std::string EventMessage::getTimestampString() const {
 }
 
 bool EventMessage::operator==(const EventMessage& other) const {
-  return mEventType == other.mEventType  
-    && (mPayload == nullptr) == (other.mPayload == nullptr)
-    && (mPayload != nullptr ? *mPayload == *other.mPayload : true);
+  return mEventType == other.mEventType  && *mPayload == *other.mPayload;
 }
 
 // *********************** //
 // AbstractPayload Methods //
 // *********************** //
+AbstractPayload::AbstractPayload(const std::string& paTypeName, const std::string& paInstanceName) 
+  : mTypeName(paTypeName), mInstanceName(paInstanceName)  {
+}
 
 AbstractPayload::AbstractPayload(const bt_field* field) {
   auto const typeNameField = bt_field_structure_borrow_member_field_by_name_const(field, "typeName");
   auto const instanceNameField = bt_field_structure_borrow_member_field_by_name_const(field, "instanceName");
   
   mTypeName = bt_field_string_get_value(typeNameField);
-  instanceName = bt_field_string_get_value(instanceNameField);
+  mInstanceName = bt_field_string_get_value(instanceNameField);
 }
 
 std::string AbstractPayload::getString() const {
-  return "{ typeName = \"" + mTypeName + "\", instanceName = \"" + instanceName + "\"" + specificPayloadString() + " }";
+  return "{ typeName = \"" + mTypeName + "\", instanceName = \"" + mInstanceName + "\"" + specificPayloadString() + " }";
 }
 
-
 bool AbstractPayload::operator==(const AbstractPayload& other) const {
-  return mTypeName == other.mTypeName && instanceName == other.instanceName && specificPayloadEqual(other);
+  return mTypeName == other.mTypeName && mInstanceName == other.mInstanceName && specificPayloadEqual(other);
 }
 
 // ************** //
 // FBEventPayload //
 // ************** //
+FBEventPayload::FBEventPayload(const std::string& paTypeName, const std::string& paInstanceName, const uint64_t paEventId) : 
+  AbstractPayload(paTypeName, paInstanceName),
+  mEventId(paEventId) {
+}
 
 FBEventPayload::FBEventPayload(const bt_field* field) : AbstractPayload(field){
-  eventId = bt_field_integer_unsigned_get_value(
+  mEventId = bt_field_integer_unsigned_get_value(
       bt_field_structure_borrow_member_field_by_name_const(field, "eventId"));
 }
 
 std::string FBEventPayload::specificPayloadString() const {
-  return ", eventId = " + std::to_string(eventId);
+  return ", eventId = " + std::to_string(mEventId);
 }
 
 bool FBEventPayload::specificPayloadEqual(const AbstractPayload& other) const {
   const auto& childInstance = dynamic_cast<const FBEventPayload&>(other); 
-  return eventId == childInstance.eventId;
+  return mEventId == childInstance.mEventId;
 }
 
 // ************* //
 // FBDataPayload //
 // ************* //
+FBDataPayload::FBDataPayload(const std::string& paTypeName, const std::string& paInstanceName, uint64_t paDataId, const std::string& paValue) :
+  AbstractPayload(paTypeName, paInstanceName),
+  mDataId(paDataId), mValue(paValue){
+}
+
 
 FBDataPayload::FBDataPayload(const bt_field* field) : AbstractPayload(field){
-  dataId = bt_field_integer_unsigned_get_value(
+  mDataId = bt_field_integer_unsigned_get_value(
       bt_field_structure_borrow_member_field_by_name_const(field, "dataId"));
 
-  value = bt_field_string_get_value(
+  mValue = bt_field_string_get_value(
       bt_field_structure_borrow_member_field_by_name_const(field, "value"));
 }
 
 std::string FBDataPayload::specificPayloadString() const {
-  auto convertedString = value;
+  auto convertedString = mValue;
 
   // add backslash to quoutes and double quotes
   for(auto toReplace : {"\"", "\'"}){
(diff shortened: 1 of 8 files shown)


tests/core/trace/EventMessage.h
```diff
@@ -0,0 +1,106 @@
...
+#include <string>
+#include <vector>
+
+#include <babeltrace2/babeltrace.h>
+
+class AbstractPayload;
+
+class EventMessage {
+public:
+  EventMessage(const std::string& paEventType, std::unique_ptr<AbstractPayload> paPayload, int64_t paTimestamp);
```
mx990: I would suggest passing the string parameters by value instead of by reference. For an explanation why this is generally preferred, please refer to https://xania.org/202101/cpp-by-value-args.

tests/core/trace/EventMessage.h
```diff
@@ -0,0 +1,106 @@
...
+class FBEventPayload : public AbstractPayload {
+public:
+  FBEventPayload(const std::string& paTypeName, const std::string& paInstanceName, const uint64_t eventId);
+
+  FBEventPayload(const bt_field* field);
+
+private:
+  std::string specificPayloadString() const;
+
+  bool specificPayloadEqual(const AbstractPayload& other) const;
```
mx990: I believe those methods should be marked `override` (same in the other concrete payload classes)?

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+#include <array>
+#include <iomanip>
+
+// ******************** //
+// EventMessage methods //
+// ******************** //
+
+EventMessage::EventMessage(const std::string& paEventType, std::unique_ptr<AbstractPayload> paPayload, int64_t paTimestamp) : 
+  mEventType(paEventType), mPayload(std::move(paPayload)), mTimestamp(paTimestamp){
+}
```
mx990, line 13: I would suggest moving this rather trivial implementation into the header file (same for the similar constructors in the other classes and `operator==`s).

tests/core/trace/EventMessage.h
```diff
@@ -0,0 +1,106 @@
...
+
+class AbstractPayload {
+public:
+  AbstractPayload(const std::string& paTypeName, const std::string& paInstanceName);
+
+  AbstractPayload(const bt_field* field);
+  
+  virtual std::string getString() const;
+
+  virtual bool operator==(const AbstractPayload& other) const;
```
mx990: Why is this `operator==` marked as virtual?

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+  const bt_event_class *event_class = bt_event_borrow_class_const(event);
+
+  const bt_field *payload_field =
+      bt_event_borrow_payload_field_const(event);
+
+  mEventType = bt_event_class_get_name(event_class);
+  mPayload = PayloadFactory::createPayload(mEventType, payload_field);
+
+  if(mPayload == nullptr){
+    std::abort();
```
mx990, line 30: Please log an error message instead of just calling `abort()` to help diagnose test failures.

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+// *********************** //
+// AbstractPayload Methods //
+// *********************** //
+AbstractPayload::AbstractPayload(const std::string& paTypeName, const std::string& paInstanceName) 
+  : mTypeName(paTypeName), mInstanceName(paInstanceName)  {
+}
+
+AbstractPayload::AbstractPayload(const bt_field* field) {
+  auto const typeNameField = bt_field_structure_borrow_member_field_by_name_const(field, "typeName");
+  auto const instanceNameField = bt_field_structure_borrow_member_field_by_name_const(field, "instanceName");
```
mx990: Please check the return values to make sure the fields exist and have the correct class.

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+// FBEventPayload //
+// ************** //
+FBEventPayload::FBEventPayload(const std::string& paTypeName, const std::string& paInstanceName, const uint64_t paEventId) : 
+  AbstractPayload(paTypeName, paInstanceName),
+  mEventId(paEventId) {
+}
+
+FBEventPayload::FBEventPayload(const bt_field* field) : AbstractPayload(field){
+  mEventId = bt_field_integer_unsigned_get_value(
+      bt_field_structure_borrow_member_field_by_name_const(field, "eventId"));
```
mx990: Please make sure the field exists and has the correct class.

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+  mDataId(paDataId), mValue(paValue){
+}
+
+
+FBDataPayload::FBDataPayload(const bt_field* field) : AbstractPayload(field){
+  mDataId = bt_field_integer_unsigned_get_value(
+      bt_field_structure_borrow_member_field_by_name_const(field, "dataId"));
+
+  mValue = bt_field_string_get_value(
+      bt_field_structure_borrow_member_field_by_name_const(field, "value"));
```
mx990: Please make sure the fields exist and have the correct class.

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+std::string FBDataPayload::specificPayloadString() const {
+  auto convertedString = mValue;
+
+  // add backslash to quoutes and double quotes
+  for(auto toReplace : {"\"", "\'"}){
+    for(auto pos = convertedString.find(toReplace); pos != std::string::npos; pos = convertedString.find(toReplace, pos + 2)){
+      convertedString.insert(pos, "\\");
+    }
+  }
```
mx990, line 180: Why is adding extra backslashes necessary here? What if there is already a `STRING` with an IEC 61131-3 escape `$'`?

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+}
+
+FBInstanceDataPayload::FBInstanceDataPayload(const bt_field* field) : AbstractPayload(field){
+
+  std::array<const char*, 4> fieldNames{"inputs", "outputs", "internal", "internalFB"};
+  std::array<std::vector<std::string>*, 4> storage = {&mInputs, &mOutputs, &mInternal, &mInternalFB};
+
+  for(size_t fieldRunner = 0; fieldRunner < fieldNames.size(); fieldRunner++){
+    auto arrayField = bt_field_structure_borrow_member_field_by_name_const(field, fieldNames[fieldRunner]);
```
mx990: Please make sure the field exists and has the correct class.

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+    auto arrayLength = bt_field_array_get_length(arrayField);
+    for(decltype(arrayLength) dynamicArrayRunner = 0; dynamicArrayRunner < arrayLength; dynamicArrayRunner++) {
+      storage[fieldRunner]->emplace_back(bt_field_string_get_value(
+          bt_field_array_borrow_element_field_by_index_const(arrayField, dynamicArrayRunner)
+      ));
+    } 
+  }
```
mx990, line 214: Instead of using two arrays and a for loop for reading the fields, I would suggest to extract this into a separate method, which is passed the field, name, and destination pointer (e.g., `readField(field, "inputs", &mInputs)`).

tests/core/trace/EventMessage.h
```diff
@@ -0,0 +1,106 @@
...
+
+#include <babeltrace2/babeltrace.h>
+
+class AbstractPayload;
+
+class EventMessage {
+public:
+  EventMessage(const std::string& paEventType, std::unique_ptr<AbstractPayload> paPayload, int64_t paTimestamp);
+
+  EventMessage(const bt_message* message);
```
mx990: Please stick to the coding guidelines for parameter names (e.g., `paMessage`).

tests/core/trace/TraceReaderSink.cpp
```diff
@@ -0,0 +1,157 @@
...
+  return BT_COMPONENT_CLASS_SINK_CONSUME_METHOD_STATUS_OK;
+}
+
+}
+
+// ************************** //
+// C-Style functions needed   // 
+// by the babeltrace2 library //
+// to create the components   //
+// ************************** //
```
mx990, line 93: I believe those functions should each be declared as `static` and put inside an `extern "C" { ... }` block.

tests/core/trace/EventMessage.h
```diff
@@ -0,0 +1,106 @@
...
+  std::string mEventType;
+  std::unique_ptr<AbstractPayload> mPayload;
+  int64_t mTimestamp;
+};
+
+class AbstractPayload {
+public:
+  AbstractPayload(const std::string& paTypeName, const std::string& paInstanceName);
+
+  AbstractPayload(const bt_field* field);
```
mx990: Please add a default virtual destructor to this class.

tests/core/trace/ctfTracerTest.cpp
```diff
@@ -0,0 +1,272 @@
...
+    BOOST_TEST_INFO("Expexted event number " + std::to_string(i));
+    BOOST_CHECK(ctfMessages[i] == expectedMessages[i]);
+  }
+
+  // add extra event to check that the comparison fails
+  expectedMessages.emplace_back("sendOutputEvent", std::make_unique<FBEventPayload>("E_RESTART", "START", 2),0);
+  BOOST_CHECK(ctfMessages != expectedMessages);
+}
+
+std::vector<EventMessage> getEventMessages(const std::string& path){
```
mx990: Please consider separating this generic support functionality from the concrete test above.

tests/core/trace/EventMessage.cpp
```diff
@@ -0,0 +1,214 @@
...
+      std::setw(2) << std::setfill('0') << hours + 1 << ":" <<
+      std::setw(2) << std::setfill('0') << minutes << ":" <<
+      std::setw(2) << std::setfill('0') << seconds << "." <<
+      std::setw(6) << std::setfill('0') << nanoseconds <<  "]";
+
+  return ss.str();
+}
+
+bool EventMessage::operator==(const EventMessage& other) const {
+  return mEventType == other.mEventType  && *mPayload == *other.mPayload;
```
mx990: Please check `mPayload` for `nullptr` (since you also have a check in `getPayloadString` above).

mx990 (changes requested):

tests/core/trace/EventMessage.h
cochicde: nice one! done

tests/core/trace/EventMessage.h
cochicde: yes! done

tests/core/trace/EventMessage.cpp
cochicde: I'm particularly a fan of fully separating the definition from the declaration, mainly to have the header file easier to read in case I don't need information from the implementation. I'd keep as it is, at least for now

tests/core/trace/EventMessage.h
cochicde: wrong copy+paste from somewhere. fixed

tests/core/trace/EventMessage.cpp
cochicde: added info before every `abort`

tests/core/trace/EventMessage.cpp
cochicde: right! added everywhere (I think)

tests/core/trace/EventMessage.cpp
cochicde: I added the string representation actually mainly for debugging purposses, to make sure I was reading the the ctf messages from the files exactly as the `babeltrace2` command does and quickly check with `diff`. `babeltrace2` prints the strings and adds the escape character. That's the only reason to have it also here.

tests/core/trace/EventMessage.cpp
cochicde: good one. much nicer now

tests/core/trace/EventMessage.h
cochicde: done

tests/core/trace/TraceReaderSink.cpp
cochicde: I added the static but the macros down in the file just take the function pointers and do the magic to offer the plugin and components, and no `extern C` is needed, as I quickly saw, the function pointers are stored, but the function name itself is not used from the outside

tests/core/trace/EventMessage.h
cochicde: added

tests/core/trace/ctfTracerTest.cpp
cochicde: done

tests/core/trace/EventMessage.cpp
cochicde: right!

tests/core/trace/TraceReaderSink.cpp
mx990: Some compilers also have different *calling conventions* for C and C++ functions, besides the name mangling issue, see https://en.cppreference.com/w/cpp/language/language_linkage and also https://stackoverflow.com/questions/2594178/why-do-you-need-extern-c-for-c-callbacks-to-c-functions for the concrete case of passing a callback defined in C++ to a C function.

cochicde: interesting! I'll add _calling convetions_ in C callbacks to the things that will break my code when I least expect it. Updated!

mx990 (approved):

PR #117, opened 2024-05-26
Clean-up Device Creation and Execution procedure

azoitl:
In order to prepare the different architectures that the CDevice to be used on startup can be configured via CMake options in this commit some clean-ups and generalizations have been performed.

@dok-net, @franz-hoepfinger-4diac, @kumajaya  FYI

### DIFF
--- a/src/core/device.h
+++ b/src/core/device.h
@@ -64,6 +64,13 @@ class CDevice : public CResource {
       return 1;
     }
 
+
+    /*!\brief wait till the execution of this device ends.
+     *
+     * The execution of devices can be stopped by sending it the kill command.
+     */
+    virtual void awaitShutdown() = 0;
+
     /*!\brief Execute the given management command
      *
      * Evaluates the mDestination parameter of the command if empty this class tries to execute the management command if not
--- a/src/arch/ecos/forte_instance.cpp
+++ b/src/arch/ecos/forte_instance.cpp
@@ -98,7 +98,7 @@ void forteStopInstance(int paSig, TForteInstance paResultDevice){
   RMT_DEV *poDev = static_cast<RMT_DEV*>(paResultDevice);
   if(0 != poDev){
     poDev->changeFBExecutionState(EMGMCommandType::Kill);
-    poDev->MGR.joinResourceThread();
+    poDev->awaitShutdown();
     DEVLOG_INFO("FORTE finished\n");
     delete poDev;
   }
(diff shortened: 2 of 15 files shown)


src/core/device.h
```diff
@@ -64,6 +64,13 @@ class CDevice : public CResource {
       return 1;
     }
 
+
+    /*!\brief wait till the execution of this device ends.
+     *
+     * The execution of devices can be stopped by sending it the kill command.
+     */
+    virtual void joinDevice() = 0;
```
cochicde: Two small things regarding the name of the function: 
1. I'd say `Device` is not needed in the name, since the method belongs already to the `CDevice` class, it seems redundant to have again the word in it. I see that also `startDevice` have the same problem, but I think it also has a reduncancy (not to be handled in this PR)
2. The word `join` is too specific for thread/process behaviour, which you're actually abstracting with this PR. I thought `wait` could be an option, since this is what the method does according to the documentation.

So, I'd say `wait` could be used instead of `joinDevice`.

MartinMelikMerkumians: While i principally agree, `wait` is used in the Java context to suspend a thread, so its somehow in the 4diac world also part of the `thread/process` concept.
In addition, calling `wait()` on a device, I would not understand from the name what we are waiting for.
The doxygen comment also seems not to be overly helpful.

At the end, we are exactly `join`ing a thread, so I would argue for using `join()`, but what keeps me off that suggesting a bit, is that `CDevice` isn't a thread, but only using a thread.

Would a lifecycle name of the thread, like `stop` be appropriate, and the doxygen documentation then accordingly updated?

MartinMelikMerkumians (changes requested):

src/core/device.h
azoitl: I agree that joinDevice is not good. I'm also not happy with wait.

This method is also not equivalent to end. 

The idea behind the method is that I want a blocking wait until the IEC 61499 device's execution is finished. 

The reason for that is that currently our device model is such that the architecture specific code is creating a device (this code will be reworked next), then starting it and let it run. Depending on the architecture we sometimes need a blocking wait until the device execution resumes and sometimes not. Therefore we can not encapsulate this in the device. But on nearly all architectures at one point we want to wait for the device to resume its execution to shutdown 4diac FORTE. That would be the task of this new method. I'm open to a good name suggestion for this.

mx990: How about naming the method `awaitShutdown()`?

azoitl: I like that a lot!

cochicde: me too!

azoitl: done

mx990 (approved):

MartinMelikMerkumians (approved):

PR #118, opened 2024-05-26
Added mingw64 toolchain file for quick tests under linux

azoitl:


buildsupport/toolchains/toolchain-mingw64-linux-cross.cmake
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Alois Zoitl - Initial contribution.
+###############################################################################
+
+SET(CMAKE_SYSTEM_NAME Windows)
+SET(TOOLCHAIN_PREFIX x86_64-w64-mingw32)
+
+SET(CMAKE_C_COMPILER ${TOOLCHAIN_BIN}/arm-none-eabi-gcc)
```
ernstblechaPT: I guess lines 17 and 18 are not needed as the variables are modified again in 20 and 21?

ernstblechaPT (changes requested):

buildsupport/toolchains/toolchain-mingw64-linux-cross.cmake
MartinMelikMerkumians: I assume that they are leftovers, as they also target arm systems, which I guess is not the intended target system.

As @ernstblechaPT please remove them, if we am correct

MartinMelikMerkumians (changes requested):

buildsupport/toolchains/toolchain-mingw64-linux-cross.cmake
azoitl: you both are correct. Forgot to remove them. Will do that now.

MartinMelikMerkumians (approved):

jwalt (approved):

PR #119, opened 2024-05-27
Moved device creation to CDevice

azoitl:
With this change CDevice provides a static method that will create an isntance of the device to be used by this 4daic FORTE instance.

This reduces code duplication and wrong device creation. Furthermore it is the basis that different device types can be configured via CMake.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -253,17 +253,26 @@ ENDFOREACH(FILE)
 FILE(WRITE ${CMAKE_BINARY_DIR}/file_list.txt "${WRITE_FILE}")
 
 ######################################################################################
-# Architecutre to build forte on
+# Architecutre to build 4diac FORTE on
 ######################################################################################
 set(FORTE_ARCHITECTURE "None" CACHE STRING "Architecture to build FORTE on")
 GET_PROPERTY(architectures GLOBAL PROPERTY FORTE_ARCHITECTURES)
 list(SORT architectures)
 set_property(CACHE FORTE_ARCHITECTURE PROPERTY STRINGS None ${architectures})
 
+######################################################################################
+# Device to be used in this 4diac FORTE instance
+######################################################################################
+set(FORTE_DEVICE "RMT_DEV" CACHE STRING "Device to be used in this 4diac FORTE instance")
+GET_PROPERTY(devices GLOBAL PROPERTY FORTE_DEVICES)
+list(SORT devices)
+set_property(CACHE FORTE_DEVICE PROPERTY STRINGS None ${devices})
+
+
 #######################################################################################
 # Link Libraries to the Executable
 #######################################################################################
-# get Linker-Direcotries from Modules
+# get Linker-Directories from Modules
 get_property(LINK_LIBRARY GLOBAL PROPERTY FORTE_LINK_LIBRARY)
 
 #######################################################################################
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -268,6 +268,7 @@ GET_PROPERTY(devices GLOBAL PROPERTY FORTE_DEVICES)
 list(SORT devices)
 set_property(CACHE FORTE_DEVICE PROPERTY STRINGS None ${devices})
 
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/forte_device_config.h.in ${CMAKE_BINARY_DIR}/forte_device_config.h)
 
 #######################################################################################
 # Link Libraries to the Executable
--- a/src/arch/ecos/phycoreat91/phycoreAT91main.cpp
+++ b/src/arch/ecos/phycoreat91/phycoreAT91main.cpp
@@ -10,7 +10,7 @@
  *   Alois Zoitl - initial API and implementation and/or initial documentation
  *******************************************************************************/
 #include <fortenew.h>
-#include <device.h>
+#include "device.h"
 #include <cyg/kernel/kapi.h>
 #include <network.h>
 #include <cyg/hal/hal_io.h>             // IO macros
@@ -19,8 +19,7 @@ externC void
 cyg_user_start( void );
 
 void cyg_user_start(){
-  CDevice *dev = CDevice::createDev("");
-  dev->startDevice();
+  CDevice::startupNewDevice("");
 }
 
 //get the stuff that fixes some ecos cpp problems which have to be near the main
--- a/src/arch/ecos/phycoreat91/phycoreAT91main.cpp
+++ b/src/arch/ecos/phycoreat91/phycoreAT91main.cpp
@@ -10,7 +10,7 @@
  *   Alois Zoitl - initial API and implementation and/or initial documentation
  *******************************************************************************/
 #include <fortenew.h>
-#include "../../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
 #include <cyg/kernel/kapi.h>
 #include <network.h>
 #include <cyg/hal/hal_io.h>             // IO macros
@@ -19,7 +19,7 @@ externC void
 cyg_user_start( void );
 
 void cyg_user_start(){
-  RMT_DEV *dev = new RMT_DEV;  //otherwise we would loose the var if it is only on the stack
+  CDevice *dev = CDevice::createDev("");
   dev->startDevice();
 }
 
--- a/src/arch/ecos/phycoreat91/phycoreAT91main.cpp
+++ b/src/arch/ecos/phycoreat91/phycoreAT91main.cpp
@@ -10,7 +10,7 @@
  *   Alois Zoitl - initial API and implementation and/or initial documentation
  *******************************************************************************/
 #include <fortenew.h>
-#include "device.h"
+#include "forteinstance.h"
 #include <cyg/kernel/kapi.h>
 #include <network.h>
 #include <cyg/hal/hal_io.h>             // IO macros
@@ -18,8 +18,10 @@
 externC void
 cyg_user_start( void );
 
+C4diacFORTEInstance g4diacForteInstance;
+
 void cyg_user_start(){
-  CDevice::startupNewDevice("");
+  g4diacForteInstance.startupNewDevice("");
 }
 
 //get the stuff that fixes some ecos cpp problems which have to be near the main
--- a/src/arch/freeRTOS/forte_Init.cpp
+++ b/src/arch/freeRTOS/forte_Init.cpp
@@ -18,7 +18,7 @@
 #include "forte_printer.h"
 #include <stdio.h>
 #include <string>
-#include <device.h>
+#include "device.h"
 
 #include "../utils/mainparam_utils.h"
 
@@ -30,7 +30,6 @@ unsigned int forte_default_port = 61499;
  */
 
 bool checkEndianess();
-void createDev(const char *paMGRID, TForteInstance* paResultInstance);
 
 void forteGlobalInitialize() {
   CForteArchitecture::initialize();
@@ -40,7 +39,7 @@ void forteGlobalDeinitialize() {
   CForteArchitecture::deinitialize();
 }
 
-int forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance) {
+int forteStartInstance(unsigned int paPort) {
 
   if(65535 < paPort) {
     return FORTE_WRONG_PARAMETERS;
@@ -58,10 +57,10 @@ int forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance) {
   strcat(address, port);
 
   char* arguments[] = { progName, flag, address };
-  return forteStartInstanceGeneric(3, arguments, paResultInstance);
+  return forteStartInstanceGeneric(3, arguments);
 }
 
-int forteStartInstanceGeneric(int paArgc, char *paArgv[], TForteInstance* paResultInstance) {
+int forteStartInstanceGeneric(int paArgc, char *paArgv[]) {
 
   if(!CForteArchitecture::isInitialized()) {
     return FORTE_ARCHITECTURE_NOT_READY;
@@ -81,7 +80,10 @@ int forteStartInstanceGeneric(int paArgc, char *paArgv[], TForteInstance* paResu
 
   const char *pIpPort = parseCommandLineArguments(paArgc, paArgv);
   if((0 != strlen(pIpPort)) && (nullptr != strchr(pIpPort, ':'))) {
-    createDev(pIpPort, paResultInstance);
+    if(!CDevice::startupNewDevice(ipPort)) {
+      return FORTE_COULD_NOT_CREATE_DEVICE;
+    }
+    DEVLOG_INFO("FORTE is up and running\n");
   } else { //! If needed call listHelp() to list the help for FORTE
     return FORTE_WRONG_PARAMETERS;
   }
@@ -89,36 +91,17 @@ int forteStartInstanceGeneric(int paArgc, char *paArgv[], TForteInstance* paResu
   return FORTE_OK;
 }
 
-void forteJoinInstance(TForteInstance paInstance) {
-  CDevice *poDev = static_cast<CDevice*>(paInstance);
-  if(0 != poDev) {
-    poDev->awaitShutdown();
-  }
+void forteJoinInstance() {
+  CDevice::awaitDeviceShutdown();
 }
 
-void forteStopInstance(int paSig, TForteInstance paInstance) {
+void forteStopInstance(int ) {
   if(!CForteArchitecture::isInitialized()) {
     return;
   }
-  (void) paSig;
-  CDevice *poDev = static_cast<CDevice*>(paInstance);
-  if(0 != poDev) {
-    poDev->changeFBExecutionState(EMGMCommandType::Kill);
-    poDev->awaitShutdown();
-    DEVLOG_INFO("FORTE finished\n");
-    delete poDev;
-  }
-}
-
-/*!\brief Creates the Device-Object
- * \param paMGRID A string containing IP and Port like [IP]:[Port]
- * \param The result
- */
-void createDev(const char *paMGRID, TForteInstance* paInstance) {
-  CDevice *device = CDevice::createDev(paMGRID);
-  device->startDevice();
-  *paInstance = device;
-  DEVLOG_INFO("FORTE is up and running\n");
+  CDevice::triggerDeviceShutdown();
+  CDevice::awaitDeviceShutdown();
+  DEVLOG_INFO("FORTE finished\n");
 }
 
 bool checkEndianess() {
--- a/src/arch/freeRTOS/forte_Init.cpp
+++ b/src/arch/freeRTOS/forte_Init.cpp
@@ -18,7 +18,7 @@
 #include "forte_printer.h"
 #include <stdio.h>
 #include <string>
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
 
 #include "../utils/mainparam_utils.h"
 
@@ -90,7 +90,7 @@ int forteStartInstanceGeneric(int paArgc, char *paArgv[], TForteInstance* paResu
 }
 
 void forteJoinInstance(TForteInstance paInstance) {
-  RMT_DEV *poDev = static_cast<RMT_DEV*>(paInstance);
+  CDevice *poDev = static_cast<CDevice*>(paInstance);
   if(0 != poDev) {
     poDev->awaitShutdown();
   }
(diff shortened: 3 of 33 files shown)


src/stdfblib/ita/RMT_DEV.cpp
```diff
@@ -16,6 +17,15 @@
 #endif
 #include <stringdict.h>
 
+CDevice* CDevice::createDev(const std::string &paMGRID) {
```
mx990: This seems like an inversion of responsibility when the base class directly creates an instance of a subclass. It also seems confusing to put a method belonging to `CDevice` in `RMT_DEV.cpp`.

How about having a separate `createDevice` in `CDevice` that chooses the correct implementation and delegates to a `createDevice` in the concrete class (e.g., `RMT_DEV`)? Maybe we could even use the typelib for this, since `CDevice` inherits from `CFunctionBlock`?

mx990 (changes requested):

src/stdfblib/ita/RMT_DEV.cpp
azoitl: The idea is to have a method that will create the respective device to be used for a specific 4diac FORTE instance. There will only be one sub-class of CDevice in any 4diac FORTE instance. Therefore I typelibrary will be an overkill and the wrong thing. As you see in may next commit this method will not be used by anyone else then CDevice and with that it is a kind of factory method that I can add to the code with compile time flags.

src/arch/ecos/phycoreat91/phycoreAT91main.cpp
```diff
@@ -10,7 +10,7 @@
  *   Alois Zoitl - initial API and implementation and/or initial documentation
  *******************************************************************************/
 #include <fortenew.h>
-#include "../../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: please use "" for the include, as <> is for library includes

src/arch/freeRTOS/forte_Init.cpp
```diff
@@ -18,7 +18,7 @@
 #include "forte_printer.h"
 #include <stdio.h>
 #include <string>
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: please use "" instead of <>

src/arch/freeRTOS/main.cpp
```diff
@@ -15,22 +15,18 @@
 
 #include "../forte_architecture.h"
 #include "../devlog.h"
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/pikeos_posix/main.cpp
```diff
@@ -14,15 +14,11 @@
 #include <fortenew.h>
 #include <stdio.h>
 #include <signal.h>
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/posix/main.cpp
```diff
@@ -16,7 +16,7 @@
 #include <stdio.h>
 #include <signal.h>
 #include "../startuphook.h"
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/rcX/forte_instance.cpp
```diff
@@ -17,7 +17,7 @@
 #include "forte_printer.h"
 #include <stdio.h>
 #include <string>
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/vxworks/main.cpp
```diff
@@ -12,11 +12,7 @@
 #include <fortenew.h>
 #include <stdio.h>
 #include <signal.h>
-#include "../../stdfblib/ita/RMT_DEV.h"
-
-#ifdef CONFIG_POWERLINK_USERSTACK
-#include <EplWrapper.h>
-#endif
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/win32/forte_instance.cpp
```diff
@@ -12,27 +12,21 @@
 #include "forte_instance.h"
 #include "fortenew.h"
 #include "forte_architecture.h"
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/win32/main.cpp
```diff
@@ -13,7 +13,7 @@
 #include "../forte_architecture.h"
 #include "../devlog.h"
 #include "../startuphook.h"
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

src/arch/zephyr/forte_Init.cpp
```diff
@@ -17,7 +17,7 @@
 #include "forte_printer.h"
 #include <stdio.h>
 #include <string>
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include <device.h>
```
MartinMelikMerkumians: same here

MartinMelikMerkumians (changes requested):

src/arch/ecos/phycoreat91/phycoreAT91main.cpp
azoitl: Done

src/arch/freeRTOS/forte_Init.cpp
azoitl: Done

src/arch/freeRTOS/main.cpp
azoitl: Done

src/arch/pikeos_posix/main.cpp
azoitl: Done

src/arch/posix/main.cpp
azoitl: Done

src/arch/rcX/forte_instance.cpp
azoitl: Done

src/arch/vxworks/main.cpp
azoitl: Done

src/arch/win32/forte_instance.cpp
azoitl: Done

src/arch/win32/main.cpp
azoitl: Done

src/arch/zephyr/forte_Init.cpp
azoitl: Done

src/stdfblib/ita/RMT_DEV.cpp
mx990: I have created a PR based on this one with a slightly different solution that chooses the implementation based on a macro, see https://github.com/azoitl/4diac-forte/pull/1.

It avoids having to put `CDevice::createDev` into each of the concrete implementations and therefore allows to include all devices in compilation, meaning `OPCUA_DEV` can still inherit from `RMT_DEV`.

What do you think?

azoitl: I like the factory methods in the devices. It allows to make less device specific information public. Furthermore, it would open up in the future to select the device through a command line param. Although I don't know if we ever want to do that. 

What I don't find so nice is the macro based approach. I think a CMake code generation with `.in` files is better. What do you think @MartinMelikMerkumians?

`OPCUA_DEV` inheriting from RM_DEV was only an intermediate test for simplifying its development. One of the next steps would be to get rid of that (FYI @m-meingast).

mx990: > Furthermore, it would open up in the future to select the device through a command line param. Although I don't know if we ever want to do that.

I believe that could make sense in situations when using a pre-compiled forte, such as when loading types with Lua or other means.

> What I don't find so nice is the macro based approach. I think a CMake code generation with .in files is better.

I changed the PR to use a configure file and type alias instead of the macro.

azoitl: Interesting I would have moved the whole `CDevice::createDev` to the `.in` file. Like we do it for the communication layer factory

mx990: I usually try to keep any configure file purely declarative if possible. I believe this makes them easier to read and maintain compared to having code mixed with replacements.

cochicde: I also think the whole `CDevice::createDev` should not be part of `CDevice`. 
I see at least 3 things trying to be solved which should not be part of `CDevice`, since it adds too much responsabiliy to it:
1. Creation of a device with `createDev`: maybe  you could have a `DeviceFactory` class/file handling the creation, and in it, you would have everything needed for handling (`.in`, dynamic or static polymorphism, ....). An idea could be to have it simple for now as:
  ```cpp
    #include "src/forte_device_config.h.in"

   std::unique_ptr<CDevice> createDevice()
   {
      return new FORTE_DEVICE(); 
   }

  ```
2. Handling the device life cycle: all these functions `startupNewDevice`,  `triggerDeviceShutdown` and `awaitDeviceShutdown` should be part of a `DeviceHandler` (or actually a ForteInstance class? ) class with a private `std::unique_ptr<CDevice>` member which uses the factory from point 1. If at some point you want to instantiate based on command line paramaeters, the constructor of this handler should take this parameter, and the factory should be a dynamic factory based on this parameter.

image

3. Repeated code in `forte_instance`/forte_init: Many if not all of these files, should be in one class. I'd recommend to have ForteInstance as a class, throw all the code from the `forte_instance` files into it, and let the `forte_instance` files be a purely `C` interface which creates a `ForteInstance` object and calls its methods for handling the device lifecycle.

src/arch/freeRTOS/forte_Init.h
```diff
@@ -36,17 +36,16 @@ extern "C" {
...
-  typedef void* TForteInstance;
-
   /**
    * \brief Start forte instance
    * @param paPort The port on which to forte will listen. Use 0 for default (normally 61499)
    * @param paResultInstance Address of an instance of forte where the new instance is stored
    * @return FORTE_OK if no error occurred, other values otherwise
    */
-  FORTE_SHARED_PREFIX int FORTE_SHARED_CALL forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance);
+  FORTE_SHARED_PREFIX int FORTE_SHARED_CALL forteStartInstance(unsigned int paPort);
```
cochicde: From what I remember, this methods were intended to be used as a `C` interface to the dynamic/static forte library linked to `C` code. 

The idea of having the forte pointer managed by the caller was to avoid problems from a global pointer in a dynamic library which would be shared by many callers. I understand that this is maybe not the case now and it's a very strange case, but thinking this file as a purely `C` interface I would keep the pointer management to the caller. 

Also, I'm not sure why we have so many copies of this file, since the idea of the interface is to have a unique one for all cases.

src/CMakeLists.txt
```diff
@@ -253,17 +253,26 @@ ENDFOREACH(FILE)
...
 set_property(CACHE FORTE_ARCHITECTURE PROPERTY STRINGS None ${architectures})
 
+######################################################################################
+# Device to be used in this 4diac FORTE instance
+######################################################################################
+set(FORTE_DEVICE "RMT_DEV" CACHE STRING "Device to be used in this 4diac FORTE instance")
+GET_PROPERTY(devices GLOBAL PROPERTY FORTE_DEVICES)
+list(SORT devices)
+set_property(CACHE FORTE_DEVICE PROPERTY STRINGS None ${devices})
+
```
ernstblechaPT, line 270: I'm not sure if there should be some validation for the possible values of FORTE_DEVICE - the set_property call only limits the values that can be selected via CMakeGUI

```
if(NOT ${FORTE_DEVICE} IN_LIST devices)
  message(FATAL_ERROR "Invalid device selected: ${FORTE_DEVICE}")
endif()
```

```
if(NOT ${FORTE_DEVICE} IN_LIST devices)
  set(FORTE_DEVICE "RMT_DEV" CACHE STRING "Device to be used in this 4diac FORTE instance")
endif()
```

src/stdfblib/ita/RMT_DEV.cpp
azoitl: I like the idea of a `C4diacFORTEInstance` class. Will, based on @mx990's improvement, rework it accordingly. At that point I find a DeviceFactory a bit an overkill and would suggest to keep it for now with the `C4diacFORTEInstance`.

src/CMakeLists.txt
azoitl: Goo point. However sofar we also do not have it for architecture and there it also works fine. In the worst case you would get a compile error.

src/arch/freeRTOS/forte_Init.h
azoitl: I think the reason for the duplication is coming from the fact that you originally implemented it as part of the RCx architecture. As you originally did it. Could you after my changes harmonize that?

src/stdfblib/ita/RMT_DEV.cpp
azoitl: @cochicde I did a first prototype of your suggestion just for posix platform. If you think that is in the right direction I would implemented it across all architetures and clean-up the code.

src/core/forteinstance.cpp
```diff
@@ -0,0 +1,49 @@
...
+ *    Alois Zoitl, Martin Jobst - initial implementation and rework communication
+ *                                infrastructure
+ *******************************************************************************/
+
+#include "forteinstance.h"
+#include "forte_device_config.h"
+
+
+std::unique_ptr<CDevice> C4diacFORTEInstance::createDev(const std::string &paMGRID) {
+  return FORTE_DEVICE::createDev(paMGRID);
```
cochicde: I don't think this static method in each derived class from `CDevice` is a nice solution. It would be easier to have each derived classs get the `const std::string &paMGRID` as parameter in their constructor and have here
`return FORTE_DEVICE(paMGRID);`

if there's some plan for adding more parameters to the constructors of devices, you could have this method templated as:

```cpp
template<typename... Ts>
std::unique_ptr<CDevice> C4diacFORTEInstance::createDev(Ts&&... params)
{
  return FORTE_DEVICE(std::forward<Ts>(params)...); 
}
```

src/core/forteinstance.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+bool C4diacFORTEInstance::startupNewDevice(const std::string &paMGRID){
```
cochicde, line 25: What about this:
1. Move the implementation of this method in the constructor, so the `if(mActiveDevice)` can be removed
2. If failed, throw an exception. If exceptions are a problem to have, then maybe add a `public bool isInitialized() {mActiveDevice != nullptr;}` method to `C4diacFORTEInstance`
3. Move the implementation of `triggerDeviceShutdown` and `awaitDeviceShutdown` in the destructor.
4. in the main, you could have a `std::unique_ptr<C4diacFORTEInstance> g4diacForteInstance` instead of an object, and call `new` on it when intializing, and `reset()` on `endForte`

src/arch/freeRTOS/forte_Init.h
cochicde: > I think the reason for the duplication is coming from the fact that you originally implemented it as part of  the RCx architecture. As you originally did it. Could you after my changes harmonize that?

oh yes. please let me clean this stuff afterwards. We should probably remove the rcX arch, because I think it was never used. At that time, I'm not even sure if I got it working.

cochicde (commented):
I like it! I left some small comments

src/core/forteinstance.cpp
azoitl: I tend to disagree. Implementing a correct RMT_RES just with a constructor is not as easy as it looks. Having a factory method like proposed by  @mx990  is much easier and gives more flexibility.

src/core/forteinstance.cpp
azoitl: I have to sleep that over.

src/core/forteinstance.cpp
cochicde: but you wouldn't touch the RMT_RES. `RMT_DEV` would take `paMGRID` and call `setMGR_ID(paMGRID);` directly in the constructor, storing it locally in its `DI`. `RMT_RES` is constructed as before.
Calling `RMT_DEV::initialize()` could be done probably in the constructor, but since it could fail and to avoid `execptions` , you could have it in `C4diacFORTEInstance::startupNewDevice` before starting the device.

mx990: > `RMT_DEV` would take `paMGRID` and call `setMGR_ID(paMGRID);` directly in the constructor, storing it locally in its `DI`.

The input variable `MGR_ID` of `RMT_RES` has not yet been initialized in the constructor of `RMT_DEV`. This is done later in `RMT_DEV::initialize()`, which delegates to `RMT_RES::initialize()`. As a result, we cannot call `RMT_DEV::setMGR_ID(...)` in the constructor of `RMT_DEV`.

> Calling `RMT_DEV::initialize()` could be done probably in the constructor

Please refer to https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctors for the reason why calling `initialize()` (or any other virtual function, directly or indirectly) in the constructor is problematic.

azoitl: What could work is that I rework RMT_DEV as we have reworked FBs where the interface vars are now member vars. 

Then we would not need any creation or new and the device would be a normal member var of `C4diacFORTEInstance`. 

`initialize()` would then be called either in `C4diacFORTEInstance` constructor or maybe better in `C4diacFORTEInstance::startupNewDevice`.

azoitl: @cochicde I tried your proposed approach. In general to code is much nicer. However I can not manage to shut-down 4diac FORTE with it. While with the last commit everything works nice and fine.

src/core/forteinstance.cpp
cochicde: don't think much about this. I tried locally and only brought me problems.

src/core/forteinstance.cpp
cochicde: I tried it here and seems to be working fine https://github.com/azoitl/4diac-forte/pull/2 storing the string locally since the interface is not ready at that point as mentioned. 

I let the call to `initialize` still outside of it for now.

is there a reason the interface is not setup in the constructor?

azoitl: @cochicde your PR is not far enough. This is what I did. I think the problem is in shutting down the device in the destructor of `C4diacFORTEInstance`: 
[deviceParmas.diff.txt](https://github.com/eclipse-4diac/4diac-forte/files/15496560/deviceParmas.diff.txt)

cochicde: yes, I also tried the reworking `C4diacFORTEInstance` but could not go further. I think this class looks good as you have it and wouldn't complicate more. 

in https://github.com/azoitl/4diac-forte/pull/2 I left only the changes related to removing the `createDev` method in each derive class from `CDevice`

azoitl: If you don't mind I would take your PR as inspiration but when @mx990 PR #120 is merged it can be implemented a bit better. So I would wait for that and rework mine.

azoitl: @cochicde @mx990 I think now it is a nice solution. What do you think?

src/stdfblib/ita/OPCUA_DEV.cpp
```diff
@@ -21,10 +21,19 @@ OPCUA_DEV::OPCUA_DEV() :
 OPCUA_DEV::~OPCUA_DEV() {
 }
 
+std::unique_ptr<CDevice> OPCUA_DEV::createDev(const std::string &paMGRID) {
```
cochicde: this should be probably be deleted

src/stdfblib/ita/OPCUA_DEV.h
```diff
@@ -24,5 +24,7 @@ class OPCUA_DEV : public RMT_DEV {
   OPCUA_DEV();
   virtual ~OPCUA_DEV();
 
+  static std::unique_ptr<CDevice> createDev(const std::string &paMGRID);
```
cochicde: this too should be removed

src/core/forteinstance.cpp
cochicde: looks good!

src/stdfblib/ita/OPCUA_DEV.cpp
azoitl: done

src/stdfblib/ita/OPCUA_DEV.h
azoitl: done.

Just for completness: OPCUA_DEV will in addition need a major rework. But that is something for another PR.

cochicde (commented):
LGTM!

mx990 (approved):
I like it :+1:

MartinMelikMerkumians (approved):

PR #120, opened 2024-05-30
Move dynamic interface setup to generic function block

mx990:
Move the dynamic interface setup and required variables to the generic function block and make accessor functions pure virtual.

This avoids duplicate memory allocations for function blocks following the new memory layout.

### DIFF
--- a/src/core/funcbloc.cpp
+++ b/src/core/funcbloc.cpp
@@ -375,27 +375,38 @@ CIEC_ANY *CFunctionBlock::getVar(CStringDictionary::TStringId *paNameList,
   return nullptr;
 }
 
-CAdapter *CFunctionBlock::getAdapter(CStringDictionary::TStringId paAdapterNameId) const{
+CAdapter *CFunctionBlock::getAdapter(CStringDictionary::TStringId paAdapterNameId) {
   TPortId adpPortId = getAdapterPortId(paAdapterNameId);
 
   if(cgInvalidPortId != adpPortId){
-    return mAdapters[adpPortId];
+    return getAdapterUnchecked(adpPortId);
   }
   return nullptr;
 }
 
-TPortId CFunctionBlock::getAdapterPortId(CStringDictionary::TStringId paAdapterNameId) const{
+const CAdapter *CFunctionBlock::getAdapter(CStringDictionary::TStringId paAdapterNameId) const {
+  TPortId adpPortId = getAdapterPortId(paAdapterNameId);
+
+  if(cgInvalidPortId != adpPortId){
+    return const_cast<CFunctionBlock*>(this)->getAdapterUnchecked(adpPortId);
+  }
+  return nullptr;
+}
+
+TPortId CFunctionBlock::getAdapterPortId(CStringDictionary::TStringId paAdapterNameId) const {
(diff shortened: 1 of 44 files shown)


src/core/funcbloc.cpp
```diff
@@ -375,27 +310,38 @@ CIEC_ANY *CFunctionBlock::getVar(CStringDictionary::TStringId *paNameList,
...
+  if(cgInvalidPortId != adpPortId){
+    return const_cast<CFunctionBlock*>(this)->getAdapterUnchecked(adpPortId);
+  }
+  return nullptr;
+}
+
+TPortId CFunctionBlock::getAdapterPortId(CStringDictionary::TStringId paAdapterNameId) const {
   for(TPortId i = 0; i < mInterfaceSpec->mNumAdapters; ++i){
-    if(mAdapters[i]->getInstanceNameId() == paAdapterNameId){
+    if(const_cast<CFunctionBlock*>(this)->getAdapterUnchecked(i)->getInstanceNameId() == paAdapterNameId){
```
azoitl: with the new adapter api it makes more sense to use here the adapter entry from the interface spec

src/stdfblib/ita/RMT_DEV.cpp
```diff
@@ -38,7 +38,7 @@ bool RMT_DEV::initialize() {
     return false;
   }
   MGR.initialize();
-  MGR_ID().fromString("localhost:61499");
+  var_MGR_ID.fromString("localhost:61499");
```
azoitl: This could now be in the constructor

src/core/funcbloc.cpp
```diff
@@ -554,6 +500,21 @@ void CFunctionBlock::nextDataPoint(const CStringDictionary::TStringId *&paDataTy
   }
 }
 
+CAdapter *CFunctionBlock::createAdapter(const SAdapterInstanceDef &paAdapterInstanceDefinition, TForteUInt8 paParentAdapterlistID) {
```
azoitl, line 503: Shouldn't this method be now at the CGenFB?

azoitl (changes requested):

azoitl:
@franz-hoepfinger-4diac this patch is important for you as it will greatly reduce memory requirements on your devices!

src/core/funcbloc.cpp
mx990: That is much better, thanks.

src/stdfblib/ita/RMT_DEV.cpp
mx990: Done

src/core/funcbloc.cpp
mx990: Unfortunately, that would create a circular dependency between the definition of `CAdapter` and `CGenFunctionBlock<T>::createAdapter`. Since `CGenFunctionBlock<T>` is a class template, I wouldn't be able to separate the definition from the class (i.e., I cannot put it in a `.cpp` file) like I can in `CFunctionBlock`.

azoitl: oh. Thx for the clarification.

azoitl (approved):

PR #136, opened 2024-06-04
Renamed SIGNALPROCESSING Filenames

PStmmr:
Changed filenames and the corresponding CMakeLists.


azoitl:
these files should not be renamed. They allready have the fct file ending.

diplfranzhoepfinger:
yes, true !

PR #138, opened 2024-06-04
Add a DDS communication layer; needs external code generation for ful…

jwalt:
This is a draft of a fully working, type-safe DDS communication layer. It needs external tools to compile, as the type definitions are generated by command line tools of the DDS library.

With this draft PR, we would like to request comments about the general approach.


cochicde:
wow! nicely done! however, for me is quite hard to separate the examples from and generated files from what's would actually end up in forte's code. 

I'd recommend leaving the example independent code under `src/com`. The `.idl` and `.msg` files are also not forte-related and the `.cxx` are generated so they shouldn't be important to understand the integration. 

The `example_message/StringDebug` seems small enough to see the integration. I'd recommend having this maybe under the `test` folder.

Also, there's some embrick/sysfs changes which seem not to belong here

jwalt:
sorry, rebased it to the wrong branch. sysfs stuff is gone :-)

gravemalte:
Right, the generated files are not forte related, they got published because otherwise the examples would not be able to run (and we want that you can try it out. Without the files, it is quite complicated to get everything fully working, a lot of code generation happens in the background). The plan is that you just write an `.msg` file put it somewhere and the whole generation process is integrated into the CMake build process. The whole thing needs some clean-up, but is a first proof-of-concept. If you have issues with running the examples, let me know.

diplfranzhoepfinger:
https://opendds.org/ 

https://github.com/eProsima/Fast-DDS-Gen


is it this DDS ?

gravemalte:
Hi.

DDS (Data Distribution Service) is a specification. In simple terms, DDS specifies a low-latency data exchanged for distributed systems. You can find some information for the DDS spec here: https://www.dds-foundation.org/what-is-dds-3/

Now to the second question.

Fast-DDS-Gen is a tool that generates a C++ code (only usable by the FastDDS library) from `.idl` files (a specific format for data exchange).

diplfranzhoepfinger:
cool. Which generator did you use ? 

DDS is interesting, because the AEF is discussing to base the next ISOBUS Standard either on DDS or SOME/IP.
if DDS is supported by 4diac, this might be a indication to vote for DDS instead of SOME/IP.

gravemalte:
This PR includes a prototype implementation using the FastDDS library and Fast-DDS-Gen for generating the necessary infrastructure. The plan is the following:

1. You define your custom data-type as a struct in 4diac
2. You export the data type in 4diac to the FBe build system
3.  A message file `.msg` will be generated
4. The `.msg` file will be converted to the idl format using rosidl (https://github.com/ros2/rosidl)
5. The Fast-DDS_Gen will read the `.idl` file and generate the necessary C++ code for the PubSub system
6. You're now able to receive data from a DDS layer in FORTE

PR #139, opened 2024-06-04
Update GPIO module to new Linux GPIO API

jwalt:


azoitl:
Do I see that correctly that you are creating a device controller per GPIO pin?

jwalt:
Yes.

azoitl:
Hmh. This is a bit against the design philosophy of 4diac FORTE and the also the modular IO interface. We normally try to group external event sources into single sources where they belong together.

jwalt:
Yeah, but this approach simplifies the code significantly, as I do not need to manage arrays of (possibly sparse) pin properties, file handles and such. At the API level, GPIOs aren't sensibly aggregatable. There is the concept of a "GPIO Chip" (probably better called an I/O Bank) with multiple "Lines", but this is internal to the SoC. The board may have accessible pins scattered across all internal "chips", with the physical arrangement on the PCB in no way resembling the internal numbering. Add to this that some popular boards (BeagleBone or RPi) can have a third way of numbering GPIOs as part of a canonical naming scheme. A fourth way of numbering is the package pin arrangement, which again can be arbitrary. See https://elinux.org/RPi_Low-level_peripherals for an example.

Long story short, I don't see a sensible way to determine "belonging together" other than the user specifying each pin on its own. Given the fact that GPIOs are usually quite limited in number, this seems to be the most efficient way of implementing it.

A different matter would be external IO extender chips, but for those we usually have dedicated I/O modules like embrick.

azoitl:
@jwalt you have a point here. Event tough I think we need to find once a solution for that problem I also see that this solution is much better then before. Thx for the work.

azoitl (approved):

azoitl:
@jwalt forgot to ask: could you also provide the IO Config fbs for the tool lib?

jwalt:
[gpiochip.fbt.zip](https://github.com/user-attachments/files/15969235/gpiochip.fbt.zip)

azoitl:
> [gpiochip.fbt.zip](https://github.com/user-attachments/files/15969235/gpiochip.fbt.zip)

could you please provide them as PR to the 4diac-ide repo?

PR #141, opened 2024-06-05
fix new device interface in ctfTracer test

cochicde:


tests/core/fbtests/fbtesterglobalfixture.h
```diff
@@ -14,7 +14,34 @@
...
+      return nullptr;
+    }
+
+    CDataConnection **getDIConUnchecked(TPortId) override {
+      return nullptr;
+    }
+
+};
+
+const static SFBInterfaceSpec gscTestDevSpec = {
```
azoitl: this could be a private const static field of CTesterDevice and directly be given to CDevice. This would reduce the interface of the CTesterDevice. Didn't notice this when I changed the code. But now that you make that publicly visible I think this should be repaired.

azoitl (changes requested):

tests/core/fbtests/fbtesterglobalfixture.h
cochicde: right! done!

cochicde:
Coming back to our discussion from last time, I'd always squash and merge PRs. So feel free to always do that to my PRs :)

azoitl (approved):

azoitl:
> Coming back to our discussion from last time, I'd always squash and merge PRs. So feel free to always do that to my PRs :)

I'm getting better when to squash and when not ;-)

PR #146, opened 2024-06-07
Reset event counters to 0 on initialization

ernstblechaPT:
In #131 the memset clearing the event counters was removed, this results in random values showing up on the event counters

### DIFF
--- a/src/core/funcbloc.cpp
+++ b/src/core/funcbloc.cpp
@@ -541,11 +541,11 @@ void CFunctionBlock::setupEventMonitoringData(){
 
   if(mInterfaceSpec) {
     if (0 != mInterfaceSpec->mNumEIs) {
-      mEIMonitorCount = new TForteUInt32[mInterfaceSpec->mNumEIs];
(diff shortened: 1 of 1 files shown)


src/core/funcbloc.cpp
```diff
@@ -542,10 +542,12 @@ void CFunctionBlock::setupEventMonitoringData(){
   if(mInterfaceSpec) {
     if (0 != mInterfaceSpec->mNumEIs) {
       mEIMonitorCount = new TForteUInt32[mInterfaceSpec->mNumEIs];
+      memset(mEIMonitorCount, 0, sizeof(TForteUInt32) * mInterfaceSpec->mNumEIs);
```
mx990: I would suggest adding curly-braces in the new expression to perform value initialization instead, see also https://en.cppreference.com/w/cpp/language/new#Initialization:
```cpp
new TForteUInt32[mInterfaceSpec->mNumEIs]{}
```

mx990 (changes requested):

mx990 (approved):

azoitl (approved):

PR #151, opened 2024-06-07
Clean main and init files from different architectures

cochicde:
Duplicated code was removed and a centralized entry point to forte is given (when possible).


src/arch/common/forte_specific_architecture.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef SRC_ARCH_COMMON_FORTE_SPECIFIC_ARCHITECTURE_H_
+#define SRC_ARCH_COMMON_FORTE_SPECIFIC_ARCHITECTURE_H_
+
+#include "forte_specific_architecture.h"
+
+class CForteSpecificArchitecture{
+  public:
+
+    static int initialize(int argc , char* argv[]);
```
cochicde: missing documentaiton

src/arch/forte.cpp
```diff
@@ -11,42 +11,42 @@
...
+#include "forte_printer.h"
+#include "forteinstance.h"
 
-unsigned const int cgForteDefaultPort = 61499;
+static const unsigned int scForteDefaultPort = 61499;
+static const unsigned int scMaxPortValue = 65535;
 
 /*!\brief Check if the correct endianess has been configured.
  *
  * If the right endianess is not set this function will end FORTE.
```
cochicde: remove this obsolete comment. Decision to end forte is done by the caller

src/arch/forte.cpp
```diff
@@ -11,42 +11,42 @@
...
  */
 bool checkEndianess();
 
-void forteGlobalInitialize(){
-  CForteArchitecture::initialize();
+int forteGlobalInitialize(int argc, char *argv[]){
+  if(auto result = CForteArchitecture::initialize(argc, argv); result != 0){
+    return result;
+  }
+  return 0;
```
cochicde:
```suggestion
  return CForteArchitecture::initialize(argc, argv);
```

src/arch/forte.cpp
```diff
@@ -11,42 +11,42 @@
...
+int forteGlobalDeinitialize(){
+  return CForteArchitecture::deinitialize();
 }
 
-int forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance){
+FORTE_STATUS forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance){
 
-  if(65535 < paPort){
+  if(scMaxPortValue < paPort){
+    DEVLOG_ERROR("Provided port %d not valid\n", paPort);
```
cochicde:
```suggestion
    DEVLOG_ERROR("Provided port %d is not valid\n", paPort);
```

src/arch/forte.cpp
```diff
@@ -60,50 +60,58 @@ int forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance){
...
-int forteStartInstanceGeneric(int argc, char *arg[], TForteInstance* paResultInstance){
-
-  if(!CForteArchitecture::isInitialized()){
-    return FORTE_ARCHITECTURE_NOT_READY;
-  }
+FORTE_STATUS forteStartInstanceGeneric(int argc, char *argv[], TForteInstance* paResultInstance){
 
-  if(0 == paResultDevice){
+  if(nullptr == paResultInstance){
+    DEVLOG_ERROR("Provided result instance parameter not valid\n");
```
cochicde:
```suggestion
    DEVLOG_ERROR("Provided result instance parameter is not valid\n");
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
 
   /**
    * \brief Start forte instance
-   * @param pa_port The port on which to forte will listen. Use 0 for default (normally 61499)
-   * @param pa_resultDevice Address of an instance of forte
+   * @param paPort The port on which to forte will listen. Use 0 for default (normally 61499)
```
cochicde:
```suggestion
   * @param paPort The port on which forte will listen. Use 0 for default 61499
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
...
+   * @param paResultInstance Address of an instance of forte
    * @return FORTE_OK if no error occurred, other values otherwise
    */
-  FORTE_SHARED_PREFIX int FORTE_SHARED_CALL forteStartInstance(unsigned int pa_port, TForteInstance* paResultInstance);
+  FORTE_SHARED_PREFIX FORTE_STATUS FORTE_SHARED_CALL forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance);
```
cochicde: FORTE_SHARED_PREFIX set this somewher (cmake or in a specific file).

remove FORTE_SHARED_CALL?

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
 
   /**
    * \brief Start forte instance
-   * @param pa_port The port on which to forte will listen. Use 0 for default (normally 61499)
-   * @param pa_resultDevice Address of an instance of forte
+   * @param paPort The port on which to forte will listen. Use 0 for default (normally 61499)
+   * @param paResultInstance Address of an instance of forte
    * @return FORTE_OK if no error occurred, other values otherwise
```
cochicde: specify possible error values

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
...
-  FORTE_SHARED_PREFIX int FORTE_SHARED_CALL forteStartInstance(unsigned int pa_port, TForteInstance* paResultInstance);
+  FORTE_SHARED_PREFIX FORTE_STATUS FORTE_SHARED_CALL forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance);
 
   /**
    * \brief Start forte instance with posibilities of more arguments
    * @param argc Number of arguments in arg
-   * @param arg Arguments
-   * @param pa_resultDevice Address of an instance of forte
+   * @param argv Arguments
+   * @param paResultInstance Address of an instance of forte
```
cochicde:
```suggestion
   * @param paResultInstance Address  to store the created forte instance
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
 
   /**
    * \brief Start forte instance
-   * @param pa_port The port on which to forte will listen. Use 0 for default (normally 61499)
-   * @param pa_resultDevice Address of an instance of forte
+   * @param paPort The port on which to forte will listen. Use 0 for default (normally 61499)
+   * @param paResultInstance Address of an instance of forte
```
cochicde:
```suggestion
   * @param paResultInstance Address  to store the created forte instance
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
...
    */
-  FORTE_SHARED_PREFIX int FORTE_SHARED_CALL forteStartInstanceGeneric(int argc, char *arg[], TForteInstance* paResultInstance);
+  FORTE_SHARED_PREFIX FORTE_STATUS FORTE_SHARED_CALL forteStartInstanceGeneric(int argc, char *argv[], TForteInstance* paResultInstance);
 
   /**
-   * \brief Terminates a Forte instance
-   * @param signal  Signal value to terminate instance
-   * @param pa_resultDevice Instance to terminate
+   * \brief Request termination of a Forte instance
+   * @param paInstance Instance to terminate
```
cochicde:
```suggestion
   * @param paInstance Instance to request for termination
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
...
    */
-  FORTE_SHARED_PREFIX void FORTE_SHARED_CALL forteStopInstance(int signal, TForteInstance paInstance);
+  FORTE_SHARED_PREFIX void FORTE_SHARED_CALL forteRequestStopInstance(TForteInstance paInstance);
+
+  /**
+   * \brief Waits indefinitely for the intance to stop
```
cochicde:
```suggestion
   * \brief Waits indefinitely for the intance to stop and deletes it
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
...
+   * @param paInstance Instance to terminate
+  */
+  FORTE_SHARED_PREFIX void FORTE_SHARED_CALL forteWaitForInstanceToStop(TForteInstance paInstance);
 
   /**
    * \brief Initializes the architecture. Prepare all resources needed by the Forte's instances. Must be called once before the first Forte instance is started
+   * @param argc Number of arguments in arg
+   * @param argv Arguments
+   * @return 0 if no error occurred, other values otherwise
```
cochicde:
```suggestion
   * @return 0 if no error occurred, other values otherwise. Any error value is provided by the architecture and not by forte
```

src/arch/forte.h
```diff
@@ -44,41 +44,50 @@ extern "C" {
...
    */
-  FORTE_SHARED_PREFIX void FORTE_SHARED_CALL forteGlobalInitialize();
+  FORTE_SHARED_PREFIX int FORTE_SHARED_CALL forteGlobalInitialize(int argc, char *argv[]);
 
   /**
    * \brief Deinitializes the architecture. Frees all resources used by Forte's instances. Must be called after the last instance is ended
+   * @return 0 if no error occurred, other values otherwise
```
cochicde:
```suggestion
   * @return 0 if no error occurred, other values otherwise.  Any error value is provided by the architecture and not by forte
```

src/arch/forte_architecture.h
```diff
@@ -15,20 +15,60 @@
...
+
+#include "forteinit.h"
+#include "startuphook.h"
+
+template<typename SpecificArchitectureT>
+class CForteGeneralArchitecture{
 
   public:
-    static bool initialize();
+    static int initialize(int argc, char *argv[]);
```
cochicde, line 30: add missing documentation

src/arch/freeRTOS/main.cpp
```diff
@@ -22,7 +22,7 @@ const static unsigned mainFORTE_TASK_PRIORITY = tskIDLE_PRIORITY + 1;
 void vForteTask(void* ) {
   C4diacFORTEInstance instance;
```
cochicde: use common `forte.h` interface. Do we need to initialize forte globally here?

src/arch/common/forte_specific_architecture.h
```diff
@@ -0,0 +1,26 @@
...
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Jose Cabral - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+#ifndef SRC_ARCH_COMMON_FORTE_SPECIFIC_ARCHITECTURE_H_
+#define SRC_ARCH_COMMON_FORTE_SPECIFIC_ARCHITECTURE_H_
```
cochicde: this file should probably go under `arch` and keep here only cpp/hpp files which might be common to different architectures

franz-hoepfinger-4diac:
i test the FreeRTOS now.

franz-hoepfinger-4diac:
so, 
Compilation fails: 
https://gist.github.com/franz-hoepfinger-4diac/108f6b91c1113aae71754145c77b8276


` undefined reference to 'forteJoinInstance'`



now i changed my forte Component in ESP32 to this: 
https://gitlab.com/meisterschulen-am-ostbahnhof-munchen/esp32-4diac-example-application-no-lua/-/tree/develop/Application/components/forte_main?ref_type=heads

and it compiles again. 

i will now test if it is working.

cochicde:
> so, Compilation fails: https://gist.github.com/franz-hoepfinger-4diac/108f6b91c1113aae71754145c77b8276

I changed the interface to reflect properly what the functions are doing and more closely related to the Device functions. Also, ideally you should be able now to have everything you need by including the file `src/arch/forte.h` and `src/common/devlog.h`

`undefined reference to `forteJoinInstance' ` -> `forteJoinInstance` is now `forteWaitForInstanceToStop`. A possible implementation of `main` for freeRTOS is here https://github.com/eclipse-4diac/4diac-forte/blob/853c75649900ea3c9c4305adc88b050618ce2e55/src/arch/freeRTOS/main.cpp

diplfranzhoepfinger:
> `undefined reference to` forteJoinInstance' `->`forteJoinInstance`is now`forteWaitForInstanceToStop`. A possible implementation of` main\` for freeRTOS is here [https://github.com/eclipse-4diac/4diac-forte/blob/853c75649900ea3c9c4305adc88b050618ce2e55/src/arch/freeRTOS/main.cpp](https://github.com/eclipse-4diac/4diac-forte/blob/853c75649900ea3c9c4305adc88b050618ce2e55/src/arch/freeRTOS/main.cpp)

which is nice, but not so practical. 

[https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/freertos.html#application-entry-point](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/freertos.html#application-entry-point) 

`main` is not the App Entry Point, but `app_main` but also this is normally not true for 4diac, as you might want to start things like Network Stack, Wifi, Webserver etc.. before.

cochicde:
> > `undefined reference to` forteJoinInstance' `->`forteJoinInstance`is now`forteWaitForInstanceToStop`. A possible implementation of` main` for freeRTOS is here https://github.com/eclipse-4diac/4diac-forte/blob/853c75649900ea3c9c4305adc88b050618ce2e55/src/arch/freeRTOS/main.cpp
> 
> which is nice, but not so practical. 
> 
> https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/freertos.html#application-entry-point 
> 
> `main` is not the App Entry Point, but `app_main` but also this is normally not true for 4diac, as you might want to start things like Network Stack, Wifi, Webserver etc.. before.

I see, but the `app_main` is something specific to `ESP-IDF` and not something generic to `freeRTOS`. The `main.cpp` file here in forte is mostly for guidance purposes, as you said other things might need to be done before and each application might do it differently. 

The way you use it hasn't change much, just the signatures of the methods and the `includes` should be different now.

kaartee:
Just letting you know that I wouldn't be able to review this PR with regard to the Zephyr port in the near future. I'm sorry.

diplfranzhoepfinger:
> I see, but the `app_main` is something specific to `ESP-IDF` and not something generic to `freeRTOS`. 

wrong. 

it is vanilla FreeRTOS: 

https://github.com/FreeRTOS/FreeRTOS/blob/bf046c15f756ac204071bdac04637a3a9d5344ed/FreeRTOS/Demo/CORTEX_M0%2B_NUCLEO_L010RB_GCC_IAR/Demo/app_main.c#L51

> The `main.cpp` file here in forte is mostly for guidance purposes, as you said other things might need to be done before and each application might do it differently.

true

> The way you use it hasn't change much, just the signatures of the methods and the `includes` should be different now.

kaartee:
To compile with a C (not C++) main source file:

`src/arch/forte.h`:
```
@@ -38,7 +38,7 @@ extern "C" {
   * @param paResultInstance Address  to store the created forte instance
   * @return FORTE_OK if no error occurred, other values otherwise
   */
-  FORTE_SHARED_PREFIX FORTE_STATUS forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance);
+  FORTE_SHARED_PREFIX enum FORTE_STATUS forteStartInstance(unsigned int paPort, TForteInstance* paResultInstance);

  /**
   * \brief Start forte instance with posibilities of more arguments
@@ -47,7 +47,7 @@ extern "C" {
   * @param paResultInstance Address  to store the created forte instance
   * @return FORTE_OK if no error occurred, other values otherwise
   */
-  FORTE_SHARED_PREFIX FORTE_STATUS forteStartInstanceGeneric(int argc, char *argv[], TForteInstance* paResultInstance);
+  FORTE_SHARED_PREFIX enum FORTE_STATUS forteStartInstanceGeneric(int argc, char *argv[], TForteInstance* paResultInstance);

  /**
   * \brief Request termination of a Forte instance
```

kaartee:
With above change, this PR builds for Zephyr. I haven't started or tested it any other way yet, though.

cochicde:
> With above change, this PR builds for Zephyr. I haven't started or tested it any other way yet, though.

thanks! that's was helpful. I added a typedef to fix that issue

cochicde:
@azoitl this is ready now

azoitl (commented):
I think the startuphook.h should not be moved. The rest look ok. Although at a first glance a bit complicated.

cochicde:
> I think the startuphook.h should not be moved. The rest look ok. Although at a first glance a bit complicated.

I saw that `startupHook.h` (and its `cpp.in`) don't have anything related to the `arch` folder. I see the `arch` folder as the place for code that may vary for each platform, so I didn't see the any difference from `forteinit.cpp.in` or `stringlist.cpp.in` which are located directly under `src`. 

I see the complication too. Maybe I was too greedy to try to make it really platform abstract that it became complex. One thing I see that is too much for example is the `main_helper`. My problem was that `windows` doesn't have one of the three signals we're using for all other cases. We could just have a `#ifndef WIN32` or similar and include that signal in a simpler `main.cpp` file. what do you think?

azoitl:
> > I think the startuphook.h should not be moved. The rest look ok. Although at a first glance a bit complicated.
> 
> I saw that `startupHook.h` (and its `cpp.in`) don't have anything related to the `arch` folder. I see the `arch` folder as the place for code that may vary for each platform, so I didn't see the any difference from `forteinit.cpp.in` or `stringlist.cpp.in` which are located directly under `src`.

the startuphook is and shall only be used in architecture. Thats the reason I originally put the h file there. code generation is correcter and easier in the main src directory therefore I kept it there. placing the .h file in src may imply others can or shall use it. Which I find a bit missleading. 

> 
> I see the complication too. Maybe I was too greedy to try to make it really platform abstract that it became complex. One thing I see that is too much for example is the `main_helper`. My problem was that `windows` doesn't have one of the three signals we're using for all other cases. We could just have a `#ifndef WIN32` or similar and include that signal in a simpler `main.cpp` file. what do you think?

Originally I thought it was the way how github presented your changes therefore I fetched them to look at them locally. However it didn't get better. To understand how your new code works I have to look to at least 8 files, maybe more. A 4diac FORTE internal cpp file now uses very strange and abstract C functions instead of a more C++ style. And even there stuff is separated into different parts that I think may not be necessary. 

For me the C functions are there to allow to create a 4diac FORTE instance from C code, e.g., when using it as a lib or linking it to an small os. 

The endhook is and never was platform independent. It was only for posix platforms. Here one size fits all is definitely wrong. 

However, I think you did here is an important work to make the 4diac platform specific start-up code less duplicated and better maintainable. But we are not fully there yet.

cochicde:
Just organizing the points for easy discussing them:
1. `startuphook`: I see your point in the sense that no one else should use it. I'll move it back
2. `endhook`: this is also being used by the windows platform, that's the reason I tried to abstract it. We could have one main for each posix and another for windows, but they are so similar and therefore we end up with duplicated code. I'm not sure what to do with it
3. C vs C++ inteface: You're right, I think I pushed it too far. My main idea (pun intended) was to have the main also test the C interface, but I could move this C-style main into the test folder, basically in order to test this interface. I see two problems still regarding this:
   - extra code will be compiled: as this PR is now, forte as a library offers the C interface basically so anyone can use it (meaning C compilers can work with it). To offer this interface, the "implementation" of this interface (`forte.cpp` now)  where all the C style creation of forte happens, uses  the `C4diacFORTEInstance` class "bridging" to C++. If we have the main in a C++ style, basically as it is outside this PR, then it looks very similar as the `forte.cpp` which is then not used and you endup with unused code (which is not that bad). 
   
    While writing this message, I thought about the following: we could have a `forte` library which includes all until `C4diacFORTEInstance` and none of the C interface. Then, another `forte_c` library (maybe under a CMake parameter to enable it) which basically adds these C-interface file and links to `forte`. The `main` in `arch` in C++ style would then link to the `forte` library, and a `main` in the `test` folder would test the `forte_c` library.  
   - `forte.h` name: the naming of this file can be confusing. I thought `forte_c_interface.h` for clarity, what do you think?

azoitl:
ad 2)  me neither

ad 3) your last proposal sounds quite good. I also have no problem that not all code is used in all cases. There are different options for different use-cases. And depending on what I need I would pull in the right one.

cochicde:
> Just organizing the points for easy discussing them:
> 
> 1. `startuphook`: I see your point in the sense that no one else should use it. I'll move it back
> 2. `endhook`: this is also being used by the windows platform, that's the reason I tried to abstract it. We could have one main for each posix and another for windows, but they are so similar and therefore we end up with duplicated code. I'm not sure what to do with it
> 3. C vs C++ inteface: You're right, I think I pushed it too far. My main idea (pun intended) was to have the main also test the C interface, but I could move this C-style main into the test folder, basically in order to test this interface. I see two problems still regarding this:
>    
>    * extra code will be compiled: as this PR is now, forte as a library offers the C interface basically so anyone can use it (meaning C compilers can work with it). To offer this interface, the "implementation" of this interface (`forte.cpp` now)  where all the C style creation of forte happens, uses  the `C4diacFORTEInstance` class "bridging" to C++. If we have the main in a C++ style, basically as it is outside this PR, then it looks very similar as the `forte.cpp` which is then not used and you endup with unused code (which is not that bad).
>    
>    While writing this message, I thought about the following: we could have a `forte` library which includes all until `C4diacFORTEInstance` and none of the C interface. Then, another `forte_c` library (maybe under a CMake parameter to enable it) which basically adds these C-interface file and links to `forte`. The `main` in `arch` in C++ style would then link to the `forte` library, and a `main` in the `test` folder would test the `forte_c` library.
>    
>    * `forte.h` name: the naming of this file can be confusing. I thought `forte_c_interface.h` for clarity, what do you think?

I implemented the things presented here:
1. `startuphook` was moved back where it was under `arch`
2. `main.cpp` is using c++ style code as it was before using `C4diacFORTEInstance`.
  -  I removed all the complex stuff surrounding `endhook` and now everything is back in `main.cpp` and for the special signal in posix, I just have a special `#ifndef WIN32`
  - Some PLCNext startup code was moved into the PLCNext module using the startuphook
  - forteinit was moved into the architecture initialization
  - Some windows code in `fdhandler` was moved into the architecture initialization of windows.
3. C Interface:
  - The C interface code was moved to `arch/c_interface` and are included only on demand via cmake.
  - They are linked to `FORTE_LITE` and create a static and/or shared library depending on cmake variables
  -  Tests were added to make sure that an executable can link to these new libraries and perform some test on the interface itself
  - Similar to this, "compilation" tests were added to make sure that the static and shared libraries of forte can be linked to (not related to the c-interface libraries)
4. Re-organization of target preparation in `src/CMakeListst.txt`:
  - First, all the files to compile and configurations are figured out, and only then the targets are prepared. 
  - For the target preparation, `FORTE_LITE` is used as a based target where all confurations are set (compilation flags, link flags, libraries, directories, ....).
  - All other targets (forte exectuable, forte static/shared libraries, c interface static/shared libraries) link to `FORTE_LITE` and doing that, they "inherit" all the configurations from it.
 

Things not resolved: 
- in windows, the shared libraries are different and I could not make them work properly for some reason, so I excluded for now the tests for it.
- The c-interface shared library: I wanted to keep the exported symbols only limited to the C interface to make its loading time shorter, but still all symbols are exported. It's not a show stopper. 
-

src/modules/PLCnext/CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ forte_add_link_library_beginning(Arp.Plc.AnsiC)
 
 forte_add_link_flags("-Wno-undef")
 
+forte_add_startup_hook(DeviceStatus::startup)
+
+
```
dok-net: redundant blank line

src/arch/common/forte_specific_architecture.cpp
```diff
@@ -1,27 +1,21 @@
 /*******************************************************************************
- * Copyright (c) 2017 fortiss GmbH
+ * Copyright (c) 2024
```
azoitl: please put your name as the copyright holder of this file

azoitl (changes requested):
I think this is now a very nice result. Thx for all the effort.

The only thing that needs to be done is that you update the copyright headers of all touched files.  Add there 2024 as the last change year and as you are an individual contributor your name as copyright holder.

src/arch/common/forte_specific_architecture.cpp
cochicde: done

cochicde:
tests are failing because of this https://github.com/eclipse-4diac/4diac-forte/commit/2a0a3e410a47cbda302b5010f6e146e851e947d2 which should be fixed here #195

azoitl:
> tests are failing because of this [2a0a3e4](https://github.com/eclipse-4diac/4diac-forte/commit/2a0a3e410a47cbda302b5010f6e146e851e947d2) which should be fixed here #195

I merged the fix back could you please rebase sot hat we can have the unit test chekcing your pr.

src/arch/c_interface/forte_c.cpp
```diff
@@ -11,42 +11,38 @@
...
-#include "../../stdfblib/ita/RMT_DEV.h"
+#include "forte_c.h"
 
 #include "../utils/mainparam_utils.h"
+#include "forte_architecture.h"
+#include "forte_printer.h"
+#include "forteinstance.h"
 
-unsigned const int cgForteDefaultPort = 61499;
+static const unsigned int scForteDefaultPort = 61499;
```
MartinMelikMerkumians: the coding guidelines doesn't know the prefix sc for static const, only cg for const global.

While file-scope globals theoretically have less impact, they are only an `extern` keyword, used by someone else, away being visible in other compilation units as well.

If we want to code the difference of usage into the name, I would rather suggest cf for "constant file" instead of 
sg "static const", as static const doesn't imply anything than the variable is some sort of global variable and lives in this compilation unit. "cf" would also be more in line with "cg" for const global,

i will mark this only here, to not split up the discussion on several review comments, but my comment is valid for all "sc" changes 

@azoitl @mx990 additional opinions are appreciated

MartinMelikMerkumians: I would also recommend to add the "U" suffix to the number, as its an unsigned constant

src/arch/main.cpp
```diff
@@ -52,27 +38,35 @@ int main(int argc, char *arg[]){
...
-  startupHook(argc, arg);
+  if(auto result = CForteArchitecture::initialize(argc, arg); result != 0){
+     return result;
+  }
 
-#ifdef FORTE_IO_PLCNEXT
-  sleep(3);
-  DeviceStatus::startup();
-#endif
+  struct callOnExit {
```
MartinMelikMerkumians: Open for discussion, but do we really want to use structs as more than plain data containers? @azoitl @mx990 
From my perspective, its would be clearer to use classes everywhere we need more than a collection of data entries, and reserve the use of structs exactly for that, plain data containers.
My rationale for this is, that you know what to expect if you see class or struct that way, without needing to check the code.
The cost is quite low, only adding `public:` as additional line here

src/arch/main.cpp
```diff
@@ -52,27 +38,35 @@ int main(int argc, char *arg[]){
...
+    int (*onExit)();
+    ~callOnExit() { onExit(); }
+  };
+
+  callOnExit{CForteArchitecture::deinitialize};
 
   hookSignals();
 
   const char *pIpPort = parseCommandLineArguments(argc, arg);
```
MartinMelikMerkumians: could you please remove the prefixed `p` here, as the coding guidelines abolished the use of Hungarian notation for most parts

src/arch/pikeos_posix/forte_specific_architecture.cpp
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Jose Cabral - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+#include "arch/forte_specific_architecture.h"
+
+/* CG: testing */
```
MartinMelikMerkumians, line 15: what does this comment mean?

tests/arch/forte_c_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <chrono>
+#include <thread>
+
+BOOST_AUTO_TEST_CASE(forte_c_interface)
+{
+  TForteInstance instance{nullptr};
+
+  // forteStartInstance
+  // invalid instance result
+  BOOST_CHECK_EQUAL(forteStartInstance(61499, nullptr), FORTE_WRONG_PARAMETERS);
```
MartinMelikMerkumians: Could you please use `BOOST_TEST` instead of `BOOST_CHECK`, as it results in better test failure output and is nicer to read (at least in my opinion)

tests/core/trace/ctfTracerTest.cpp
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 fotiss GmbH
+ * Copyright (c) 2024 Jose Cabral
```
MartinMelikMerkumians, line 2: I am not sure if changing the copyright here from Fortiss GmbH to exclusively Jose Cabral is valid here, as much of the original code is still present.
Even if you wrote the code originally, it seems you transferred the copyright to Fortiss back then.

@azoitl

tests/core/trace/ctfTracerTest.cpp
```diff
@@ -105,7 +105,7 @@ BOOST_AUTO_TEST_CASE(sequential_events_test) {
     command.mFirstParam.clear();
     command.mFirstParam.pushBack(counterInstanceName);
     command.mFirstParam.pushBack(g_nStringIdPV);
-    BOOST_CHECK(EMGMResponse::Ready == resource->writeValue(command.mFirstParam, CIEC_STRING(std::string("1")), false));
+    BOOST_CHECK(EMGMResponse::Ready == resource->writeValue(command.mFirstParam, std::string("1"), false));
```
MartinMelikMerkumians: please use `BOOST_TEST` and std::string literals here `"1"s`

MartinMelikMerkumians (changes requested):

src/arch/c_interface/forte_c.cpp
mx990: I would strongly suggest to use an [unnamed namespace](https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces) in C++ instead of `static` linkage.

src/arch/main.cpp
mx990: Why not change `CForteArchitecture::deinitialize` to a destructor and use a local variable of type `CForteArchitecture` to avoid the additional struct? Alternatively, there is [std::atexit](https://en.cppreference.com/w/cpp/utility/program/atexit).

src/arch/forte_architecture.h
```diff
@@ -15,20 +15,80 @@
 #ifndef SRC_ARCH_FORTE_ARCHITECTURE_H_
 #define SRC_ARCH_FORTE_ARCHITECTURE_H_
 
-class CForteArchitecture{
+#include "forte_specific_architecture.h"
+
+#include "forteinit.h"
+#include "startuphook.h"
+
+template<typename SpecificArchitectureT>
```
mx990: It seems overkill to me to use a template parameter on top of conditional compilation. What is the benefit instead of just using `CForteSpecificArchitecture` directly?

src/arch/c_interface/forte_c.h
mx990, line 1: This file does not seem to be architecture-dependent. Should this not rather live directly in `src`?

src/modules/PLCnext/CMakeLists.txt
cochicde: done

src/arch/c_interface/forte_c.cpp
cochicde: I went with the `unnamed namespace` since it's more c++ style. Regarding the naming convetion, since there's no defined convention for constant variables in unnames namespaces I didn't add any prefix to avoid having a "convention" only in that file which is not documented anywhere

src/arch/main.cpp
cochicde: I like the `std::atexit` since it applies in the context of the `main` function in a c++ environment.

src/arch/main.cpp
cochicde: done

src/arch/pikeos_posix/forte_specific_architecture.cpp
cochicde: I don't know. I only moved things around, so this came from the original code

tests/arch/forte_c_test.cpp
cochicde: How does the better output look like? In my opinion, if this test fails you get the output of the actual vs expected value which helps you understand why the test failed, which I believe you don't have if you just use `BOOST_TEST`

tests/core/trace/ctfTracerTest.cpp
cochicde: This was a correction from a commit from a couple of months ago actually. The original file is not from 2019 and not done at fortiss, but by me this year

tests/core/trace/ctfTracerTest.cpp
cochicde: done

src/arch/c_interface/forte_c.h
cochicde: I agrre, but before moving it I would leave that for discussion mainly because there was a similar case somewhere up there in the comments regarding the `startuphook` file

@azoitl

src/arch/forte_architecture.h
cochicde: My main motivation was usability. When you are using the architecture inside forte, the specifics of the architecture should not be visible to the user. By having a "general" architecture, we can define a proper interface with a clear documentation of what it does. 

Specific architectures might add variables or other private methods which might distract the user. 

Also, when creating a new architecture, you should only look at the "general" architecture and nothing else.

mx990: I guess I worded that poorly. I am not against splitting this into a general and specific part, far from it. I am merely wondering why we need a template parameter instead of just using `CForteSpecificArchitecture` **inside** `CForteGeneralArchitecture`.

tests/arch/forte_c_test.cpp
mx990: [BOOST_TEST](https://www.boost.org/doc/libs/1_84_0/libs/test/doc/html/boost_test/testing_tools/boost_test_universal_macro.html) handles expressions and produces output including the full expression with individual values.

src/arch/c_interface/forte_c.h
azoitl: currently we have everything that is  used to start 4diac FORTE on a specific device in the arch folder. Therefore I would like to keep it here. Because when I'm porting 4diac FORTE to a new device I would look here first.

src/arch/c_interface/forte_c.cpp
MartinMelikMerkumians: I like the unnamed namespace solution. Perhaps it wasn't that clear in my comment, if we feel we need another prefix to show the intent/usage we could add one to the coding guidelines. 
I wasn't suggesting using another undocumented prefix in favor of another

tests/arch/forte_c_test.cpp
cochicde: true, but I get the exact same output. If  I have invalid cases as in:

```
    BOOST_CHECK_EQUAL(forteGlobalInitialize(validArgc, validArgV), 1);
    BOOST_TEST(forteGlobalInitialize(validArgc, validArgV) == 1);
```

I get 

```
~/4diac-forte/tests/arch/forte_c_test.cpp(56): error: in "forte_c_interface": check forteGlobalInitialize(validArgc, validArgV) == 1 has failed [0 != 1]
~/4diac-forte/tests/arch/forte_c_test.cpp(57): error: in "forte_c_interface": check forteGlobalInitialize(validArgc, validArgV) == 1 has failed [0 != 1]
```

maybe I missed something, but I don't see any better output

src/arch/forte_architecture.h
cochicde: I see. I completely missed that. You're right. It's updated now

tests/arch/forte_c_test.cpp
MartinMelikMerkumians: In this simple case, the output seems to be the same, but in my personal experience of writing tests in the past, i found BOOST_TEST generating better test output, but your mileage may vary.

Additionally as mentioned in my first comment about this, one of my reasons to stick to `BOOST_TEST` is the better readability of the check itself.
Just to explain my reasoning, with `BOOST_CHECK_EQUAL` I first get the information the parameters are to be compared for equality, and only after I analyse the parameters (assuming reading from left to right), needing to hold the information of what kind of comparison needs to be performed in my head.
For `BOOST_CHECK_GT` etc. its even worse, as I need to know the order of operators and keep them in mind while checking the parameter expression.

On the other hand, `BOOST_TEST` let you write the expression to be checked in native C/C++ and I can see them in "natural" order of the language.
In addition, I find it easier to only need to know (usually) one test macro, and ignoring it as code noise, focusing on the expression itself.

Also as said above, this is subject to personal preference, and open to discussion of course.
We should also write the results down into our coding guidelines, so the consensus is documented and publicly available in the future.

mx990: There is a difference when using arithmetic expressions. Please compare the output for:
```c++
int a = 3;
int b = 2;
BOOST_TEST(a + 1 == b);
BOOST_CHECK_EQUAL(a + 1, b);
```
which prints:
```console
test.cpp(8): error: in "test_macro_overview": check a + 1 == b has failed [3 + 1 != 2]
test.cpp(9): error: in "test_macro_overview": check a + 1 == b has failed [4 != 2]
```

> On the other hand, `BOOST_TEST` let you write the expression to be checked in native C/C++ and I can see them in "natural" order of the language. In addition, I find it easier to only need to know (usually) one test macro, and ignoring it as code noise, focusing on the expression itself.

I also personally prefer the ability to write arbitrary conditions in `BOOST_TEST` and let boost figure it out, without having to worry about which test function to use. I find this particularly useful when copy-pasting existing conditions from somewhere.

> Also as said above, this is subject to personal preference, and open to discussion of course. We should also write the results down into our coding guidelines, so the consensus is documented and publicly available in the future.

I would like to add, that according to the boost sources, the `BOOST_CHECK*` macros seem to be deprecated, see
https://github.com/boostorg/test/blob/b0127f48d23b94387c9e57b6b481f9119a3a25db/include/boost/test/tools/interface.hpp#L273-L308. When disabling the "old tools" by defining `BOOST_TEST_NO_OLD_TOOLS`, they simply forward to `BOOST_TEST_CHECK` (as shown in the linked snippet and the output above then becomes identical).

MartinMelikMerkumians: > I would like to add, that according to the boost sources, the BOOST_CHECK* macros seem to be deprecated, see
> https://github.com/boostorg/test/blob/b0127f48d23b94387c9e57b6b481f9119a3a25db/include/boost/test/tools/interface.hpp#L273-L308. When disabling the "old tools" by defining BOOST_TEST_NO_OLD_TOOLS, they simply forward to BOOST_TEST_CHECK (as shown in the linked snippet and the output above then becomes identical).

Thanks for finding this, I thought I have read that BOOST_CHECK* have been deprecated, but couldn't find that anymore

cochicde: very interesting. indeed the output is better. I'll change all to `BOOST_TEST`

cochicde: done. One thing that I realized while changing now to `BOOST_TEST` is that you need the `<<` operator for the type you're comparing. 

For example, 

` BOOST_TEST(EMGMResponse::Ready == resource->writeValue(command.mFirstParam, "1", false));`
 does not compile and it requires 

```
std::ostream& operator<<(std::ostream& os, const EMGMResponse& val ){
  // ...  
 return os;
}
```

MartinMelikMerkumians: Boost.test gives you the possibility to add a helper function (https://github.com/eclipse-4diac/4diac-forte/blob/develop/tests/forte_boost_output_support.h) instead of the stream operator, in cases where it doesn't make sense for the type itself.

cochicde: wow! I'm completely out of touch regarding `boost`. I updated accordingly

src/CMakeLists.txt
```diff
@@ -291,78 +244,193 @@ if(NOT DEFINED CMAKE_CXX_STANDARD_REQUIRED)
...
 if(existing_link_flags)
   set_property(GLOBAL APPEND PROPERTY FORTE_LINK_FLAGS ${existing_link_flags})
 endif()
 
 GET_PROPERTY(link_flags GLOBAL PROPERTY FORTE_LINK_FLAGS)
+list(JOIN link_flags " " link_flags_string)
 
-add_library(FORTE_LITE OBJECT ${SOURCE_FILES}) #no executables.
-TARGET_LINK_LIBRARIES (FORTE_LITE ${LINK_LIBRARY})
+target_link_options(${FORTE_BASE_TARGET_NAME} INTERFACE "${link_flags_string}")
```
ernstblechaPT: target_link_options will pass the given string as a quoted parameter to the linker, which turns the link_flags_string to a single parameter; this breaks compiling a static forte (on windows with mingw) and also setting additional linker options

you can just pass the list directly to target_link_options without joining it first:
```
target_link_options(${FORTE_BASE_TARGET_NAME} INTERFACE ${link_flags})
```

src/arch/win32/CMakeLists.txt
ernstblechaPT, line 1: line 74/73 (`forte_add_link_flags("-static-libgcc -static-libstdc++"`):

needs to be adapted to `forte_add_link_flags("-static-libgcc" "-static-libstdc++")` otherwise this leads to a linker warning about unknown options

ernstblechaPT (changes requested):

src/CMakeLists.txt
```diff
@@ -363,7 +363,7 @@ if(FORTE_C_INTERFACE)
     set_target_properties(${TARGET_NAME}
        PROPERTIES PUBLIC_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/arch/c_interface/forte_c.h)
 
-    TARGET_LINK_LIBRARIES (${TARGET_NAME} PRIVATE ${FORTE_BASE_TARGET_NAME})
+    TARGET_LINK_LIBRARIES (${TARGET_NAME} PUBLIC ${FORTE_BASE_TARGET_NAME})
```
ernstblechaPT, line 366: I assume this is needed for the new C-Interface of FORTE to be linked correctly?

cochicde: exactly. The tests linking to these library weren't getting the right flags, in this case was the `-fsanitze=address` I think

ernstblechaPT (approved):

src/arch/freeRTOS/main.cpp
```diff
@@ -13,31 +13,38 @@
...
-#include "../forte_architecture.h"
-#include "../devlog.h"
-#include "forteinstance.h"
+#include "devlog.h"
+#include "../c_interface/forte_c.h"
 
-const static unsigned mainFORTE_TASK_PRIORITY = tskIDLE_PRIORITY + 1;
+static const unsigned scForteTaskPriority = tskIDLE_PRIORITY + 1;
+
+static const unsigned int scDesiredFortePort = 61499;
```
MartinMelikMerkumians: Please correct me if I am wrong, but didn't we agree to move this into an unnamed namespace and not to use prefixes here?

MartinMelikMerkumians (changes requested):

src/arch/freeRTOS/main.cpp
cochicde: you're right. It was in a similar file, but not this one. I updated this one for consistency.

azoitl (approved):

PR #156, opened 2024-06-12
Update fmuComLayer.cpp

franz-hoepfinger-4diac:
fix Typo


MartinMelikMerkumians:
Without is still spelled wrong, just now differently wrong

franz-hoepfinger-4diac:
> Without is still spelled wrong, just now differently wrong

thanks, should be good now.

azoitl (approved):

PR #158, opened 2024-06-13
Merge freeze into develop

azoitl:
In #131 the memset was removed, this results in random values showing up on the event counters


azoitl:
OPC UA system tests are failing again. As the other tests pass I'm merging.

PR #160, opened 2024-06-13
Deactivate OPC UA in System Tests

azoitl:


cochicde (approved):
I will start creating unit tests for opcua

PR #177, opened 2024-06-26
Replaced CIEC_STRING with std::string in SManagementCMD

azoitl:
SManagementCMD stillused a CIEC_STRING for mAdditionalParams. With this clean-up this is finally migrated to std::string.

### DIFF
--- a/src/core/datatypes/forte_any_string.h
+++ b/src/core/datatypes/forte_any_string.h
@@ -87,7 +87,7 @@ class CIEC_ANY_STRING : public CIEC_ANY_CHARS {
      *
      * @return number of bytes that this string has allocated for use
      */
-    TForteUInt16 getCapacity() const {
+    virtual TForteUInt16 getCapacity() const {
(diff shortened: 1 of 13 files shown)


MartinMelikMerkumians (commented):
Can you please check on resource.cpp lines 293, 306, and  316 if those conditions still make sense? Sorry, I cannot comment on them directly, as they have not been changed, but checking if the returned string fits into a CIEC_STRING seems to make no sense if a std::string is sed to save the value.

azoitl:
@MartinMelikMerkumians thx for noting. Yes this ifs where not needed in that form anymore. I also did a bit of clean-up around that.

azoitl:
Thx for noting. Yes this ifs can be simplified. Did also some further clean-ups.

MartinMelikMerkumians (approved):

MartinMelikMerkumians:
Thanks, looks good now

PR #179, opened 2024-07-02
Forte device restart option

MandKastner:
After forte receives a restart command (kill, reset, start) it will now start running again. The initial values are stored inside the resource during depolyment and are applied with a reset.

### DIFF
--- a/src/core/resource.cpp
+++ b/src/core/resource.cpp
@@ -164,20 +164,20 @@ EMGMResponse CResource::executeMGMCommand(forte::core::SManagementCMD &paCommand
 
 EMGMResponse CResource::changeFBExecutionState(EMGMCommandType paCommand){
   EMGMResponse retVal = CFunctionBlock::changeFBExecutionState(paCommand);
-  if(EMGMResponse::Ready == retVal){
+  if(retVal == EMGMResponse::Ready){
     retVal = changeContainedFBsExecutionState(paCommand);
-    if(EMGMResponse::Ready == retVal){
-      if(EMGMCommandType::Start == paCommand && nullptr != mInterfaceSpec) { //on start, sample inputs
+    if(retVal == EMGMResponse::Ready){
+      if(paCommand == EMGMCommandType::Start && mInterfaceSpec != nullptr) { //on start, sample inputs
         for(TPortId i = 0; i < mInterfaceSpec->mNumDIs; ++i) {
           CDataConnection *conn = *getDIConUnchecked(i);
-          if(nullptr != conn) {
+          if(conn != nullptr) {
             conn->readData(*getDI(i));
           }
         }
-      }else if(EMGMCommandType::Reset == paCommand){
+      }else if(paCommand == EMGMCommandType::Reset){
         setInitialValues();
       }
-      if(nullptr != mResourceEventExecution){
+      if(mResourceEventExecution != nullptr){
         // if we have a mResourceEventExecution handle it
         mResourceEventExecution->changeExecutionState(paCommand);
       }
--- a/src/core/resource.cpp
+++ b/src/core/resource.cpp
@@ -283,10 +283,6 @@ EMGMResponse CResource::writeValue(forte::core::TNameIdentifier &paNameList, con
 
 void CResource::setInitialValues() {
   CFunctionBlock::setInitialValues();
-  if(mInitialValues.empty()){
-    DEVLOG_ERROR("Error initializing values during reset\r\n");
-  }
-
   for(auto it : mInitialValues){
     it.getIECVariable().fromString(it.getInitString().c_str());
   }
--- a/src/core/resource.cpp
+++ b/src/core/resource.cpp
@@ -270,10 +270,7 @@ EMGMResponse CResource::writeValue(forte::core::TNameIdentifier &paNameList, con
             con->writeData(*var);
           }
         }else{
-          InitValue value;
-          value.ptr = var;
-          value.str = paValue.getStorage().c_str();
-          mInitList.push_back(value);
+          mInitialValues.emplace_back(*var, paValue.getStorage());
         }
         retVal = EMGMResponse::Ready;
       }else{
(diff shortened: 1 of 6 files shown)


src/core/resource.h
```diff
@@ -148,6 +148,15 @@ class CResource : public CFunctionBlock, public forte::core::CFBContainer{
       return nullptr;
     }
 
+    struct InitValue {
```
MartinMelikMerkumians: i would recommend, making this a regular class, with getters, setters, a ctor setting the two fields and making the fields const (assuming they are never changed after set at the beginning).
Then you can add a move operator for efficient insertion into the saved InitalValues.
Also, please rename the fields. A string called "str" repeats only its type information, but not its intend. I assume its a name of some sort, so please change the variable name to something suitable. Same for the pointer called ptr. How about iecVariable or similar here?


Personally I am not a big fan of nested classes, as the"advantage" of reduced visibility is usually not worth the cost of later maintenance (especially as visibility is also controlled by includes). If for some reason, this is later needed in other context, the additional work of extracting this can be substantial

src/core/resource.h
```diff
@@ -148,6 +148,15 @@ class CResource : public CFunctionBlock, public forte::core::CFBContainer{
...
+        CIEC_ANY* ptr;
+        std::string str;
+    };
+
+    std::vector<InitValue> mInitList;
```
MartinMelikMerkumians: I would recommend to rename this to mInitValues, as the name "List" implies a certain implementation class/style.

src/core/resource.cpp
```diff
@@ -174,6 +174,8 @@ EMGMResponse CResource::changeFBExecutionState(EMGMCommandType paCommand){
             conn->readData(*getDI(i));
           }
         }
+      }else if(EMGMCommandType::Reset == paCommand){
```
MartinMelikMerkumians: @azoitl correct me if I am wrong, but I thought stopped we have using Yoda conditions?

src/core/resource.cpp
```diff
@@ -277,6 +284,19 @@ EMGMResponse CResource::writeValue(forte::core::TNameIdentifier &paNameList, con
   return retVal;
 }
 
+void CResource::setInitialValues() {
+  CFunctionBlock::setInitialValues();
+  if(mInitList.empty()){
+    DEVLOG_ERROR("Error initializing values during reset\r\n");
+  }
+
+  for(std::vector<InitValue>::iterator it = mInitList.begin(); it < mInitList.end(); it++){
```
MartinMelikMerkumians: you could use "auto it" here, and let the compiler fill in the concrete type

src/stdfblib/events/E_RESTART.cpp
```diff
@@ -89,7 +89,9 @@ EMGMResponse FORTE_E_RESTART::changeFBExecutionState(EMGMCommandType paCommand){
   if(EMGMResponse::Ready == eRetVal){
     switch(paCommand){
       case EMGMCommandType::Start:
-        mEventToSend = (scmEventSTOPID == mEventToSend) ? scmEventWARMID : scmEventCOLDID;
+        if (scmEventWARMID != mEventToSend) {
```
MartinMelikMerkumians: I am not sure if this is the right approach here.
Just to verify, the current implemented behavior here is, that WARM is sent if the FB has been STOPPED earlier, and COLD is only sent if the E_RESTART FB has just been created, but never used before.

Now, how I understand the IEC 61131, a PLC can either get the command to COLD or WARM start anytime during its operation. Is this somehow handled then by the IDE? Like if a COLD start command is used, then everything is teared down on FORTE and then rebuild? Or is this handled somewhere else in FORTE if a COLD restart is requested by the user?

src/core/resource.cpp
mx990: and also use a [range-based for loop](https://en.cppreference.com/w/cpp/language/range-for)

src/core/resource.cpp
```diff
@@ -267,6 +269,11 @@ EMGMResponse CResource::writeValue(forte::core::TNameIdentifier &paNameList, con
             //if we have got a connection it was a DO mirror the forced value there
             con->writeData(*var);
           }
+        }else{
+          InitValue value;
+          value.ptr = var;
+          value.str = paValue.getStorage().c_str();
+          mInitList.push_back(value);
```
mx990: Why are you using `c_str()` when `getStorage()` already gives you a `std::string`? I would also suggest avoiding the intermediate local variable by using [emplace_back()](https://en.cppreference.com/w/cpp/container/vector/emplace_back) to construct the `InitValue` in-place.

src/core/resource.cpp
```diff
@@ -277,6 +284,19 @@ EMGMResponse CResource::writeValue(forte::core::TNameIdentifier &paNameList, con
   return retVal;
 }
 
+void CResource::setInitialValues() {
+  CFunctionBlock::setInitialValues();
+  if(mInitList.empty()){
+    DEVLOG_ERROR("Error initializing values during reset\r\n");
+  }
```
mx990: Is this always an error? What if there simply were no initial values set on this particular resource?

src/core/resource.h
mx990: I believe (the renamed) `ptr` could then also be a reference, which would save on `nullptr` checks.

src/core/resource.h
mx990: I would suggest `mInitialValues` since we also have the method `setInitialValues()`.

mx990 (changes requested):

src/stdfblib/events/E_RESTART.cpp
MandKastner: I think i misunterstood something here before..If i unterstand it correctly there is no such a thing like a warm or cold command, the restart function block just determines the event to be sent from the transitions of the state machine. In order to trigger RESET and setIntitialValues we first need to send KILL, RESET and START. I was wrong here because, if i read the standard correctly then this should also trigger a COLD restart event. the warm restart event should only occur if the block has been stopped but not killed. Is this correct?

MartinMelikMerkumians: With standard you mean the IEC 61499 I assume? If yes, I need to check on that first.
i am asking, because some part of the desired functionality is also based on the IEC 61131 (I think part 1), but perhaps I am understanding the intend here wrong.

mx990 (approved):

src/core/resource.cpp
azoitl: Yes we stopped that. @MandKastner please change to `(paCommand == EMGCommandType::Reset)`

azoitl (approved):

PR #181, opened 2024-07-03
Fix signature mismatches from #131 letting Zephyr target compile again

dok-net:
[@azoitl Here's the fix you expected to be obsoleted by another PR shortly, but as usual, things are taking a little longer than they should :-)](https://github.com/eclipse-4diac/4diac-forte/pull/180)

### DIFF
--- a/src/arch/zephyr/forte_Init.h
+++ b/src/arch/zephyr/forte_Init.h
@@ -63,13 +63,13 @@ extern "C" {
    * @param paSignal  Signal value to terminate instance
    * @param paInstance Instance to terminate
    */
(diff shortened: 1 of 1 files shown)


dok-net:
Fixing #151 to match this patch is a no-brainer, in the meantime, using FORTE with Zephyr is not as easy as it should for anyone using 4DIAC IDE milestones and FORTE develop branch, where I am not certain, but wouldn't be surprised that it takes a recent FORTE revision to work with the IDE?
@azoitl Could you please merge this miniscule fix despite the concerns you voiced recently?

azoitl:
Although nobody complained yet that 4diac FORTE has issues in Zephyr I get your point.

PR #185, opened 2024-07-04
Make Scripts for Windows build

franz-hoepfinger-4diac:
Make Scripts for Windows build

### DIFF
--- a/make_posix.sh
+++ b/make_posix.sh
@@ -7,7 +7,7 @@ export forte_bin_dir="bin/posix"
 #rm old folder
 rm -r "./$forte_bin_dir"
 
-#call setup_esp32c3.sh
+#call setup_posix.sh
 ./setup_posix.sh
 
 #goto new Directory
--- a/make_windows.cmd
+++ b/make_windows.cmd
@@ -1,25 +1,25 @@
-#!/bin/bash
+@echo off
 echo "----------------------------------------------------------------------------"
 
 
-export forte_bin_dir="bin/windows"
+set forte_bin_dir="bin\windows"
 
-#rm old folder
-rm -r "./$forte_bin_dir"
+::rm old folder
+rm -r "%forte_bin_dir%"
 
-#call setup_esp32c3.sh
-./setup_windows.sh
+::call setup_windows.sh
+sh.exe ./setup_windows.sh
 
-#goto new Directory
-cd "./$forte_bin_dir"
+::goto new Directory
+cd "%forte_bin_dir%"
 
-#make
+::make
 make
 
 
 echo "----------------------------------------------------------------------------"
 echo " DONE. to re-make just type:"
-echo "cd ./$forte_bin_dir"
+echo "cd %forte_bin_dir%"
 echo "make -j" 
 echo "----------------------------------------------------------------------------"
 
--- a/make_windows.cmd
+++ b/make_windows.cmd
@@ -2,7 +2,7 @@
 echo "----------------------------------------------------------------------------"
 
 
-set forte_bin_dir="bin\windows"
+set forte_bin_dir="bin\win32"
 
 ::rm old folder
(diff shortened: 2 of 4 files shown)


azoitl:
Do we really need that?

diplfranzhoepfinger:
> Do we really need that?

Without that we must work with the buggy CMAKE GUI. I hate it so much. 
If You have alternatives I am very open. 
But it should be Command line based not Clickie-Clickie based.

azoitl:
> > Do we really need that?
> 
> Without that we must work with the buggy CMAKE GUI. I hate it so much. If You have alternatives I am very open. But it should be Command line based not Clickie-Clickie based.

I know that you are not happy with the CMake GUI. Which as I told you, I don't understand at all. But I guess this is something we both will never come on par. 

That aside, the question was mostly targeted for @jwalt. Because your shell script is just duplicating what is now better available in fbe.

jwalt:
I agree with @azoitl we should start transitioning to 4diac-fbe instead. Technically it is pretty stable, but we probably will need to update/add docs for the new preferred way of building forte. FBE has configs for windows in the `configurations/tests` subdir

diplfranzhoepfinger:
> I know that you are not happy with the CMake GUI. Which as I told you, I don't understand at all. 

take only the fact that CMAKE GUI set the Loglevel back to NONE always, and nobody can solve it since 2 years .

this makes me anxious and unhappy, because such a tool is unreliable. What else can be set without manual intervention ? in a Textfile i can set things on and off, and can control this safely with git.

diplfranzhoepfinger:
> I agree with @azoitl we should start transitioning to 4diac-fbe instead.

yes, agree. 

>  Technically it is pretty stable, but we probably will need to update/add docs for the new preferred way of building forte. FBE has configs for windows in the `configurations/tests` subdir

4diac-fbe Build fails in release, and not even starts in develop. 

the build described in [https://eclipse.dev/4diac/en_help.php?helppage=html/installation/install.html#ownFORTE](https://eclipse.dev/4diac/en_help.php?helppage=html/installation/install.html#ownFORTE) runs. 

so, keep rocking, i would love to see 4diac-fbe working first time.

azoitl:
I'm regularly using FBE builds, I can not confirm your observation.

diplfranzhoepfinger:
> I'm regularly using FBE builds, I can not confirm your observation.

this is clear. you developed the System, you know how to operate it. 

i am new, i did it exactly according to the Manual on a fresh clone, and it not works in multiple ways. I ask for Help. Logs are provided in big variety.

azoitl:
> > I'm regularly using FBE builds, I can not confirm your observation.
> 
> this is clear. you developed the System, you know how to operate it. 

No I didn't I'm just a user like you.

> 
> i am new, i did it exactly according to the Manual on a fresh clone, and it not works in multiple ways. I ask for Help. Logs are provided in big variety.

I rechecked my setup. I'm using fbe release branch. However I'm providing the configuration to build to the configure.sh script. As I don't want to build all. Please have a look in the configurations directory for sample configurations you can use for your builds as starters. The default examples enable a lot. This is maybe not what you want.

diplfranzhoepfinger:
i think continue this Discussion in 

*   [https://github.com/eclipse-4diac/4diac-fbe](https://github.com/eclipse-4diac/4diac-fbe) 
*   eclipse-4diac/4diac-fbe#13

MartinMelikMerkumians:
> take only the fact that CMAKE GUI set the Loglevel back to NONE always, and nobody can solve it since 2 years .

I never had that issue in all the years using it (assuming you mean FORTE log level). You sure, that not something in the CMake configuration is doing it, like setting the build type to release?

Also, I am not aware of any bug report about this behavior. Correct my if I am wrong, but if there is no issue, nobody will investigate.

PR #192, opened 2024-07-19
Replace __ with :: in full type name of struct

MandKastner:


src/core/datatypes/forte_any_variant.cpp
```diff
@@ -316,10 +319,20 @@ size_t CIEC_ANY_VARIANT::getToStringBufferSize() const {
...
   }
   return result;
 }
 
+void CIEC_ANY_VARIANT::replaceUnderscoreInPackageName(char *paValue, size_t paLength) {
+  for (size_t i = 1; i < paLength - 1; i++) {
+    if (paValue[i] == '_' && paValue[i-1] == '_') {
+      paValue[i] = ':';
+      paValue[i-1] = ':';
+      i++;
```
azoitl, line 334: is this i++ really needed? I understand it is a bit feaster but harder to read.

mx990:
I don't quite understand why we are making these replacements on FORTE, when we do the opposite replacements in the IDE? Should we not rather stick to using `__` on FORTE for both management commands and monitoring values?

azoitl:
> I don't quite understand why we are making these replacements on FORTE, when we do the opposite replacements in the IDE? Should we not rather stick to using `__` on FORTE for both management commands and monitoring values?

Our philosphy so far was that 4diac FORTE on the interface always uses in its literals correct IEC 61131-3 syntax and hide implementation details. Did we change that for packages?

mx990:
Yes, until now, we used the mangled names (with `__`) on FORTE (e.g. when creating FBs, specifying structs, etc.) and did all the replacements in the IDE, see also [ForteTypeNameCreator](https://github.com/eclipse-4diac/4diac-ide/blob/develop/plugins/org.eclipse.fordiac.ide.deployment/src/org/eclipse/fordiac/ide/deployment/interactors/ForteTypeNameCreator.java). `CIEC_ANY_VARIANT::fromString` also does not perform any replacements and thus expects the type to use `__`.

PR #195, opened 2024-07-23
fix modify after free crash

ernstblechaPT:
mThreadHandle must not be reset before the Thread has rejoined


azoitl:
This breaks restarting 4diac FORTE resources via management commands. There we only stop the thread and later want to start it again. If only stop is called then `threadhandle` is never set to `nullHandle`. 

A potential solution is that `end()` cashes the `threadhandle` and resets it before `join()` and both `end()` and `join()` set it to `nullHandle`. Another option is that at the end of  `runThread()` it is set to `nullHandle`

ernstblechaPT:
Thank you for the feedback, i guess setting it to nullHandle at the end of runThread() is the most sensible solution

I wasn't sure if this might lead to a race if a thread is restarted later: until runThread() has finished the following start() will fail silently?

azoitl (approved):

PR #198, opened 2024-07-25
Use FBContainer for all Function Blocks

azoitl:
As basic and simple fbs can have FBs as children and to reduce a lot of code complexity in the FBContainer all FBs are now FBContainers.

PR #202, opened 2024-08-05
fix compilation of trace CTF

ernstblechaPT:
#198 broke the interface used by the trace CTF implementation

### DIFF
--- a/src/core/basicfb.cpp
+++ b/src/core/basicfb.cpp
@@ -161,11 +161,10 @@ size_t CBasicFB::getToStringBufferSize() const {
 
 #ifdef FORTE_TRACE_CTF
 void CBasicFB::traceInstanceData() {
-
   std::vector<std::string> inputs(mInterfaceSpec->mNumDIs);
   std::vector<std::string> outputs(mInterfaceSpec->mNumDOs);
   std::vector<std::string> internals(cmVarInternals ? cmVarInternals->mNumIntVars : 0);
-  std::vector<std::string> internalFbs(getInternalFBNum());
+  std::vector<std::string> internalFbs(getChildren().size());
   std::vector<const char *> inputs_c_str(inputs.size());
   std::vector<const char *> outputs_c_str(outputs.size());
   std::vector<const char *> internals_c_str(internals.size());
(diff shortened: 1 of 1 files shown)


src/core/basicfb.cpp
```diff
@@ -195,12 +204,16 @@ void CBasicFB::traceInstanceData() {
...
 
-  for(TPortId i = 0; i < internalFbs.size(); ++i) {
-    CFunctionBlock *value = getInternalFB(i);
-    std::string &valueString = internalFbs[i];
-    valueString.reserve(value->getToStringBufferSize());
-    value->toString(valueString.data(), valueString.capacity());
-    internalFbs_c_str[i] = valueString.c_str();
+  TPortId i = 0;
+  for(TFBContainerList::iterator it(getChildren().begin()); it != getChildren().end(); ++it){
+    if((*it)->isFB()) {
```
mx990: A basic FB can only contain internal FBs (or adapters), both of which happen to return `true` for `isFB()`, so I would suggest to simply use `getChildren()` as-is.

mx990 (changes requested):

src/core/basicfb.cpp
```diff
@@ -160,12 +160,19 @@ size_t CBasicFB::getToStringBufferSize() const {
...
+  }
+  return i;
+}
 
+void CBasicFB::traceInstanceData() {
   std::vector<std::string> inputs(mInterfaceSpec->mNumDIs);
   std::vector<std::string> outputs(mInterfaceSpec->mNumDOs);
   std::vector<std::string> internals(cmVarInternals ? cmVarInternals->mNumIntVars : 0);
-  std::vector<std::string> internalFbs(getInternalFBNum());
+  std::vector<std::string> internalFbs(internalFBCount(this));
```
mx990: I would suggest simply using `getChildren().size()` here.

src/core/basicfb.cpp
```diff
@@ -195,12 +202,14 @@ void CBasicFB::traceInstanceData() {
     internals_c_str[i] = valueString.c_str();
   }
 
-  for(TPortId i = 0; i < internalFbs.size(); ++i) {
-    CFunctionBlock *value = getInternalFB(i);
+  TPortId i = 0;
+  for(TFBContainerList::iterator it(getChildren().begin()); it != getChildren().end(); ++it){
```
mx990: I would suggest either iterating from `0` to `getChildren().size()` or using a [range-for](https://en.cppreference.com/w/cpp/language/range-for) loop.

mx990 (changes requested):

mx990 (approved):

ernstblechaPT:
seems like there still is a problem - need to look into that

appears to be this problem: https://github.com/eclipse-4diac/4diac-forte/pull/198#issuecomment-2266698110

PR #206, opened 2024-08-10
Turn interfacespec into a reference

azoitl:
As the interface spec is a key data element for FBs it always has to be there. By turning it into an reference this can be better ensured. Also it simplified a bit the interface spec hendling of generic fbs.


src/core/anyadapter.cpp
```diff
@@ -30,40 +30,25 @@ const SFBInterfaceSpec CAnyAdapter::scmFBInterfaceSpec = {
...
-  mCurrentFBInterfaceSpec.mEOWith = paPeer->getAdapterInterfaceSpec()->mEIWith;
-  mCurrentFBInterfaceSpec.mEOWithIndexes = paPeer->getAdapterInterfaceSpec()->mEIWithIndexes;
-  mCurrentFBInterfaceSpec.mNumDIs = paPeer->getAdapterInterfaceSpec()->mNumDOs;
-  mCurrentFBInterfaceSpec.mDINames = paPeer->getAdapterInterfaceSpec()->mDONames;
-  mCurrentFBInterfaceSpec.mDIDataTypeNames = paPeer->getAdapterInterfaceSpec()->mDODataTypeNames;
-  mCurrentFBInterfaceSpec.mNumDOs = paPeer->getAdapterInterfaceSpec()->mNumDIs;
-  mCurrentFBInterfaceSpec.mDONames = paPeer->getAdapterInterfaceSpec()->mDINames;
-  mCurrentFBInterfaceSpec.mDODataTypeNames = paPeer->getAdapterInterfaceSpec()->mDIDataTypeNames;
-  mCurrentFBInterfaceSpec.mNumDIOs = paPeer->getAdapterInterfaceSpec()->mNumDIOs;
-  mCurrentFBInterfaceSpec.mDIONames = paPeer->getAdapterInterfaceSpec()->mDIONames;
```
mx990, line 56: Is this not necessary to assign inputs to outputs and vice versa?

src/core/anyadapter.cpp
```diff
@@ -30,40 +30,25 @@ const SFBInterfaceSpec CAnyAdapter::scmFBInterfaceSpec = {
...
-
-  setupFBInterface(&mCurrentFBInterfaceSpec);
```
mx990, line 58: I believe `mCurrentFBInterfaceSpec` is now unused?

src/core/genfb.tpp
```diff
@@ -278,11 +270,11 @@ void CGenFunctionBlock<T>::freeFBInterfaceData(){
 }
 
 template<class T>
-void CGenFunctionBlock<T>::setupAdapters(const SFBInterfaceSpec *paInterfaceSpec, TForteByte *paFBData){
-  if((nullptr != paInterfaceSpec) && (nullptr != paFBData) && (paInterfaceSpec->mNumAdapters)) {
+void CGenFunctionBlock<T>::setupAdapters(const SFBInterfaceSpec& paInterfaceSpec, TForteByte *paFBData){
```
mx990: Could we also remove the parameter `paInterfaceSpec` here and use `getFBInterfaceSpec()` below instead?

src/core/genfb.tpp
```diff
@@ -154,116 +152,110 @@ size_t CGenFunctionBlock<T>::calculateFBVarsDataSize(const SFBInterfaceSpec &paI
...
 
-  T::mInterfaceSpec = const_cast<SFBInterfaceSpec *>(paInterfaceSpec);
-
-  if (nullptr != paInterfaceSpec) {
-    size_t connDataSize = calculateFBConnDataSize(*paInterfaceSpec);
-    size_t varsDataSize = calculateFBVarsDataSize(*paInterfaceSpec);
-    mFBConnData = connDataSize ? operator new(connDataSize) : nullptr;
-    mFBVarsData = varsDataSize ? operator new(varsDataSize) : nullptr;
+  size_t connDataSize = calculateFBConnDataSize(mGenInterfaceSpec);
+  size_t varsDataSize = calculateFBVarsDataSize(mGenInterfaceSpec);
```
mx990: Could we use `getFBInterfaceSpec()` instead of `mGenInterfaceSpec` to be consistent?

src/core/io/configFB/io_base.cpp
```diff
@@ -15,14 +15,14 @@
...
-IOConfigFBBase::IOConfigFBBase(forte::core::CFBContainer &paContainer, const SFBInterfaceSpec *paInterfaceSpec, const CStringDictionary::TStringId paInstanceNameId) :
+IOConfigFBBase::IOConfigFBBase(forte::core::CFBContainer &paContainer, const SFBInterfaceSpec &paInterfaceSpec, const CStringDictionary::TStringId paInstanceNameId) :
     CGenFunctionBlock<CEventSourceFB>(paContainer, paInterfaceSpec, paInstanceNameId) {
 }
 
 IOConfigFBBase::~IOConfigFBBase() = default;
 
 bool IOConfigFBBase::createInterfaceSpec(const char *, SFBInterfaceSpec &paInterfaceSpec) {
-  paInterfaceSpec = *mInterfaceSpec;
+  paInterfaceSpec = mInterfaceSpec;
```
mx990: I believe we can remove this assignment since `paInterfaceSpec` is already initialized in the constructor of `CGenFunctionBlock`?

src/stdfblib/ita/DEV_MGR.cpp
```diff
@@ -583,21 +583,21 @@ void DEV_MGR::appendIdentifierName(CIEC_STRING& paDest, forte::core::TNameIdenti
 DEV_MGR::DEV_MGR(CStringDictionary::TStringId paInstanceNameId, forte::core::CFBContainer &paContainer) :
     CCommFB(paInstanceNameId, paContainer, forte::com_infra::e_Server),
     mDevice(*paContainer.getDevice()) {
+  getGenInterfaceSpec() = scmFBInterfaceSpec;
```
mx990, line 586: Could we instead add a constructor to `CCommFB` and `CBaseCommFB` that allows to pass the interface spec to the `CGenFunctionBlock` constructor?

tests/core/mgmstatemachinetest.cpp
```diff
@@ -14,12 +14,13 @@
 #include "../../src/core/funcbloc.h"
 #include "fbcontainermock.h"
 
+const SFBInterfaceSpec gcEmptyInterface = {};
```
mx990: I would suggest making this a static constant in `CFunctionBlockMock` instead of a global variable.

tests/core/fbtests/fbtestfixture.cpp
```diff
@@ -153,7 +150,6 @@ void CFBTestFixtureBase::setup(const char* paConfigString){
 }
 
 bool CFBTestFixtureBase::createInterfaceSpec(const char *, SFBInterfaceSpec &paInterfaceSpec) {
```
mx990: I believe the parameter `paInterfaceSpec` is now unused?

mx990 (changes requested):

src/core/anyadapter.cpp
azoitl: Yes it is may bad. Fixed.

src/core/anyadapter.cpp
azoitl: yes it is removed.

src/core/genfb.tpp
azoitl: yes, good point!

src/core/genfb.tpp
azoitl: Good Idea!

src/core/io/configFB/io_base.cpp
azoitl: Yes I missed that.

I also think in the long run IOConfigFBBase should not be a configurable FB. But that is for another PR.

src/stdfblib/ita/DEV_MGR.cpp
azoitl: this is a very special case and I'm doing things in that block that I would not like to offer as API. With the work on different management intrfaces like OPC UA I also hope that the DEV_MGR will become obsolete. SO I don't think that API should be added for now.

tests/core/fbtests/fbtestfixture.cpp
azoitl: yes, fixed.

tests/core/mgmstatemachinetest.cpp
azoitl: good point.

src/stdfblib/ita/DEV_MGR.cpp
mx990: Makes sense to me.

mx990 (approved):

azoitl:
fixed merge conflicts with last force push.

PR #208, opened 2024-08-19
Add smart pointer for internal FBs

mx990:
Add a smart pointer to manage the lifecycle of internal FBs in an FB container.

### DIFF
--- a/src/core/fbcontainer.h
+++ b/src/core/fbcontainer.h
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2015, 2024 fortiss GmbH, Primetals Technologies Austria GmbH
+ *                          Martin Erich Jobst
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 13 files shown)


src/core/fbcontainer.h
```diff
@@ -155,6 +167,46 @@ namespace forte {
...
+                mFB(std::make_unique<T>(paInstanceNameId, paContainer)) {
+          mFB->configureFB(paConfigString);
+          paContainer.addFB(*mFB);
+        }
+
+        ~CInternalFB() {
+          mFB->getParent().removeFB(*mFB);
+        }
+
+        T *get() noexcept {
```
azoitl, line 190: do we really need access to the pointer?

src/core/fbcontainer.h
```diff
@@ -155,6 +167,46 @@ namespace forte {
...
+          return mFB.get();
+        }
+
+        T &operator*() noexcept {
+          return mFB.operator*();
+        }
+
+        T *operator->() noexcept {
```
azoitl, line 198: same

mx990: We need the `->` operator for access to members of the wrapped FB class (e.g., `fb_FB1->var_DI1`).

src/core/fbcontainer.h
mx990: When we do need a direct pointer to the wrapped FB class, this avoids constructs such as `&(*fb_FB1)`.

azoitl: make sense

src/core/fbcontainer.h
azoitl: ah forgot about that.

PR #209, opened 2024-08-19
Use variables for internal FBs in CFBs

mx990:


src/core/cfb.cpp
```diff
@@ -156,22 +154,6 @@ void CCompositeFB::executeEvent(TEventID paEIID, CEventChainExecutionThread * co
   }
 }
 
-bool CCompositeFB::createInternalFBs(){
```
azoitl, line 159: should we imediatly move this method to the lua CFB?

azoitl (approved):

PR #223, opened 2024-09-05
First version of a CVS reader block

azoitl:


src/modules/utils/GEN_CSV_READER_fbt.cpp
```diff
@@ -0,0 +1,189 @@
...
+    paInterfaceSpec.mDODataTypeNames = mDataOutputTypeIds;
+    return true;
+  }
+  return false;
+}
+
+void GEN_CSV_READER::openCSVFile() {
+  QO() = false_BOOL;
+  if(!mCSVFile.is_open()) {
+    mCSVFile.open(FILE_NAME().getStorage().c_str());
```
MartinMelikMerkumians, line 135: you are calling c_str() several times. I would suggest to store the result of the first call and use it later, as FILE_NAME will probably not modified in this function anymore

src/modules/utils/GEN_CSV_READER_fbt.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+void GEN_CSV_READER::readCSVFileLine() {
+  if(mCSVFile.is_open()) {
+    std::string line;
+    if(getline(mCSVFile, line)){
+      std::size_t oldPos = 0;
+
+      for(std::size_t i = 0;  i < getNumRD(); i++) {
+        std::size_t newPos = line.find(';', oldPos);
```
MartinMelikMerkumians, line 165: seeing this, you implicitly assume a German-language CSV file, correct?
English-language (and several others) CSVs use the comma ',' as the separator.
In the current version the use of the CSV Reader is limited to German-language (or others using the semicolon as separator) CSVs. I would suggest to extract the separator to an additional input for the FB, enhancing usability of the FB

MartinMelikMerkumians (changes requested):

diplfranzhoepfinger:
GOOD ! 

as said, i would like to make a INI-File Reader-Writer Block.

azoitl:
I don't think I find time in the near future to turn this into something better usable.

PR #226, opened 2024-09-13
Fix rounding for real to integer conversions

mx990:
Fix rounding for real to integer conversions to use the current rounding mode, which rounds towards the nearest even integer by default according to IEEE 754 (IEC 60559) and in conformance with IEC 61131-3.

### DIFF
--- a/src/core/datatypes/forte_lreal.cpp
+++ b/src/core/datatypes/forte_lreal.cpp
@@ -108,33 +108,10 @@ void CIEC_LREAL::castLRealData(const CIEC_LREAL &paSrcValue, CIEC_ANY &paDestVal
       // bit string will cast to the binary representation of the real value
       paDestValue.setValue(paSrcValue);
       break;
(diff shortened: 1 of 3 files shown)


azoitl (approved):

MartinMelikMerkumians:
Is the correct rounding mode guaranteed or should it be set at the start of forte to be save?

mx990:
> Is the correct rounding mode guaranteed or should it be set at the start of forte to be save?

It is the default rounding mode for binary floating point according to IEEE 754 section 4.3.3. It is also explicitly specified as the default for [gcc](https://www.gnu.org/software/libc/manual/html_node/Rounding.html) and [MSVC](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fegetround-fesetround2?view=msvc-170#remarks).

MartinMelikMerkumians (approved):

PR #67, opened 2024-04-02
Build with MQTT and Modbus Support in Github Workflow 

mhansal:

PR #228, opened 2024-09-16
Add open62541 and mqtt as separate actions into the github workflows

cochicde:
Merge parts of #66 and #67 but without modbus (to separate PRs) and without SSL for MQTT (since it is failing for [linux](https://github.com/eclipse-4diac/4diac-forte/actions/runs/10888263604/job/30212353315))

The main idea is to separate certains steps into their own files to have it more organized

### DIFF
--- a/.github/actions/mqtt/action.yml
+++ b/.github/actions/mqtt/action.yml
@@ -1,0 +1,52 @@
+name: Clone and compile paho-mqtt
+author: Jose Cabral
+Description: "Clone paho-mqtt and build a release static library with SSL disabled"
+
+inputs:
+  git_version:
+    type: string
+    Decription: Version of paho-mqtt to clone
+    required: True
+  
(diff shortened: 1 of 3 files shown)


azoitl:
I don't think we do need this any more. We now have the FBE and the FBE has a dedicated Jenkins build building all platforms with all libraries every day. So adding this here would only extend the build time.

cochicde:
> I don't think we do need this any more. We now have the FBE and the FBE has a dedicated Jenkins build building all platforms with all libraries every day. So adding this here would only extend the build time.

good to know. I suspected this but wasn't sure. I will close this then, because actually it was getting hard to make things work

azoitl:
that is the reason I'm happy that @jwalt provided [4diac FBE](https://github.com/eclipse-4diac/4diac-fbe). Definitely worth a try.

PR #239, opened 2024-09-22
fix and move ldsMe to its own class

cochicde:
This PR takes the fix from #75 which seems to have gone stale and moves all the logic related to the LDS-ME to its own class to have all encapsulated and make the LocalHandler lighter.

Pointers and deletions where also removed when possible and/or properly encapsulated for better memory managment.

### DIFF
--- a/src/com/opc_ua/CMakeLists.txt
+++ b/src/com/opc_ua/CMakeLists.txt
@@ -91,6 +91,8 @@ if (FORTE_COM_OPC_UA)
 
     add_subdirectory(FBs)    
     add_subdirectory(types)
+    add_subdirectory(detail)
+
     
     if(FORTE_COM_OPC_UA_ENCRYPTION)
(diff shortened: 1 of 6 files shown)


azoitl:
@kumajaya could you please have a look

cochicde:
Using 1.4.5, I couldn't make it work with the UA-LDS, but not even when trying to register the open62541 examples. It works properly with the open62541 example of lds server. The function which took care of registering itself every 10 minutes is gone (`UA_Server_addPeriodicServerRegisterCallback`), and technically this should be done on our side now, but I didn't see any official comments or updates on  this, so if needed this could be implemented in a separate PR

cochicde:
can we merge this as there's no answer? If there's still any error later, it would be easier to fix in the new class

m-meingast (approved):

azoitl (approved):

PR #241, opened 2024-09-25
Update compiler flags to remove "no deprecated warnig" declarations.

franz-hoepfinger-4diac:
Changed compiler flags to **exclude** "deprecated declarations" for C and C++ files in multiple toolchain configurations.

### DIFF
--- a/buildsupport/toolchains/toolchain-esp32-DualOut.cmake
+++ b/buildsupport/toolchains/toolchain-esp32-DualOut.cmake
@@ -4,8 +4,8 @@ SET(FORTE_ARCHITECTURE "FreeRTOSLwIP")
 
 
 
-set(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -ffunction-sections -fdata-sections -fno-threadsafe-statics -fno-rtti -fno-exceptions ")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ffunction-sections -fdata-sections -fno-threadsafe-statics -fno-rtti -fno-exceptions ")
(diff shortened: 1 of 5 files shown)


ernstblechaPT:
Could you give an example of the deprecated-warnings you receive for the esp32 platform? I'd rather not suppress those warnings as they are meant as a call to action for everyone using deprecated features.

cochicde:
> Could you give an example of the deprecated-warnings you receive for the esp32 platform? I'd rather not suppress those warnings as they are meant as a call to action for everyone using deprecated features.

I agree. What's the motivation to remove warnings? I think is better to remove the deprecated things that it's complaining about

mx990:
I believe the warnings were previously suppressed by `-Wno-deprecated-declarations` and removing the option re-enables them. From the gcc manual:
> -Wno-deprecated-declarations
    Do not warn about uses of functions, variables, and types marked as deprecated by using the deprecated attribute.

ernstblechaPT (approved):
@mx990 you are right, the change removes the flag. I completly misread the title and the message of the pull request

Maybe the title and the message of the pull request could be improved before merging.

franz-hoepfinger-4diac:
> Could you give an example of the deprecated-warnings you receive for the esp32 platform? I'd rather not suppress those warnings as they are meant as a call to action for everyone using deprecated features.

YES, i REMOVED the ` -Wno-deprecated-declarations` Option. 

so Deprecated Code will be SHOWN

franz-hoepfinger-4diac:
> Maybe the title and the message of the pull request could be improved before merging.

yes, this comes from using Chat GPT for generating this, sometimes this saves time, sometimes it leads to misunderstand. 
i updated it to be cleaner.

ernstblechaPT (approved):

azoitl (approved):

PR #252, opened 2024-09-29
adds a Timestamp SIFB

moniwe:


azoitl (approved):

MartinMelikMerkumians:
The same effect of timestamp should be achievable with now_monitonic and converting to an integer type via TIME_IN_NS_TO_<type>

Is there a special need to do this as an SIFB?

diplfranzhoepfinger:
> The same effect of timestamp should be achievable with now_monitonic and converting to an integer type via TIME_IN_NS_TO_
> 
> Is there a special need to do this as an SIFB?

I like the Question very much. 
My counter Question is: how to use NOW_MONOTONIC in graphical FB Programming? Seems only available in ST Programming in Textual Algorithms?

MartinMelikMerkumians:
> > The same effect of timestamp should be achievable with now_monitonic and converting to an integer type via TIME_IN_NS_TO_
> > Is there a special need to do this as an SIFB?
> 
> I like the Question very much. My counter Question is: how to use NOW_MONOTONIC in graphical FB Programming? Seems only available in ST Programming in Textual Algorithms?

If I didnt miss it, it seems nobody bothered until know to create an SIFB wrapper for those functions.
You could use them if such wrappers would be created

src/modules/utils/timing/Timestamp_fbt.cpp
```diff
@@ -0,0 +1,103 @@
...
+};
+
+void FORTE_influxDB__Timestamp::setInitialValues() {
+  var_timestamp = 0_ULINT;
+}
+
+void FORTE_influxDB__Timestamp::executeEvent(const TEventID paEIID, CEventChainExecutionThread *const paECET) {
+  switch(paEIID) {
+    case scmEventREQID:
+        var_timestamp = CIEC_ULINT(getNanoSecondsRealtime());
```
ernstblechaPT: at least for the win32 platform the value returned by `getNanoSecondsRealtime()` is (at least for now) not related to the UNIX epoch

https://learn.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps#general-faq-about-qpc-and-tsc

diplfranzhoepfinger:
> > I like the Question very much. My counter Question is: how to use NOW\_MONOTONIC in graphical FB Programming? Seems only available in ST Programming in Textual Algorithms?
> 
> If I didnt miss it, it seems nobody bothered until know to create an SIFB wrapper for those functions. You could use them if such wrappers would be created

just done it. see #253 and eclipse-4diac/4diac-ide#577

moniwe:
as I wrote this within the corresponding commit in 4diac-ide, I wrote this that I can push IO changes to an InfluxDB 2.0 with line protocol. Line protocol wants a number like ULINT, therefore TIME was not an option for me. I think I tried now monotonic and conversions but InfluxDB did not accept the results. Maybe I did not convert well enough, I am not completely sure about this, but this way the values were pushed to InfluxDB even if I got only UTC there I think.

mx990:
> as I wrote this within the corresponding commit in 4diac-ide, I wrote this that I can push IO changes to an InfluxDB 2.0 with line protocol. Line protocol wants a number like ULINT, therefore TIME was not an option for me. I think I tried now monotonic and conversions but InfluxDB did not accept the results. Maybe I did not convert well enough, I am not completely sure about this, but this way the values were pushed to InfluxDB even if I got only UTC there I think.

I believe you should be able to achive the same result with `TIME_IN_NS_TO_ULINT(NOW() - DT#1970-01-01-00:00:00)`. That should give you a timestamp in nanoseconds since the Unix Epoch.

MartinMelikMerkumians:
> > as I wrote this within the corresponding commit in 4diac-ide, I wrote this that I can push IO changes to an InfluxDB 2.0 with line protocol. Line protocol wants a number like ULINT, therefore TIME was not an option for me. I think I tried now monotonic and conversions but InfluxDB did not accept the results. Maybe I did not convert well enough, I am not completely sure about this, but this way the values were pushed to InfluxDB even if I got only UTC there I think.
> 
> I believe you should be able to achive the same result with `TIME_IN_NS_TO_ULINT(NOW() - DT#1970-01-01-00:00:00)`. That should give you a timestamp in nanoseconds since the Unix Epoch.

Sorry, for causing confusion, I didnt realize that you need a unix timestamp, and TIME_IN_NS_TO_ULINT only takes a TIME, but the solution presented by @mx990 should solve your issue.

diplfranzhoepfinger:
why still conflicts?

moniwe:
thanx @mx990 this works

mx990 (approved):

azoitl (approved):

PR #260, opened 2024-10-10
Upgrade FORTE to Open62541 1.4 release

m-meingast:
Update FORTE to support Open62541 1.4.x release

### DIFF
--- a/src/com/opc_ua/opcua_handler_abstract.cpp
+++ b/src/com/opc_ua/opcua_handler_abstract.cpp
@@ -14,7 +14,7 @@
 #include <forte_printer.h>
 #include <devlog.h>
 #include <parameterParser.h>
-#include <map>
+#include <unordered_map>
 
 //we tried to use the logLevelNames and logCategoryNames as extern from open62541 but it failed when using with shared libray
 const std::unordered_map<size_t, const char*> LogLevelNames {
--- a/src/com/opc_ua/opcua_handler_abstract.cpp
+++ b/src/com/opc_ua/opcua_handler_abstract.cpp
@@ -14,24 +14,32 @@
 #include <forte_printer.h>
 #include <devlog.h>
 #include <parameterParser.h>
+#include <map>
 
 //we tried to use the logLevelNames and logCategoryNames as extern from open62541 but it failed when using with shared libray
-const char *const LogLevelNames[6] = {
-  "trace",
-  "debug",
-  "info",
-  "warning",
-  "error",
-  "fatal" };
-const char *const LogCategoryNames[7] = {
+const std::unordered_map<size_t, const char*> LogLevelNames {
+  {100, "trace"},
+  {200, "debug"},
+  {300, "info"},
+  {400, "warning"},
+  {500, "error"},
+  {600, "fatal"},
+};
+
+const char *const LogCategoryNames[10] = {
   "network",
   "channel",
   "session",
   "server",
   "client",
-  "userland" };
+  "userland",
+  "securitypolicy",
+  "eventloop",
+  "pubsub",
+  "discovery" 
+};
 
-const UA_Logger COPC_UA_HandlerAbstract::UA_Forte_logger = {UA_Log_Forte, nullptr, UA_Log_Forte_clear};
+UA_Logger COPC_UA_HandlerAbstract::UA_Forte_logger = {UA_Log_Forte, nullptr, UA_Log_Forte_clear};
 
 COPC_UA_HandlerAbstract::COPC_UA_HandlerAbstract(CDeviceExecution& paDeviceExecution) :
     CExternalEventHandler(paDeviceExecution) {
@@ -52,7 +60,7 @@ int COPC_UA_HandlerAbstract::getPriority() const {
   return 0;
 }
 
-UA_Logger COPC_UA_HandlerAbstract::getLogger() {
+UA_Logger &COPC_UA_HandlerAbstract::getLogger() {
   return UA_Forte_logger;
 }
 
@@ -61,7 +69,7 @@ void COPC_UA_HandlerAbstract::UA_Log_Forte( //We omit SONAR only for the paramet
     ) {
 
   char tmpStr[mMaxLogLength];
-  forte_snprintf(tmpStr, mMaxLogLength, "[OPC UA LOGGER] %s/%s\t", LogLevelNames[paLevel], LogCategoryNames[paCategory]);
+  forte_snprintf(tmpStr, mMaxLogLength, "[OPC UA LOGGER] %s/%s\t", LogLevelNames.at(paLevel), LogCategoryNames[paCategory]);
   char *start = &tmpStr[strlen(tmpStr)];
 
   forte_vsnprintf(start, mMaxLogLength, paMsg, paArgs);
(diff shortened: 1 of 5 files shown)


src/com/opc_ua/opcua_local_handler.cpp
```diff
@@ -155,22 +158,33 @@ void COPC_UA_Local_Handler::generateServerStrings(TForteUInt16 paUAServerPort, U
...
+  paUaServerConfig.serverUrls = nullptr;
+  paUaServerConfig.serverUrlsSize = 0;
+
+  UA_String serverUrls[1];
+  size_t serverUrlsSize = 0;
+  char serverUrlBuffer[512];
 #ifdef FORTE_COM_OPC_UA_CUSTOM_HOSTNAME
-  UA_String customHost = UA_String_fromChars(paServerStrings.mHostname.c_str());
-  UA_String_copy(&customHost, &paUaServerConfig.customHostname);
+  snprintf(serverUrlBuffer, sizeof(serverUrlBuffer), "opc.tcp://%s:%u", FORTE_COM_OPC_UA_CUSTOM_HOSTNAME, paServerPort);
```
ernstblechaPT: This changes the behavior of the OPC-UA server regarding how the hostname is determined in case a custom hostname is set in the CMake configuration; before a string consisting of _forte and the forte-communication-port was added (https://github.com/eclipse-4diac/4diac-forte/pull/260/files#diff-edd355cac746cc077b5f13abb9b7c46b6c05082bdd8aba0717d093a2ac9fca34R150-R157)

As this behavior was seen as confusing before I am not against this change; I only want to make sure that this does not break things in some application

src/com/opc_ua/opcua_handler_abstract.cpp
```diff
@@ -14,24 +14,32 @@
 #include <forte_printer.h>
 #include <devlog.h>
 #include <parameterParser.h>
+#include <map>
```
cochicde:
```suggestion
#include <unordered_map>
```

src/com/opc_ua/opcua_local_handler.cpp
```diff
@@ -79,11 +79,13 @@ void COPC_UA_Local_Handler::run() {
   mUaServer = UA_Server_new();
   if(mUaServer) {
     UA_ServerConfig *uaServerConfig = UA_Server_getConfig(mUaServer);
-    UA_ServerConfig_setMinimal(uaServerConfig, gOpcuaServerPort, nullptr);
+    /* The original logger is needed to avoid memory leak on shutdown */
+    UA_Logger uaLogger = *uaServerConfig->logging;
```
cochicde, line 86: since you're basically overriding the logging, you should probably delete yourself the original one before assigning yours. You could do that in the `configureUAServer` method

src/com/opc_ua/opcua_local_handler.cpp
ernstblechaPT: @cochicde do you have any input on this topic as the earliest git commits on hostname generation trace this code to you?

cochicde: I can't remember why it was set like `forte_{PORT}` and I would assume it wasn't planned, it was just some decision based on nothing in particular. 

In this case, however, I would keep it consistent and:
1. Keep the string generation inside the `generateServerStrings`
2. Don't change things related to the hostname generation in this PR, but focus only on the update to the new stack version. I'd move the hostname change to a new PR for simplicity

m-meingast: Yes, that is true. Instead of something like `customName-forte_4840` , now only `customName` is used. I can change that back to the previous version. However, the member variable `customHostname` does not exist anymore in the 1.4 versions. Instead, there is a new variable `UA_String* serverUrls` because they changed their networking infrastructure quite a lot.

Nevertheless, the custom hostname generation will need some additional refactoring because I just noticed that there is a bug when using a custom hostname in combination with a 1499 Struct. This bug already existed before the changes of this PR.

src/com/opc_ua/opcua_local_handler.cpp
m-meingast: Yes, that was my plan in the first place. However, I encountered the following issue while doing that: In 1.4 the UA logger instance is not a global constant anymore, but it is dynamically allocated in the UA_Server_new() function. It is now referenced by a series of members of the server, e.g. config.secureChannelPKI, config.sessionPKI, etc.. I tried to delete the original logger and just point all these logger references to the Forte logger, but there is one pointer to the UA logger that I can't access from the server configuration. Freeing the memory of the UA logger then leads to a segmentation fault at some point. That is why I decided to just override the UA logger, keep a pointer to it, and then reassign the original logger at server shutdown so that the logger is properly deleted and now memory leak occurs, when terminating Forte.

cochicde: I see. then you should probably add this information as a comment, since otherwise the information gets lost and someone might try to fix it and face the issue in the future. 

is this the same issue https://github.com/open62541/open62541/issues/6769?

m-meingast: I already had written a comment, but I added some more lines to give more context. Thanks!

Also, yes, this is the same issue. Applying the proposed workaround does not fix the memory leak though.

cochicde:
another question. What exact version is this aiming to? 1.4.6?

m-meingast:
> another question. What exact version is this aiming to? 1.4.6?

@cochicde I tested it in 1.4.5, but the main goal was to adapt FORTE so that it is usable with the 1.4 release. There were some breaking changes in the current implementation with the 1.3 release

cochicde:
> > another question. What exact version is this aiming to? 1.4.6?
> 
> @cochicde I tested it in 1.4.5, but the main goal was to adapt FORTE so that it is usable with the 1.4 release. There were some breaking changes in the current implementation with the 1.3 release

I see. We should probably have one fixed version when releasing the later forte version

azoitl:
> > > another question. What exact version is this aiming to? 1.4.6?
> > 
> > 
> > @cochicde I tested it in 1.4.5, but the main goal was to adapt FORTE so that it is usable with the 1.4 release. There were some breaking changes in the current implementation with the 1.3 release
> 
> I see. We should probably have one fixed version when releasing the later forte version

For the release version of 4diac FORTE I agree. For now we are experimenting with Alarms and Events and as this is still under work in open62541 I think we should try to keep up to date as quick as possible.

src/com/opc_ua/opcua_local_handler.cpp
cochicde: I'd still keep the all the string handling of the names under the `generateServerStrings` otherwise you two different part of the code doing the same thing, but differently

m-meingast: We should probably keep the old format `customName-forte_{PORT}` then, and fix the bug in a separate PR

m-meingast: I moved the generation of the hostname to the `generateServerStrings` method

cochicde: LGTM

cochicde (approved):

azoitl (approved):

PR #265, opened 2024-10-17
Clean up timer creation

cochicde:
1. Move all the creation part of the timer into a timerFactory. 
2. Each architecture adds its timer and other can be added easily
3. CMake will take care of populating the cpp file with the factory according to what was added

My main objectives were to have the posibility to dynamically change the timer, for example for testing. Also, removing duplicated code is always nice, and a little bit more of cleanup of the architectures will help keep it more stable.


src/CMakeLists.txt
```diff
@@ -118,6 +118,40 @@ CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/core/deviceExecutionHandlers.cpp.in $
...
+  set(TIMERHANDLER_DEFAULT "\"\"")
+endif()
+
+SET(TIMERHANDLER_INCLUDE "")
+SET(TIMERHANDLER_CREATE "")
+
+LIST(LENGTH TIMERHANDLER_CLASS TIMERHANDLER_CLASS_LEN)
+math(EXPR TIMERHANDLER_CLASS_LEN ${TIMERHANDLER_CLASS_LEN}-1)
+
+FOREACH(POS RANGE ${TIMERHANDLER_CLASS_LEN})
```
azoitl, line 140: as there can only be one timerhandler, why not just generating the creation function for this one. This avoids the need for the string param of the factory and would make it clearer what is going on. Also the code would only have the timerhandler needed.

src/arch/be_m1/CMakeLists.txt
```diff
@@ -19,14 +19,11 @@ if("${FORTE_ARCHITECTURE}" STREQUAL "Bachmann M1")
 
 
   forte_add_include_directories(${CMAKE_CURRENT_SOURCE_DIR})
-  if(FORTE_FAKE_TIME)
-    forte_set_timer(../fake_time/faketimerha)
-  else(FORTE_FAKE_TIME)
-    forte_set_timer(../posix/pctimeha)
-  endif(FORTE_FAKE_TIME)
+  
+  forte_add_timerhandler(CPCTimerHandler pctimeha )
```
azoitl: how is the factory known that the `pctimeha` file is in the `../posix/` directory?

azoitl (changes requested):

src/CMakeLists.txt
cochicde: During some trace testing, I need to be able to change the timer at runtime. I run the tests with the regular timer and then switch to the `fake_timer` to simulate different things.

src/arch/be_m1/CMakeLists.txt
cochicde: mmm it seems to be a bad copy + paste

src/CMakeLists.txt
ernstblechaPT: How would I change the timer handler during runtime?

Just by calling `CTimerHandlerFactory::setTimeHandlerNameToCreate` and `CTimerHandlerFactory::createTimerHandler`?

For me this would be used to switch beween FakeTime and the normal Timer with a commandline-switch

cochicde: > How would I change the timer handler during runtime?
> 
> Just by calling `CTimerHandlerFactory::setTimeHandlerNameToCreate` and `CTimerHandlerFactory::createTimerHandler`?

Yes, that's the way of doing it. My only use case for now is to change it in the unit test. 
1. I created a forte device with the default timer
2. execute some FBs 
3. get some result. 

Then  change the timer and execute the same three steps as above and compare the results.

Adding a command line could be an option for later if needed. I was also thinking if this might be needed in the future so it can be changed from 4diac-IDE with some custom command.

src/arch/fake_time/CMakeLists.txt
```diff
@@ -14,8 +14,13 @@ if(FORTE_FAKE_TIME)
 
   forte_add_definition("-DFORTE_FAKE_TIME")
```
ernstblechaPT, line 15: This definition is used to conditionally compile exchanged `getNanoSecondsMonotonic()` and `getNanoSecondsRealtime()` functions via the architecture dependent `forte_architecture_time.h` whenever FAKE_TIME is enabled during compilation;

I am not sure that/how this is handled by this change?

cochicde: I see. I completely missed this. Let me check...

cochicde: I managed to find a solution. 

There are three functions that were "replaced" by the fake timer, namely: forte_timer, getNanoSecondsMonotonic and getNanoSecondsRealtime. Now, we have an `Arch` version of them in each architecture (forte_timer_arch, getNanoSecondsMonotonicArch and getNanoSecondsRealtime), which is the same as the original. These functions are always compiled. 

The fake timer offers now three parallel functions: forte_timer_fake, getNanoSecondsMonotonicFake and getNanoSecondsRealtimeFake, which are compiled, as before, only if the FAKE_TIMER is enabled. These fake timer functions have now a conditional which checks which one is the currently timer being used (fake timer or the "regular" one form the archictures) and depending on that it calls the `Arch` version of these functions or the original fake implementation.

The `forte_architecture_time.h` offers alias to the original functions (without `Arch` nor `fake`) which are set to the `Arch` or the `Fake` version of them depending if the FAKE_TIMER is enabled. 

Performance-wise, if the FAKE_TIMER is disabled, nothing changes. If it's enabled, there's an extra `if`.

ernstblechaPT: Yes, that looks good - I'm still not fully happy with the string comparison in that extra `if` - I see that this makes the code a lot more generic but it feels like a rather high price to pay as I don't see how this could be removed by optimization

as of now there are only two possible timers (`architecture` and "faketimer`) for a compiled binary - maybe an enum would be the extensible choice? or is this still overthinking the problem at hand?

cochicde: I like the `enum` idea. This will make the comparison faster. Regardgin the extra `if`, it comes only if you compile with `fake_time` which you usually do while testing, you don't deploy on real scenarios. Also, before this PR, if you compiled `fake_time` you didn't have any other timer, so now you have both at the price of an extra `if` when testing basically.

ernstblechaPT (approved):

PR #267, opened 2024-10-29
Abstract tracer and create an internal one for test purposes

cochicde:
This PR is part a branch with some other things, so some things are still not used here. 
The folllowing was done here:
- Abstract tracer: any mention of the `barectf` tracer was removed from the main code and hidden in `BarectfPlatformFORTE`. To do that, a `CForteTracer` typedef was created which can be set to other tracers during compilation time. `BarectfPlatformFORTE` (and others tracers) have methods such as `traceInstanceData`, `traceReceiveInputEvent`, etc. which mirror the ones from `barectf`. The Function Blocks then call these methods in  `CForteTracer` class. 
- Internal tracer: Is a tracer that instead of tracing into a file, it traces into a vector of `EventMessage` which contains the same information as the `barectf` traces. This is only compile if the tracer tests are enabled.
- Flexible Tracer: A shell over both `BarectfPlatformFORTE` and the internal tracer, which allows selecting one or the other during runtime. This is the default tracer if tracer tests are enabled.
- Babeltrace2 Message Factory: everything related to `babeltrace2` was moved out of the `EventMessages` and this factory allows creating `EventMessages` from the `bt_field` (and other) types defined by `babeltrace2`. 

The main idea of this new code is to be able generate traces using `barectf` but also directly into memory and be able to have the same class to compare them (`EventMessage`)


src/core/trace/flexibleTracer.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "stringdict.h"
+#include "barectf_platform_forte.h"
+#include "internalTracer.h"
+
+/**
+ * @brief A tracer that can be changed to use the existing tracers: BarectfPlatformFORTE and CInternalTracer
+ * 
+ */
+class CFlexibelTracer final {
```
mx990:
```suggestion
class CFlexibleTracer final {
```

diplfranzhoepfinger:
we have a Tracer ? 
@azoitl 
this would be VERY helpful. 
such a Thing which X/Y Coordinates and writing Diagramms of Variables ? 
I am Impressed !

azoitl:
> we have a Tracer ? @azoitl this would be VERY helpful. such a Thing which X/Y Coordinates and writing Diagramms of Variables ? I am Impressed !

I don't think this is what you can to with that. Currently tracing is more for getting deeper insights on what is going on. But it heavily impacts performance and is not suitable for non-linux platforms.

diplfranzhoepfinger:
ahhhhhhhhhhhhhhhh .... 

so this would be a totally different thing: 
having a Chart Draw module in IDE, to make Plots during a Watch ?

mx990 (changes requested):

azoitl:
> ahhhhhhhhhhhhhhhh ....
> 
> so this would be a totally different thing: having a Chart Draw module in IDE, to make Plots during a Watch ?

yes

src/core/trace/flexibleTracer.h
cochicde: are you sure? 
https://www.merriam-webster.com/dictionary/flexibel?utm_campaign=sd&utm_medium=serp&utm_source=jsonld

mx990: Yes, when I click the link you posted, I get:
> The word you've entered isn't in the dictionary.

It even suggests the correct spelling: https://www.merriam-webster.com/dictionary/flexible

cochicde: Apparently I was very tired

mx990 (approved):

PR #269, opened 2024-11-03
Build errors on develop branch when enabling LUA (#583610)

gpiasenza:
Build errors on develop branch when enabling LUA (#583610)
    
Fixed typo in src/core/lua/CMakeLists.txt.
Completed changes about mInterfaceSpec into a reference in src/core/lua folder.
    
Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=583610.
With these changes forte builds and runs with LUA enabled, but still does not work.

### DIFF
--- a/src/core/lua/luaadapter.cpp
+++ b/src/core/lua/luaadapter.cpp
@@ -24,7 +24,6 @@ CLuaAdapter::~CLuaAdapter() = default;
 
 bool CLuaAdapter::initialize() {
   //before calling super we need to configure the interface of the FB
-  setupFBInterface(getFBInterfaceSpec());
   return CAdapter::initialize();
 }
 
--- a/src/core/lua/luaadapter.cpp
+++ b/src/core/lua/luaadapter.cpp
@@ -22,12 +22,6 @@ CLuaAdapter::CLuaAdapter(CStringDictionary::TStringId paInstanceNameId, const CL
 
 CLuaAdapter::~CLuaAdapter() = default;
 
-bool CLuaAdapter::initialize() {
-  //before calling super we need to configure the interface of the FB
-  return CAdapter::initialize();
-}
-
-
 void CLuaAdapter::readInputData(TEventID paEIID) {
   if(nullptr != getFBInterfaceSpec().mEIWithIndexes && scmNoDataAssociated != getFBInterfaceSpec().mEIWithIndexes[paEIID]) {
     const TDataIOID *eiWithStart = &(getFBInterfaceSpec().mEIWith[getFBInterfaceSpec().mEIWithIndexes[paEIID]]);
--- a/src/core/lua/luabfb.cpp
+++ b/src/core/lua/luabfb.cpp
@@ -57,7 +57,7 @@ const char CLuaBFB::LUA_NAME[] = "FORTE_CLuaFB";
 const luaL_Reg CLuaBFB::LUA_FUNCS[] = { { "__index", CLuaFB_index }, { "__newindex", CLuaFB_newindex }, { "__call", CLuaFB_call }, { nullptr, nullptr } };
 
 CLuaBFB::CLuaBFB(CStringDictionary::TStringId paInstanceNameId, const CLuaBFBTypeEntry* paTypeEntry, forte::core::CFBContainer &paContainer) :
-    CGenFunctionBlock<CBasicFB>(paContainer, paTypeEntry->getInterfaceSpec(), paInstanceNameId, paTypeEntry->getInternalVarsInformation()),
+    CGenFunctionBlock<CBasicFB>(paContainer, paInstanceNameId, paTypeEntry->getInternalVarsInformation()),
         mTypeEntry(paTypeEntry) {
   CLuaEngine *luaEngine = getResource()->getLuaEngine();
   luaEngine->registerType<CLuaBFB>();
@@ -67,7 +67,6 @@ CLuaBFB::CLuaBFB(CStringDictionary::TStringId paInstanceNameId, const CLuaBFBTyp
 
 bool CLuaBFB::initialize() {
   //before calling super we need to configure the interface of the FB
-  setupFBInterface(getFBInterfaceSpec());
   createVarInternals();
   return CGenFunctionBlock<CBasicFB>::initialize();
 }
--- a/src/core/lua/luabfb.cpp
+++ b/src/core/lua/luabfb.cpp
@@ -57,7 +57,7 @@ const char CLuaBFB::LUA_NAME[] = "FORTE_CLuaFB";
 const luaL_Reg CLuaBFB::LUA_FUNCS[] = { { "__index", CLuaFB_index }, { "__newindex", CLuaFB_newindex }, { "__call", CLuaFB_call }, { nullptr, nullptr } };
 
 CLuaBFB::CLuaBFB(CStringDictionary::TStringId paInstanceNameId, const CLuaBFBTypeEntry* paTypeEntry, forte::core::CFBContainer &paContainer) :
-    CGenFunctionBlock<CBasicFB>(paContainer, paInstanceNameId, paTypeEntry->getInternalVarsInformation()),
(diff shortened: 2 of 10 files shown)


src/core/lua/luaadapter.cpp
```diff
@@ -24,7 +24,6 @@ CLuaAdapter::~CLuaAdapter() = default;
 
 bool CLuaAdapter::initialize() {
   //before calling super we need to configure the interface of the FB
-  setupFBInterface(getFBInterfaceSpec());
   return CAdapter::initialize();
```
azoitl: as we are now only calling super we could remove the whole method and with that make the code cleaner.

src/core/lua/luabfb.h
```diff
@@ -62,7 +62,7 @@ class CLuaBFB : public CGenFunctionBlock<CBasicFB> {
     bool initialize() override;
 
     bool createInterfaceSpec(const char *paConfigString, SFBInterfaceSpec &paInterfaceSpec) override {
-      paInterfaceSpec = *mInterfaceSpec;
+      paInterfaceSpec = mInterfaceSpec;
```
azoitl: this should be:
`paInterfaceSpec = *paTypeEntry->getInterfaceSpec()`

src/core/lua/luacfb.h
```diff
@@ -33,7 +33,7 @@ class CLuaCFB : public CGenFunctionBlock<CCompositeFB> {
     bool initialize() override;
 
     bool createInterfaceSpec(const char *, SFBInterfaceSpec &paInterfaceSpec) override {
-      paInterfaceSpec = *mInterfaceSpec;
+      paInterfaceSpec = mInterfaceSpec;
```
azoitl: this should be:
`paInterfaceSpec = *paTypeEntry->getInterfaceSpec()`

azoitl (changes requested):

src/core/lua/luabfb.cpp
```diff
@@ -57,7 +57,7 @@ const char CLuaBFB::LUA_NAME[] = "FORTE_CLuaFB";
 const luaL_Reg CLuaBFB::LUA_FUNCS[] = { { "__index", CLuaFB_index }, { "__newindex", CLuaFB_newindex }, { "__call", CLuaFB_call }, { nullptr, nullptr } };
 
 CLuaBFB::CLuaBFB(CStringDictionary::TStringId paInstanceNameId, const CLuaBFBTypeEntry* paTypeEntry, forte::core::CFBContainer &paContainer) :
-    CGenFunctionBlock<CBasicFB>(paContainer, paTypeEntry->getInterfaceSpec(), paInstanceNameId, paTypeEntry->getInternalVarsInformation()),
+    CGenFunctionBlock<CBasicFB>(paContainer, paInstanceNameId, paTypeEntry->getInternalVarsInformation()),
```
mx990: I would suggest to change the return type of `CLuaBFBTypeEntry::getInterfaceSpec()` to a reference, keep passing it to the super constructor here and remove the assignment in `CLuaBFB::createInterfaceSpec()` instead.

src/core/lua/luacfb.cpp
```diff
@@ -19,15 +19,14 @@
 #include "criticalregion.h"
 
 CLuaCFB::CLuaCFB(CStringDictionary::TStringId paInstanceNameId, const CLuaCFBTypeEntry* paTypeEntry, SCFB_FBNData &paFbnData, forte::core::CFBContainer &paContainer) :
-    CGenFunctionBlock<CCompositeFB>(paContainer, paTypeEntry->getInterfaceSpec(), paInstanceNameId, paFbnData), mTypeEntry(paTypeEntry) {
+    CGenFunctionBlock<CCompositeFB>(paContainer, paInstanceNameId, paFbnData), mTypeEntry(paTypeEntry) {
```
mx990: I would suggest to change the return type of `CLuaCFBTypeEntry::getInterfaceSpec()` to a reference, keep passing it to the super constructor here and remove the assignment in `CLuaCFB::createInterfaceSpec()` instead.

src/core/lua/luacfb.cpp
```diff
@@ -19,15 +19,14 @@
...
 }
 
 CLuaCFB::~CLuaCFB() = default;
 
 
 bool CLuaCFB::initialize() {
   //before calling super we need to configure the interface of the FB
-  setupFBInterface(getFBInterfaceSpec());
   return CGenFunctionBlock<CCompositeFB>::initialize();
 }
```
mx990: I would suggest to remove this method, as well.

src/core/lua/luabfb.cpp
```diff
@@ -67,7 +67,6 @@ CLuaBFB::CLuaBFB(CStringDictionary::TStringId paInstanceNameId, const CLuaBFBTyp
 
 bool CLuaBFB::initialize() {
   //before calling super we need to configure the interface of the FB
```
mx990, line 69: Please also remove the obsolete comment.

mx990 (changes requested):

src/core/lua/luaadapter.cpp
gpiasenza: done

src/core/lua/luabfb.h
gpiasenza: uhm, there's no paInterfaceSpec parameter in createInterfaceSpec, maybe you refer to the constructor ? (I did that in the latest commit)

src/core/lua/luacfb.h
gpiasenza: same as above

src/core/lua/luabfb.cpp
gpiasenza: done. but... what is the purpose of createInterfaceSpec ? It does not seem to get called anywhere

src/core/lua/luacfb.cpp
gpiasenza: done

src/core/lua/luacfb.cpp
gpiasenza: done

src/core/lua/luabfb.cpp
gpiasenza: removed the entire function, as suggested above

src/core/lua/luabfb.cpp
mx990: The purpose of `createInterfaceSpec(...)` is for blocks where the interface is constructed dynamically based on the `paConfigString` parameter. In case of LUA blocks, the interface is already created in the respective type entry, so we can pass it directly to the super constructor. In that case, `createInterfaceSpec(...)` should simply return true and do nothing otherwise.

mx990 (changes requested):

src/core/lua/luabfb.h
azoitl: it would have been mTypeEntry. but with the changes proposed by @mx990 this is not needed anymore.

A simple return true is fine for all these methods.

gpiasenza: Ok, done

src/core/lua/luacfb.h
gpiasenza: Ok, done

src/core/lua/luabfb.cpp
gpiasenza: Ok, done

src/core/lua/luabfb.cpp
gpiasenza: Done

mx990 (approved):
Looks good to me. Thanks!

azoitl (approved):

azoitl:
@gpiasenza thx for the fix.

diplfranzhoepfinger:
LUA working now, or just compiling ?

azoitl:
> LUA working now, or just compiling ?

currently compiling. But from my recent tests at least Composite FBs and partly basics are working.

PR #270, opened 2024-11-03
Add a command line option to enable fakeTimer

ernstblechaPT:
As discussed in #265 this adds a command line option to enable the fakeTimer if the feature is compiled in.

### DIFF
--- a/src/arch/utils/mainparam_utils.cpp
+++ b/src/arch/utils/mainparam_utils.cpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2018 fortiss GmbH
+ *               2024 Primetals Technologies Austria GmbH
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -9,15 +10,22 @@
  * Contributors:
  *   Tarik Terzimehic
  *    - initial API and implementation and/or initial documentation
+ *   Ernst Blecha
+ *    - add command line option to enable fakeTimer
  *******************************************************************************/
 
 #include <forte_config.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#include "../arch/devlog.h"
 
 #ifdef FORTE_TRACE_CTF
 #include <string>
+#endif
+
+#ifdef FORTE_FAKE_TIME
+#include "arch/timerHandlerFactory.h"
 #endif
 
 /*!\brief Lists the help for FORTE
@@ -43,6 +51,9 @@ void listHelp(){
 #endif //FORTE_COM_HTTP
 #ifdef FORTE_TRACE_CTF
   printf("%-20s Set the output directory for TRACE_CTF\n", "  -t <directory>");
+#endif //FORTE_TRACE_CTF
+#ifdef FORTE_FAKE_TIME
+  printf("%-20s Enable simulated realtime using the fakeTimer\n", "  -F");
 #endif //FORTE_TRACE_CTF
 }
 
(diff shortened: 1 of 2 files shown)


ernstblechaPT:
I had another thought: since there is the `FakeTimeDev` and the `RMT_DEV` it is possible to select the required timer based on the device used.

src/arch/utils/mainparam_utils.cpp
```diff
@@ -9,17 +10,24 @@
...
  *    - initial API and implementation and/or initial documentation
+ *   Ernst Blecha
+ *    - add command line option to enable fakeTimer
  *******************************************************************************/
 
 #include <forte_config.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#include "../core/forteinstance.h"
```
cochicde: do you actually need this?

ernstblechaPT: Thank you - this should have been devlog.h

ernstblechaPT:
My Idea of tying automatic enablement of the FakeTimer to the FakeTimeDev poses a problem: the command line switch would need to be turned to a "disable FakeTimer" switch which breaks my use case.

In this state this practically makes the FakeTimer unusable on systems where FORTE is not started via command line. Is there currently someone using FakeTime on any platform other than Win32/Posix that we are aware of?

cochicde:
> FakeTimeDev

I'm not sure how the FakeTimeDev is being used. I understand that you need to compile forte with `FORTE_DEVICE=FakeTimeDev` and you get the fakeTimer in it. 

What's your use case? Do you want to also have the `disable FakeTimer` command line to be able to use a regular timer? In that case I think the FakeTimeDev becomes the same to a RMT_DEV, right?

ernstblechaPT:
> > FakeTimeDev
> 
> I'm not sure how the FakeTimeDev is being used. I understand that you need to compile forte with `FORTE_DEVICE=FakeTimeDev` and you get the fakeTimer in it.
> 
> What's your use case? Do you want to also have the `disable FakeTimer` command line to be able to use a regular timer? In that case I think the FakeTimeDev becomes the same to a RMT_DEV, right?

I want to build one FORTE binary and have it start up as a "normal" FORTE as default - and if needed switch it to FakeTime for doing some testing

I have not problem adding a switch for testing, but the default should be the "normal" FORTE

cochicde:
> I want to build one FORTE binary and have it start up as a "normal" FORTE as default - and if needed switch it to FakeTime for doing some testing
> 
> I have not problem adding a switch for testing, but the default should be the "normal" FORTE

But do you actually need the `FakeTimeDevice`? I see it has the extra input for setting the time it seems. is this something that's actually being used? Do you want to be able to switch between `RMT_DEV` and `FakeTimeDevice` or actually between `RMT_DEV(with normal timer)` and `RMT_DEV(with fakeTimer)`?

ernstblechaPT:
> > I want to build one FORTE binary and have it start up as a "normal" FORTE as default - and if needed switch it to FakeTime for doing some testing
> > I have not problem adding a switch for testing, but the default should be the "normal" FORTE
> 
> But do you actually need the `FakeTimeDevice`? I see it has the extra input for setting the time it seems. is this something that's actually being used? Do you want to be able to switch between `RMT_DEV` and `FakeTimeDevice` or actually between `RMT_DEV(with normal timer)` and `RMT_DEV(with fakeTimer)`?

Having the FakeTimeDevice allows you to set the time during a Debug-Session for a FB via the FB-Debug view in the Debug perspective - this is a really powerful tool

cochicde:
> Having the FakeTimeDevice allows you to set the time during a Debug-Session for a FB via the FB-Debug view in the Debug perspective - this is a really powerful tool

I see two possible ways of having that:
1. Allow multiple devices to be compiled: create some factory for the type of device being created. The selection could be then done via a command  line flag being the regular RMT_DEV the default one. This will allow also to have the `OPCUA_DEV` device. The FakeTimeDev would need to set the timer properly in the constructor and set it back to the regular in the destructor.
2. Extend the interface of the `FakeTimeDev` to allow switching timers: A boolean input could allow to select between the fake and the normal timer. This will basically convert your `FakeTimeDev` in to a regular `RMT_DEV` at runtime.

PR #273, opened 2024-11-13
Add event type to interface spec

MandKastner:
Add event type lists to function block interface specification for input and output events. This will also update the interface specification of exported and generic FBs.


src/core/funcbloc.h
```diff
@@ -84,10 +84,12 @@ struct SCFB_FBInstanceData {
 struct SFBInterfaceSpec {
     TEventID mNumEIs; //!< Number of event inputs
     const CStringDictionary::TStringId *mEINames; //!< List of the event input names
+    const CStringDictionary::TStringId *mEITypeNames; //!< List of the event input types
```
mx990, line 87: I suggest adding the additional fields at the end, so that existing code remains compatible.

src/core/funcbloc.h
```diff
@@ -164,6 +166,22 @@ class CFunctionBlock : public forte::core::CFBContainer {
       return static_cast<TEventID>(getPortId(paEONameId, getFBInterfaceSpec().mNumEOs, getFBInterfaceSpec().mEONames));
     }
 
+    /*! \brief Gets the EventTypeID of a specific input event
+     * \param paEIID  StringId of the event name.
+     * \return Returns the EventTypeID of a specific input event
+     */
+    CStringDictionary::TStringId getEIType(TEventID paEIID) const {
+      return getFBInterfaceSpec().mEITypeNames[paEIID];
```
mx990: I suggest returning `g_nStringIdEvent` as default when `mEITypeNames` is null. That way, most blocks do not need to have any event type list.

mx990 (changes requested):

src/core/funcbloc.h
azoitl: I don't think we should do that. Because the getEvent Type Methods will then undeteccted read from arbitrary memory.  We are working on a 3.0.0 we can now break some API.

src/core/funcbloc.h
azoitl: Oh this is a nice idea. it could be:

```
if(getFBInterfaceSpec().mEITypeNames != nullptr) {
   return getFBInterfaceSpec().mEITypeNames[paEIID];
}
return g_nStringIdEvent;
```

And the exporter could optimize to not output event type ids when all event types are EVENT.

src/core/funcbloc.h
mx990: Why would they read from arbitrary memory? Any excess fields are implicitly initialized, see https://en.cppreference.com/w/cpp/language/aggregate_initialization#Implicitly_initialized_elements.

azoitl: Ah yes my brain took the wrong turn. Still I think having them in the right order is cleaner for the future.

diplfranzhoepfinger:
@azoitl the Big change you were talking about ?

azoitl:
> @azoitl the Big change you were talking about ?

yes

azoitl (approved):

mx990 (approved):

PR #201, opened 2024-07-31
Cleanup opcua pointers

cochicde:
- removed pointers variables when not needed
- uso `auto` for readibility
- initialize members variables already in the class definition
- remove unneeded storage of strings in object structure helper

PR #279, opened 2024-11-20
Fix error with String Node Ids in Object Struct Helper

m-meingast:
The changes of #201 lead to a use-after-scope issue when creating Struct Types with String Node Ids in OPC UA, so additional changes are needed. In more detail, the `memberBrowsePathStr` variable in the `addOPCUAStructTypeComponent` method used a reference on a local buffer. I allocated memory to avoid this issue.

### DIFF
--- a/src/com/opc_ua/opcua_objectstruct_helper.cpp
+++ b/src/com/opc_ua/opcua_objectstruct_helper.cpp
@@ -65,6 +65,7 @@ bool COPC_UA_ObjectStruct_Helper::checkStructTypeConnection(CActionInfo &paActio
 }
 
 bool COPC_UA_ObjectStruct_Helper::createOPCUAStructType(CActionInfo &paActionInfo, const std::string &paStructTypeName, CIEC_STRUCT &paStructType) {
+  bool retVal = true;
   COPC_UA_Local_Handler* localHandler = static_cast<COPC_UA_Local_Handler*>(mHandler);
   if(!localHandler) {
     DEVLOG_ERROR("[OPC UA OBJECT STRUCT HELPER]: Failed to get LocalHandler because LocalHandler is null!\n");
@@ -78,10 +79,12 @@ bool COPC_UA_ObjectStruct_Helper::createOPCUAStructType(CActionInfo &paActionInf
   for(size_t i = 0; i < paStructType.getStructSize(); i++) {
     CIEC_ANY* structMember = paStructType.getMember(i);
     if(!addOPCUAStructTypeComponent(server, typeNodeId, paStructTypeName, structMember, structMemberNames[i])) {
-      return false;
+      retVal = false;
+      break;
     } 
   }
-  return true;
+  UA_NodeId_clear(&typeNodeId);
+  return retVal;
 }
 
 bool COPC_UA_ObjectStruct_Helper::createOPCUANamespace(char* nsName) {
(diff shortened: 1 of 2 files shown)


src/com/opc_ua/opcua_objectstruct_helper.cpp
```diff
@@ -133,9 +136,10 @@ bool COPC_UA_ObjectStruct_Helper::addOPCUAStructTypeComponent(UA_Server *paServe
     vAttr.dataType = COPC_UA_Helper::getOPCUATypeFromAny(*paStructMember)->typeId;
 
   UA_NodeId memberNodeId;
+  char *memberBrowsePathStr = nullptr;
   if(paParentNodeId.identifierType == UA_NODEIDTYPE_STRING) {
-    std::string memberBrowsePathStr = getStructMemberBrowsePath(paStructName, paStructMemberNameId);
-    memberNodeId = UA_NODEID_STRING(mOpcuaTypeNamespaceIndex, &memberBrowsePathStr[0]);
```
ernstblechaPT, line 138: Is there a reason not to use

`memberNodeId = UA_NODEID_STRING_ALLOC(mOpcuaTypeNamespaceIndex, memberBrowsePathStr.c_str());`

As far as I understand it that should solve this problem without having to resort to returning raw pointers and manually handling all of the memory management (The UA_NodeId_clear calls are still necessary)

cochicde: Agree. `memberNodeId` is not properly handling its memory and should copy from `memberBrowsePathStr` which is doing fine. That will also keep `getStructMemberBrowsePath` untouched.

We might want to start thinking about some wrappers around `UA_NodeId` which handle the memory safely so we don't miss this kind of errors (something like this UA_String wrapper https://github.com/eclipse-4diac/4diac-forte/pull/239/files#diff-628010b336b740e9af10dde6015a0787f2140f60a1898ce16bbc59f872d6e539R65)

src/com/opc_ua/opcua_objectstruct_helper.cpp
```diff
@@ -144,6 +152,8 @@ bool COPC_UA_ObjectStruct_Helper::addOPCUAStructTypeComponent(UA_Server *paServe
     UA_QUALIFIEDNAME(mOpcuaTypeNamespaceIndex, &structMemberName[0]),
     UA_NODEID_NUMERIC(0, UA_NS0ID_BASEDATAVARIABLETYPE), vAttr, nullptr, &memberNodeId);
 
+  UA_String_clear(&memberBrowsePath);
+  UA_NodeId_clear(&memberNodeId);
```
ernstblechaPT, line 156: Is this the correct order for clearing these allocations? Without having looked into the documentation this looks like it's the wrong way around?

src/com/opc_ua/opcua_objectstruct_helper.cpp
m-meingast: That's true. I didn't think of that. However, when I'm using something like `memberNodeId = UA_NODEID_STRING_ALLOC(mOpcuaTypeNamespaceIndex, memberBrowsePathStr.c_str());`, then the address sanitizer is reporting a memory leak.  The String allocation seems to not be correctly freed.  That is also the case when I try to clear the UA_string additionaly. 
Allocating the string with the  `UA_String_fromChars` method, assigning it manually to the NodeId and then clearing it afterwards works.

m-meingast: @cochicde That is true. A wrapper might might help to avoid these issues. We should discuss that with @azoitl.

cochicde: `UA_NODEID_STRING_ALLOC` allocates a new string, and `UA_NodeId_clear` de-allocates it.

I think the problem you're facing is related to the way you're calling  `UA_Server_addVariableNode`. The last parameter is the output node Id, i.e. the node ID of the new node it was created. You're passing the parent node (`memberNodeId`) also there, meaning that's overwritten. If you are not using the result NodeId you should pass `nullptr`

Here's the snipped code that handles that when adding a new node (`memberNodeId` becomes `outNewId` when calling the function)

```
 if(outNewNodeId)
        *outNewNodeId = result.addedNodeId;
    else
        UA_NodeId_clear(&result.addedNodeId);
```

m-meingast: Thanks for pointing that out. In this case, the output node Id is not needed indeed. I replaced it with a nullptr. Now the STRING_ALLOC function works as intended.

src/com/opc_ua/opcua_objectstruct_helper.cpp
```diff
@@ -366,7 +370,7 @@ std::string COPC_UA_ObjectStruct_Helper::getStructMemberBrowsePathWithNSIndex(co
   return ss.str();
 }
 
-std::string COPC_UA_ObjectStruct_Helper::getStructMemberBrowsePath(const std::string &paBrowsePathPrefix, const CStringDictionary::TStringId structMemberNameId) {
+std:: string COPC_UA_ObjectStruct_Helper::getStructMemberBrowsePath(const std::string &paBrowsePathPrefix, const CStringDictionary::TStringId structMemberNameId) {
```
cochicde: here's apparently something wrong

m-meingast: Thanks. I removed the blank space

cochicde (approved):

azoitl (approved):

PR #284, opened 2024-11-27
Fix VisualStudio 2022 compilation errors.

melekhove:
VS 2022 compiler seems to be a bit more strict in some aspects.

### DIFF
--- a/src/core/lua/luaadaptertypeentry.cpp
+++ b/src/core/lua/luaadaptertypeentry.cpp
@@ -33,7 +33,7 @@ CLuaAdapterTypeEntry* CLuaAdapterTypeEntry::createLuaAdapterTypeEntry(CStringDic
     return nullptr;
   }
   //interfaceSpec
-  SFBInterfaceSpec interfaceSpec = { 0, nullptr, nullptr, nullptr, 0, nullptr, nullptr, nullptr, 0, nullptr, nullptr, 0, nullptr, nullptr, 0, nullptr, 0, nullptr};
+  SFBInterfaceSpec interfaceSpec{};
   if(!luaEngine.pushField(-1, "interfaceSpec", LUA_TTABLE)) {
     return nullptr;
   }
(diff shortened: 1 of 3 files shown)


azoitl (commented):
As far as I understood @mx990 we could remove the whole initializer here as C++ would anyhow automatically initialize all values with 0 or nullptrs. Am I right @mx990?

melekhove:
I'm afraid, it is not that clear. Initialization syntax & semantics differ across time, c++ standard and compilers.
In the following [document](https://en.cppreference.com/w/cpp/language/initialization) you can see a lot of clauses like (until/since C++11/C++20 ) :-)
Taking into account different compilers (gcc/clang/lc etc)I'd prefer to be on the safe side & initialize struct on the stack explicitly.

mx990:
> As far as I understood @mx990 we could remove the whole initializer here as C++ would anyhow automatically initialize all values with 0 or nullptrs. Am I right @mx990?

Yes, I would suggest the following:
```c++
SFBInterfaceSpec interfaceSpec{};
```

> I'm afraid, it is not that clear. Initialization syntax & semantics differ across time, c++ standard and compilers. In the following [document](https://en.cppreference.com/w/cpp/language/initialization) you can see a lot of clauses like (until/since C++11/C++20 ) :-) Taking into account different compilers (gcc/clang/lc etc)I'd prefer to be on the safe side & initialize struct on the stack explicitly.

The semantics regarding initialization as defined in the C++ standard have indeed been updated several times, mostly to account for new features added to the language. This makes the article regarding initialization somewhat hard to read.

I would still like to point you to https://en.cppreference.com/w/cpp/language/zero_initialization, particularly clause (2) and also the example code at the end. As you can see, little has changed regarding that topic. Please note that we are also only targeting C++ 17 and above.

As for the behavior of different compilers, I believe it is impossible to anticipate the quirks of all the compilers out there. Being extra verbose may also lead to problems sometimes, as clearly evidenced by this PR. Aside from that, if you have any concrete information of a compiler misbehaving here, please let us know, so we can take appropriate action.

melekhove:
Ok, done.

mx990 (approved):

azoitl (approved):

azoitl:
@melekhove thx for finding this problem and the updated fix.

PR #285, opened 2024-11-28
Organize commFB interface and cleanup delete[] calls

cochicde:
- Move interface related functions from commFB to baseCommFB
- Remove all delete[] calls related to the interface in the destructors from all generic FB


cochicde:
While checking the destructor of the commFB I stumbled upon this mismatch in the handling of interface in the communiation FB. Then began the rabitt hole

src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
```diff
@@ -73,8 +68,8 @@ bool CGenBitBase::createInterfaceSpec(const char *paConfigString, SFBInterfaceSp
   if (paInterfaceSpec.mNumDIs < CFunctionBlock::scmMaxInterfaceEvents) {
 
     //create the data inputs
-    mDataInputNames = new CStringDictionary::TStringId[paInterfaceSpec.mNumDIs];
-    mDataInputTypeIds = new CStringDictionary::TStringId[paInterfaceSpec.mNumDIs];
+    mDataInputNames.reset(new CStringDictionary::TStringId[paInterfaceSpec.mNumDIs]);
+    mDataInputTypeIds.reset(new CStringDictionary::TStringId[paInterfaceSpec.mNumDIs]);
```
mx990: I would suggest to use [make_unique](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) instead to avoid the explicit `new` and also make it clear you are not replacing an existing pointer value:
```suggestion
    mDataInputNames = std::make_unique<CStringDictionary::TStringId[]>(paInterfaceSpec.mNumDIs);
    mDataInputTypeIds = std::make_unique<CStringDictionary::TStringId[]>(paInterfaceSpec.mNumDIs);
```

src/stdfblib/events/GEN_E_DEMUX_fbt.cpp
```diff
@@ -64,17 +59,17 @@ bool GEN_E_DEMUX::createInterfaceSpec(const char *paConfigString, SFBInterfaceSp
...
 
       if(paInterfaceSpec.mNumEOs < CFunctionBlock::scmMaxInterfaceEvents){
-        mEventOutputNames = new CStringDictionary::TStringId[paInterfaceSpec.mNumEOs];
-        mEventOutputTypeIds = new CStringDictionary::TStringId[paInterfaceSpec.mNumEOs];
+        mEventOutputNames.reset(new CStringDictionary::TStringId[paInterfaceSpec.mNumEOs]);
+        mEventOutputTypeIds.reset(new CStringDictionary::TStringId[paInterfaceSpec.mNumEOs]);
         const CStringDictionary::TStringId eventID = CStringDictionary::getInstance().getId("Event");
-        std::memset(mEventOutputTypeIds, static_cast<int>(eventID), sizeof(*mEventOutputTypeIds) * paInterfaceSpec.mNumEOs);
-        generateGenericInterfacePointNameArray("EO", mEventOutputNames, paInterfaceSpec.mNumEOs);
+        std::fill_n(mEventOutputTypeIds.get(), paInterfaceSpec.mNumEOs, eventID);
```
mx990: I believe this array is entirely unnecessary due to https://github.com/eclipse-4diac/4diac-forte/pull/273#discussion_r1839765837.

mx990 (changes requested):
> * Move interface related functions from commFB to baseCommFB

Could you please elaborate a bit on your reasons for making this change?

> * Remove all delete[] calls related to the interface in the destructors from all generic FB

I really like this change, but I believe we should separate this into its own PR.

MartinMelikMerkumians:
Regarding pulling this from commFB to baseCommFB, I originally made this intentional split for a reason, to enable different kind of commFBs, which I still have a private project utilizing this and planning to merge it back to the open source project.

I would therefore prefer not to have this change if there are no compelling reasons to do so now, otherwise I most likely wont merge my side project back (FYI @azoitl )

azoitl:
> Regarding pulling this from commFB to baseCommFB, I originally made this intentional split for a reason, to enable different kind of commFBs, which I still have a private project utilizing this and planning to merge it back to the open source project.
> 
> I would therefore prefer not to have this change if there are no compelling reasons to do so now, otherwise I most likely wont merge my side project back (FYI @azoitl )

@MartinMelikMerkumians thx for pointing this out. I vaguely remembered there was a reason for the split and I already thought that this shouldn't be done. With your comment I strongly oppose this to move stuff to BaseCommFB.

cochicde:
I don't mind move it back to the child class, but the problem is that the Base class is deallocating memory which was allocated in the child class. Basically the baseClass is implying that the child class would allocate certain parts of the interface.

I would still remove the `delete[]` calls in the destructor of the baseCommFB and keep the memory allocated in the child CommFB contained in it. This might break something in the private project, but the effort to adapt should be only on using `unique_ptr` also there

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -93,9 +87,15 @@ bool GEN_STRUCT_DEMUX::createInterfaceSpec(const char *paConfigString, SFBInterf
...
       size_t structSize = structInstance->getStructSize();
       if(structSize != 0 && structSize < cgInvalidPortId) { //the structure size must be non zero and less than cgInvalidPortId (maximum number of data outputs)
-        CStringDictionary::TStringId *doDataTypeNames = new CStringDictionary::TStringId[GEN_STRUCT_MUX::calcStructTypeNameSize(*structInstance)];
-        CStringDictionary::TStringId *doNames = new CStringDictionary::TStringId[structSize];
-        CStringDictionary::TStringId *diDataTypeNames = new CStringDictionary::TStringId[1];
+       
+        // we need doDataTypeNames here because we need to call fillDataPointSpec later which requires a non-const reference of the pointer
+        // wich we cannot get from the unique_ptr
+        auto doDataTypeNames = new CStringDictionary::TStringId[GEN_STRUCT_MUX::calcStructTypeNameSize(*structInstance)];
+        mDoDataTypeNames.reset(doDataTypeNames);
```
mx990: I would suggest doing it the other way around by assigning a local variable from the `unique_ptr` and also moving it directly before the call to `fillDataPointSpec` below, since the pointer should not be used for anything else after `fillDataPointSpec` has modified it.

src/modules/convert/GEN_STRUCT_MUX_fbt.cpp
```diff
@@ -106,10 +100,15 @@ bool GEN_STRUCT_MUX::createInterfaceSpec(const char *paConfigString, SFBInterfac
                 CStringDictionary::getInstance().get(structTypeNameId), cgInvalidPortId);
     return false;
   }
+  
+  // we need diDataTypeNames here because we need to call fillDataPointSpec later which requires a non-const reference of the pointer
+  // wich we cannot get from the unique_ptr
+  auto diDataTypeNames = new CStringDictionary::TStringId[calcStructTypeNameSize(*structInstance)];
+  mDiDataTypeNames.reset(diDataTypeNames);
```
mx990: same here

mx990 (changes requested):

mx990 (approved):

PR #290, opened 2024-11-29
Fix stdflib/io build

melekhove:
Add source files to project configuration


azoitl:
I'm sorry but with the current setup not all files of this directory should be added per default to a FORTE build. The specific IO implementations are in charge to pull in the right files that they need. This may change in the future if we will only have one IO API, but for now I would like to keep it as is.

melekhove:
Ok, I see, thanks.

PR #296, opened 2024-12-18
Refactor mgm commands

cochicde:
This PR is driven by the need of parsing the bootfiles without the need of the DEV_MGR, for example in unit tests. The changes are:
- Moved parsing methods outside of the class, mostly by copy pasting. 
- THey are part of class that offers two public functions: `parseAndExecuteMGMCommand` (same as before) and `generateResponse` which dispatch to the previous `generateLongReponse`, `generateShortResponse`  and `generateMonitorResponse`
- Some functions changed the RESP() data from the function block. These functions now return a string and the function block is settings RESP() with this returned string.
- The device where to excute the command is also passed as a parameter (previously it was accessed direclty as it's member from DEV_MGR.
- The response text functions and variables were moved closed to the mgm messages.
- Boot File Loader: 
   - Receives a callback std::function instead of the particular class from which it was needed to derived.
   - The boot loader receives an optional path as parameter where to load from, to avoid workaround with the global or environment variables. If this path is nullopt, the previous logic applies.


src/core/mgmcmd.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <type_traits>
+
+namespace forte::mgm_cmd {
+
+  /*!\brief Type for the response of MGM command messages
+  *
+  * TODO fully define all responses as defined in IEC 61499 inc. numbers.
+  */
+  const char * const scmMGMResponseTexts[13] = { "RDY", "BAD_PARAMS", "LOCAL_TERMINATION", "SYSTEM_TERMINATION", "NOT_READY", "UNSUPPORTED_CMD", "UNSUPPORTED_TYPE", "NO_SUCH_OBJECT", "INVALID_OBJECT", "INVALID_OPERATION", "INVALID_STATE", "OVERFLOW", "INVALID_DST" };
```
azoitl: I think with all the changes in 4diac FORTE std:string would be better for this array.

src/stdfblib/ita/CommandParser.h
```diff
@@ -0,0 +1,136 @@
...
+#define _COMMAND_PARSER_H
+
+#include "core/mgmcmd.h"
+#include "core/mgmcmdstruct.h"
+
+#include <string>
+
+class CDevice;
+
+namespace forte::command_parser {
```
azoitl: I'm not sure if this is a suitable namespace, better would be ita or something related to compliance profile for feasibility demonstrations where the format is coming from.

src/stdfblib/ita/CommandParser.h
```diff
@@ -0,0 +1,136 @@
...
+
+  class Parser {
```
azoitl: Parser is a bit to generic For me a class name should reflect the purpose.

src/stdfblib/ita/CommandParser.h
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * @brief Parse and executes a commmand on a destination in a device
+     * 
+     * @param paDest destination where to executed the command
+     * @param paCommand the command to be executed
+     * @param paDevice device where to execute the command
+     * @return EMGMResponse response of the execution of the command
+     */
+    EMGMResponse parseAndExecuteMGMCommand(const char *const paDest, char *paCommand, CDevice& paDevice);
```
azoitl: I would put the device as field and initialize it in the constructor.

azoitl (changes requested):

cochicde:
done!

azoitl (commented):
Memory consumption/performance wise the new implementation has a big drawback, for each request it creates a parser object and the parser object from scratch the mgmcommand object. The mgm command object was designed that it has strings where the memory of the strings would be reused between different invocations. 

I see two options:
 1. move the mgmcomand out of the parser and keep it as member of the DEV_MGR class. With that the execution of the command would be done outside of the parser. Which feels even a bit better. Because the parser would then really have only a single responsibility: command parsing
 2. make the parser a member of DEV_MGR

I have a slight tendency for 1.

cochicde:
> 1. mgmcomand

you're right, I missed that the command was a member of DEV_MGR. however, I have a tendency towards solution 2 actually, since as a user of the CommandParser class, you just want your command to be parsed. If we go towards 1, the object handling the CommandParser would also need to handle the mgmcommand which will be used exclusively with the CommandParser, adding some complexity and removing some encapsulation.

azoitl (approved):

azoitl:
> > 1. mgmcomand
> 
> you're right, I missed that the command was a member of DEV_MGR. however, I have a tendency towards solution 2 actually, since as a user of the CommandParser class, you just want your command to be parsed. If we go towards 1, the object handling the CommandParser would also need to handle the mgmcommand which will be used exclusively with the CommandParser, adding some complexity and removing some encapsulation.

While you contradict yourself with what you are writing: you are not just parsing the command but also executing, I see the advantages of haveing it all in the class, especially it would allow more easier for future optimizations and clean-ups more isolated.

cochicde:
> > > 1. mgmcomand
> > 
> > 
> > you're right, I missed that the command was a member of DEV_MGR. however, I have a tendency towards solution 2 actually, since as a user of the CommandParser class, you just want your command to be parsed. If we go towards 1, the object handling the CommandParser would also need to handle the mgmcommand which will be used exclusively with the CommandParser, adding some complexity and removing some encapsulation.
> 
> While you contradict yourself with what you are writing: you are not just parsing the command but also executing, I see the advantages of haveing it all in the class, especially it would allow more easier for future optimizations and clean-ups more isolated.

I see, you're right. I was thinking more about managing the lifetime of the mgmcommand object.

PR #298, opened 2024-12-24
Implement startup behavior for EInit events

MandKastner:
Unconnected event pins with the event type EInit of FBs will be triggered automatically during startup


src/core/funcbloc.cpp
```diff
@@ -50,6 +50,7 @@ bool CFunctionBlock::initialize() {
 #ifdef  FORTE_SUPPORT_MONITORING
   setupEventMonitoringData();
 #endif //FORTE_SUPPORT_MONITORING
+  mInputEventConnectionCount = std::make_unique<size_t[]>(getFBInterfaceSpec().mNumEIs);
```
mx990: I believe you need to re-create the array when there interface is reconfigured, see callers of `setupEventMonitoringData()`.

src/core/funcbloc.h
```diff
@@ -506,6 +516,17 @@ class CFunctionBlock : public forte::core::CFBContainer {
       }
     }
 #endif //FORTE_TRACE_CTF
+    /* !\brief checks if an input event pin is connected
+     *
+     */
+    [[nodiscard]] bool isConnected(TEventID paEIID) const {
```
mx990: Please rename this to `isInputEventConnected`.

src/core/funcbloc.cpp
```diff
@@ -432,6 +433,7 @@ EMGMResponse CFunctionBlock::changeExecutionState(EMGMCommandType paCommand){
       if((E_FBStates::Idle == mFBState) || (E_FBStates::Stopped == mFBState)){
         mFBState = E_FBStates::Running;
         nRetVal = EMGMResponse::Ready;
+        triggerEInitEvents();
```
mx990: It seems this will also trigger `EInit` events after a reconfiguration or warm restart? Is this the desired behavior?

mx990 (changes requested):
I am somewhat troubled that we are adding a separate memory allocation to hold the input event connection counters for *all* FB instances. In case of the similar monitoring counters, users can at least disable monitoring on constrained devices or if it is simply not needed.

I would like to propose several alternative approaches for discussion:
1. we use a common counter for all `EInit` events [^1],
2. we use a mark-and-sweep search when the resource is started, marking all FBs with input connections on `EInit` events [^1] and then triggering all unmarked FBs (also resetting the mark),
3. we only allocate the counters when `SFBInterfaceSpec::mEITypeNames` is not `nullptr`, which would at least reduce the impact for most FBs [^2].

In any case, we may want to add a compile-time flag to disable automatic triggering of unconnected `EInit` events on startup, including any potential overhead incurred thereof.

[^1]: This would slightly change the semantics if we had multiple `EInit` events, but maybe we should disallow that anyway?
[^2]: In that case, we should probably also make sure that FB types in the source tree with only `EEvent` events have `mEITypeNames` set to `nullptr`.

MandKastner:
@mx990 Thank you for your suggestions. I would prefer to keep the infrastructure to check if a event port is connected and go with your 3rd suggestion to reduce allocated memory. As you also suggested i moved the trigger mechanism to a SIFB and perform the recursive search there, this will allow users to decide about the triggering process.

src/core/funcbloc.cpp
MandKastner: I am not sure if i understand what you mean. The only caller of `setupMonitoringData()` is `initialize()`. In my understanding the existing `unique_ptr` will give up its ownership and take the new one with the second call of `initialize()` . With the release the managed memory should be freed and the new array created.

mx990: It is also called in https://github.com/eclipse-4diac/4diac-forte/blob/1511891ea208ff88fd4e3db1bc093f12f0e5ab91/src/core/genfb.tpp#L221 after setting up the interface of a generic FB.

azoitl: But with @MandKastner current implementation using a std::unique_ptr this does not really matter. Or do I miss something?

azoitl: Ah I missed the point. Yes we need to put a call there as well.

franz-hoepfinger-4diac:
I will test this now. 
eagerly waiting for this.

franz-hoepfinger-4diac:
so, First Test: Failed: 
it seems none of the EInit Pins are triggered: 

image

franz-hoepfinger-4diac:
ahhhhhhhhhhh i see... there is a new FB: E_TRIG!

franz-hoepfinger-4diac:
Working now ... Cool. 

image


image

franz-hoepfinger-4diac:
on bigger PRojects i get a lot of 
`ERROR: T#25000000000: External event queue is full, external event dropped!`
Messages.

src/core/funcbloc.cpp
```diff
@@ -671,6 +674,18 @@ size_t CFunctionBlock::getToStringBufferSize() const {
 
 }
 
+void CFunctionBlock::triggerEventsOfType(TEventTypeID paEventTypeId) {
```
azoitl: please move to E_TRIG FB

src/core/funcbloc.cpp
```diff
@@ -671,6 +674,18 @@ size_t CFunctionBlock::getToStringBufferSize() const {
...
+  //most of the FBs will only have the basic event type -> mEITypes == nullptr
+  if (getFBInterfaceSpec().mEITypeNames != nullptr) {
+    for (TEventID eventId = 0; eventId < getFBInterfaceSpec().mNumEIs; eventId++) {
+      if (getEIType(eventId) == paEventTypeId && !isInputEventConnected(eventId)) {
+        getResource()->getResourceEventExecution()->startEventChain(CConnectionPoint(this, eventId));
```
azoitl: use the eventchain execution thread which was used to triggere the R_TRIG fb and add the events to the internal event list.

src/stdfblib/events/E_TRIG_fbt.cpp
```diff
@@ -0,0 +1,121 @@
...
+    case 0: return &conn_EVENTTYPE;
+  }
+  return nullptr;
+}
+
+CDataConnection *FORTE_E_TRIG::getDOConUnchecked(TPortId) {
+  return nullptr;
+}
+
+void FORTE_E_TRIG::triggerEvents(forte::core::CFBContainer* paContainer, TEventTypeID paEventType) {
```
azoitl: paContainer is not really needed you could call the getResoruce here.

src/stdfblib/events/E_TRIG_fbt.cpp
```diff
@@ -0,0 +1,121 @@
...
+void FORTE_E_TRIG::setInitialValues() {
+  var_EVENTTYPE = ""_STRING;
+}
+
+void FORTE_E_TRIG::executeEvent(const TEventID paEIID, CEventChainExecutionThread *const paECET) {
+  switch(paEIID) {
+    case scmEventREQID:
+      const TEventTypeID eventTypeId = CStringDictionary::getInstance().getId(var_EVENTTYPE.c_str());
+      if (eventTypeId != CStringDictionary::scmInvalidStringId) {
+        triggerEvents(getResource(), eventTypeId);
```
azoitl: use the paECET for triggereEvents so that you do not need to use the external events everywhere.

azoitl (changes requested):

azoitl:
> on bigger PRojects i get a lot of `ERROR: T#25000000000: External event queue is full, external event dropped!` Messages.

I would expect that, we need more memory to handle that. But with my comments from above you only need to make the internal event queue larger

franz-hoepfinger-4diac:
> > on bigger PRojects i get a lot of `ERROR: T#25000000000: External event queue is full, external event dropped!` Messages.
> 
> I would expect that, we need more memory to handle that. But with my comments from above you only need to make the internal event queue larger

Understand. 
yes, waiting for this Improvement. 
but still a big step forward.

franz-hoepfinger-4diac:
> > on bigger PRojects i get a lot of `ERROR: T#25000000000: External event queue is full, external event dropped!` Messages.
> 
> I would expect that, we need more memory to handle that. But with my comments from above you only need to make the internal event queue larger

you are true. 

using a Project with 2416 Blocks after putting 
`-DFORTE_EventChainExternalEventListSize=512`

i get 
`ERROR: T#2000000000: Event queue is full, event dropped!`


so it seems we need to increase both Queues now. 

later the External one will be unaffected right ? 
@MandKastner --> great work so far. 
@azoitl --> good reviewer . 



after: 

```
    -DFORTE_EventChainExternalEventListSize=512 \
    -DFORTE_EventChainEventListSize=512 \
```

it is gone.

MandKastner:
@franz-hoepfinger-4diac thank you for testing thats really helpful.


> 
> later the External one will be unaffected right ? @MandKastner --> great work so far. @azoitl --> good reviewer .
> 

Yes, i think the external event queue should not be affected anymore. I believe that you just have to increase the size of the internal one.

src/core/funcbloc.cpp
```diff
@@ -50,6 +50,9 @@ bool CFunctionBlock::initialize() {
 #ifdef  FORTE_SUPPORT_MONITORING
   setupEventMonitoringData();
 #endif //FORTE_SUPPORT_MONITORING
+  if (getFBInterfaceSpec().mEITypeNames != nullptr) {
+    setupInputConnectionTrackingData(getFBInterfaceSpec().mNumEIs);
+  }
```
mx990: Can we move the surrounding if and `getFBInterfaceSpec().mNumEIs` argument into `setupInputConnectionTrackingData()?

src/core/funcbloc.h
```diff
@@ -414,6 +414,32 @@ class CFunctionBlock : public forte::core::CFBContainer {
...
+        mInputEventConnectionCount[paEIID]--;
+      }
+    }
+
+    /*!\brief initialize the data structure which holds conneciton counts per pin
+     * \param paNumEIs number of eventInputs
+     */
+    void setupInputConnectionTrackingData(TEventID paNumEIs) {
+      mInputEventConnectionCount = std::make_unique<size_t[]>(paNumEIs);
+    }
```
mx990: I would suggest making `setupInputConnectionTrackingData()` protected.

mx990 (changes requested):

src/core/funcbloc.h
azoitl: yes it should definitly not be public.

azoitl (commented):
appart from @mx990 comments everything else looks good.

src/core/funcbloc.h
```diff
@@ -604,6 +623,15 @@ class CFunctionBlock : public forte::core::CFBContainer {
...
     const SFBInterfaceSpec &mInterfaceSpec; //!< Pointer to the interface specification
 
+    /*!\brief initialize the data structure which holds connection counts per pin
+    * \param paNumEIs number of eventInputs
+    */
+    void setupInputConnectionTrackingData(TEventID paNumEIs) {
+      if (getFBInterfaceSpec().mEITypeNames != nullptr) {
+        mInputEventConnectionCount = std::make_unique<size_t[]>(paNumEIs);
+      }
+    }
```
mx990: Please also move the `getFBInterfaceSpec().mNumEIs` argument into `setupInputConnectionTrackingData()`:
```suggestion
    void setupInputConnectionTrackingData() {
      if (getFBInterfaceSpec().mEITypeNames != nullptr) {
        mInputEventConnectionCount = std::make_unique<size_t[]>(getFBInterfaceSpec().mNumEIs);
      }
    }
```
or is there a reason why we would pass different values here?

mx990 (changes requested):

src/core/funcbloc.h
MandKastner: you are right there is no reason i missed that, thx

mx990 (approved):

azoitl (approved):

PR #301, opened 2025-01-11
Reset data inputs pointers in Adapter before interface is destroyed

cochicde:
Starting the PR mostly to see if the solution is the best. 

The context is the following:
- 2 FB with an compatible adapter each
- When the Adapter is created,  `CGenFunctionBlock<T>::setupFBInterface`  does its Voldermort-level-memory-magic, storing the data inputs and data outupts into member variables `mDIs` and `mDOs`
- Whent the connecction between the adapters is created, `mDIs = paPeer->mDOs;` is done in each FB, meaning each FB has the pointer from the other, which is reseted to `mLocalDIs` when disconnected.

Two problems: 
1. `mLocalDIs`  is set with `mDIs` in the constructor, but `mDIs` is set only afterwards in `initialize`, therefore this PE contrins the change in this `initialize` to store the updated value of `mDIs`
2. `disconnect`  (which resets `mDIs` with `mLocalDIs`) is called after the destructor of the Adapter, specifically  from `CAdapterConnection::performDisconnect` in its destructor. That means whent he destructor of the Adapter is called, `mDIs` still has the `mDOs` value of the connected FB, and when the interface is destroyed in `CGenFunctionBlock<T>::freeFBInterfaceData` really bad things happen. 
I tried calling `disconnect` in the destructor of the Adapter, but the call later to `performDisconnect` is not happy (probably because the FB was destroyed alreayd, but not sure) 

With the changes in the first commit of the PR, I don't get a crash. I'm surprised this was not detected before, so I'm not sure I'm doing something wrong. I used the  ReferenceExamples from the Examples repository. There `Ex2a_05_Adapter` and `Ex3a_05_Adapter` have this issue. The particular case I see is that the Adapter has both Data Inputs and Data Ouputs, which is not the case for example of the `ATimeout`, but I didn't test too deep to see if this affects the context in which the crash occurs.

In my opinion, the adapter should perform the disconnection in its destructor, and the `CAdapterConnection::performDisconnect` should not exist

### DIFF
--- a/src/core/adapter.cpp
+++ b/src/core/adapter.cpp
@@ -34,10 +34,12 @@ bool CAdapter::initialize() {
     return false;
   }
   setupEventEntryList();
+  mLocalDIs = mDIs;
   return true;
 }
 
 CAdapter::~CAdapter(){
+  mDIs = mLocalDIs;
   if (mIsPlug) {
     if (mAdapterConn != nullptr) {
       delete mAdapterConn;


azoitl:
As you bring up this point I'm not so shure if we should mirror the input and output arrays between plugs and sockets. Should we instead copy them over to have a stronger decoupling of both sides and a more correct with behavior, because now if an FB writes to an adapter it is immediately visible on the other side, and this would be wrong.

mx990:
Yes, this is something I wanted to fix for a long time but never got around to actually doing. I would also split adapters into separate classes for plugs and sockets. We could then add the required interface (and connection) variables directly in each class (as we do for regular FBs) and also perform proper sampling for events.

cochicde:
I don't know much about adaptaers or how they should behave. From the code perspective, it seems that the connection is little bit hacky. Separating into plug and sockets classes seems to be a nice solution.

Now, what do we do until then? As it's now at least for me forte crashes, so technically it doesn't support adapters. 

I'd recommend merge this as a quick fix to have it working and open an issue about this issue to proper work on it.

azoitl (approved):

azoitl:
You have a point here. Unless you would like to do it I don't see that happen soon. Will merge it therefore.

PR #307, opened 2025-01-21
Fix configuration of OPC UA A&C and Events layer

m-meingast:


src/com/opc_ua/CMakeLists.txt
```diff
@@ -64,10 +64,18 @@ if (FORTE_COM_OPC_UA)
...
     # OPC UA Event Layer
-    forte_add_network_layer(OPC_UA_Events ON "ua_ev" COPC_UA_Event_Layer opcua_event_layer "Enable OPC UA Event Support")
-    forte_add_network_layer(OPC_UA_AC ON "ua_ac" COPC_UA_AC_Layer opcua_ac_layer "Enable OPC UA A&C Support")
+    forte_add_network_layer(OPC_UA_Events OFF "ua_ev" COPC_UA_Event_Layer opcua_event_layer "Enable OPC UA Event Support")
+    forte_add_network_layer(OPC_UA_AC OFF "ua_ac" COPC_UA_AC_Layer opcua_ac_layer "Enable OPC UA A&C Support")
     forte_add_sourcefile_hcpp(opcua_layer opcua_local_handler opcua_helper opcua_remote_handler opcua_handler_abstract opcua_client_information opcua_action_info opcua_objectstruct_helper 
-        opcua_event_layer opcua_ac_layer opcua_client_config_parser)
+        opcua_client_config_parser)
+    
+    if (FORTE_COM_OPC_UA_AC)
```
azoitl: these two ifs should not be necessary as the add network layer is allready adding the source and hpp files to the build.

m-meingast: The issue here is that when I remove the ifs, then `opc_ua_ac_layer` is compiled regardless of the state of the `FORTE_COM_OPC_UA_AC` flag. The build fails if a open62541 build is used without enabled A&C.

azoitl: This must then be another problem, because the add network layer macro does that exactly for you

azoitl: I just tested your patch on my machine and removed the ifs and it worked exactly as it should.

ernstblechaPT: the ifs are not necessary after `opcua_event_layer` and `opcua_ac_layer` are removed on line 70 - otherwise adding those to the build manually will override whatever checks are done in `forte_add_network_layer`

m-meingast: Ah yes, it works for me as well now. I still had them in the configuration.
Should we also remove the `opcua_layer` in line 69 to keep it consistent with the CMake configurations of other layers?

azoitl: As we are cleaning up yes please.

azoitl (approved):

PR #310, opened 2025-01-26
Add debug device

cochicde:
This PR allows pausing, continuing and adding and removing breakpoints. 

1. The FakeEcet was adapted with the desired behaviour when debugging
2. The OPCUA_MGR was changed to allow new methods to be added to the ones it implements
3. A DebugMGR builds on top of the OPCUA_MGR and adds the following functionalities:
   - enable/disable remote control (pausing/ resuming forte)
   - add/remove breakpoints: hitting on a breakpoints will halt forte and all ecets are set to remote control
 4. A DebugDevice simply contain a DebugMGR and is added to the list of devices that can be compiled

### DIFF
--- a/src/stdfblib/ita/OPCUA_MGR.cpp
+++ b/src/stdfblib/ita/OPCUA_MGR.cpp
@@ -12,9 +12,8 @@
  *******************************************************************************/
 
 #include "OPCUA_MGR.h"
-#include <string>
-#include <vector>
-#include "OPCUA_DEV.h"
+
+#include "device.h"
 #include "stringdict.h"
 
 /* Management and Resource Type */
@@ -251,7 +250,7 @@ const std::map<EMGMResponse, UA_StatusCode> OPCUA_MGR::scResponseMap = {
 
 const UA_UInt16 OPCUA_MGR::smNamespaces[] = {0, 1};
 
-OPCUA_MGR::OPCUA_MGR(OPCUA_DEV& paUaDevice) :
+OPCUA_MGR::OPCUA_MGR(CDevice& paUaDevice) :
   mUaDevice(paUaDevice),
     mUaHandler(paUaDevice.getDeviceExecution().getExtEvHandler<COPC_UA_Local_Handler>()){
 }
@@ -295,6 +294,7 @@ EMGMResponse OPCUA_MGR::createIEC61499MgmtObject(UA_Server* paServer) {
 #ifdef FORTE_SUPPORT_MONITORING
   if (addReadWatchesMethod(paServer) != EMGMResponse::Ready) return eRetVal;
 #endif // FORTE_SUPPORT_MONITORING
+  if(addExtraMethods(paServer, mMgmtTypeId, mExtraMgmMethods) != EMGMResponse::Ready) return eRetVal;
   return addMgmtObjectInstance();
 }
 
@@ -351,6 +351,7 @@ EMGMResponse OPCUA_MGR::createIEC61499ResourceObjectType(UA_Server* paServer) {
   if (addForceValueMethod(paServer) != EMGMResponse::Ready) return eRetVal;
   if (addClearForceMethod(paServer) != EMGMResponse::Ready) return eRetVal;
 #endif // FORTE_SUPPORT_MONITORING
+  if(addExtraMethods(paServer, mResourceTypeId, mExtraResourceMethods) != EMGMResponse::Ready) return eRetVal;
   return EMGMResponse::Ready;
 }
 
@@ -1215,7 +1216,7 @@ UA_StatusCode OPCUA_MGR::onClearForce(UA_Server*,
 
 EMGMResponse OPCUA_MGR::addMethodNode(UA_Server* paServer, char* paMethodNodeName, UA_NodeId paParentNodeId, UA_MethodAttributes paAttr, 
     const UA_Argument *paInArgs, size_t paInArgSize, const UA_Argument *paOutArgs, 
-    size_t paOutArgSize, UA_MethodCallback paCallback) {
+    size_t paOutArgSize, UA_MethodCallback paCallback, void* nodeContext) {
   EMGMResponse eRetVal = EMGMResponse::Ready;
   UA_NodeId methodId;
   UA_StatusCode status = UA_Server_addMethodNode(paServer, UA_NODEID_STRING(smNamespaces[1], paMethodNodeName),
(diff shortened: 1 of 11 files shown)


src/stdfblib/ita/debug_device/DebugDevice.h
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+#include "RMT_DEV.h"
+#include "DebugMGR.h"
+
+/**
+ * @brief Device that adds debug commands to the device. The commands are defined in DebugMGR
+ * 
+ */
+class DebugDevice : public RMT_DEV {
```
azoitl: why remote device when you use the OPC UA DEV_MGR?

src/stdfblib/ita/debug_device/DebugDevice.h
```diff
@@ -0,0 +1,34 @@
...
+public:
+
+  DebugDevice(const std::string &paMGRID = "localhost:61499");
```
azoitl, line 29: shouldn't this be the default opc ua setup?

cochicde: I double check the OPCUA_DEV and it seems I changed it wronly in my last commit. It seems this value gets passed to the data input of the RMT_DEV which ultimately is being used by the RMT_RES as incoming port. So, I seem this is correct, and the one on OPCUA_DEV is now wrong. 

The `opc.tcp://localhsot:4840` is being set in the opcua local handler, but this parameter is not passed to it in any form

or did I miss something? if not I would revert the changes I did last time in OPCUA_DEV and use also there `localhost:61499`

src/stdfblib/ita/debug_device/DebugDevice.h
cochicde: The OPCUA_DEV is not offering any extra inputs or capabilities than the ones offered by the `OPCUA_MGR`, so I found it better to have the OPCUA_MGR as a member instead of getting it via inheritance. Also, if one would want to add even more capbilities on top of debugging (something I'm currently working on, that's when I realized this seemed a better choice), one would need to inherit again from `DebugDevice`, which is not a good practice.

src/stdfblib/ita/debug_device/DebugDevice.h
azoitl: I'm not sure if we have here a general design issue.

src/stdfblib/ita/debug_device/DebugDevice.h
azoitl: Sorry I just noticed my comment was way to short. I totally agree with you that aggregation is much better then in inheritance. What I wanted to say is that I don't think you should inherit your DebugDevice from RM_DEV and also not from OPCUA_DEV. I meant just from CDevice. This would be the cleanest and most correct solution.

src/stdfblib/ita/debug_device/DebugDevice.h
azoitl: > I'm not sure if we have here a general design issue. Especially under consideration of my comment above.

src/stdfblib/ita/debug_device/DebugDevice.h
cochicde: I see. Yes, that was probably a lazy copy paste from opcua dev without checking properly

src/stdfblib/ita/debug_device/DebugDevice.h
cochicde: I think the `paMGRID` was always the only possible configuration for a forte device. One idea I can think of is to have a `DeviceConfig` class which is passed as a parameter and the device stored it and offers it in its interface as a readonly, so it's not changed after creation.

This `DeviceConfig` class could be a simple dictionary<string, string> and needed config (such as `paMGRID` or the endpoint needed by the opcua server) can be retrieved by the class that needs it.

azoitl: For my your proposal looks like an overkill. The MGR_ID was always some configuration parameter that is used in the 4diac IDE to tell 4diac IDE how to reach the device and on 4diac FORTE on how to setup any listening communication system so that 4diac IDE can connection and send management commands. What speaks against the fact that we use a device specific default MGR_ID and allow the users to specify a specific one via a command line parameter. With that 90% of the existing code would already work and we just have to correct the OPC UA device and maybe clear up some commandline parsing. But maybe that is better discussed next week in person at the 4days of Eclipse 4diac Winter School.

cochicde: loved the last sentence!

cochicde: but I changed it back in the opcua dev again as it was before so the normal MGR still works in parallel to the opcua one

src/stdfblib/ita/OPCUA_MGR.cpp
```diff
@@ -1323,3 +1324,24 @@ void OPCUA_MGR::parseDestinationName(const std::string& paDestination, std::vect
   }
   paFbName.push_back(dst.substr(0, index));
 }
+
+void OPCUA_MGR::addExtraMgmMethod(const MethodInformation& paMethod){
+  mExtraMgmMethods.emplace_back(paMethod);
```
azoitl, line 1329: what is the advantage of having that list. These methods only have to be crated once. This could be easily don by overriding the àddExtraMethods` function.

azoitl: I would do the same for the resource methods

azoitl (changes requested):

src/stdfblib/ita/OPCUA_MGR.cpp
cochicde: the main idea of the lists is that you can have your own `MGR` class easily compose the `OPCUA_MGR` by just adding new  methods. I have another device on my pipeline of devices which needs the `debug` methods plus some more, so they just keep adding methods https://github.com/cochicde/4diac-forte/blob/replay-algorithm/src/stdfblib/ita/replay/ReplayMGR.cpp#L25

It's true that they're used only once, so it should be fine to just remove them all after they're used. Overriding the method sounds like there will be inheritance involved, which would difficult if you want to mix methods from one or other devices.

azoitl: Ah I missed that you don't want to use the OPC UA MGR in an inheritence relation. Thx for the clarification.

azoitl (approved):

PR #321, opened 2025-02-20
Add parseTypeName method to CommandParser

m-meingast:
Use parseTypeName to parse the CreateFB request instead of parseIdentifier

### DIFF
--- a/src/stdfblib/ita/CommandParser.cpp
+++ b/src/stdfblib/ita/CommandParser.cpp
@@ -297,7 +297,7 @@ bool CommandParser::parseFBData(char *paRequestPartLeft){
     if(acBuf != nullptr){
       if(acBuf[1] != '*'){
         ++acBuf;
-        i = parseIdentifier(acBuf, mCommand.mSecondParam);
+        i = parseTypeName(acBuf, mCommand.mSecondParam);
         if(-1 != i){
           acBuf = strchr(&(acBuf[i + 1]), '\"');
           if(acBuf != nullptr){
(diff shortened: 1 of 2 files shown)


src/stdfblib/ita/CommandParser.cpp
```diff
@@ -338,6 +338,18 @@ int CommandParser::parseIdentifier(char *paIdentifierStart, forte::core::TNameId
   return -1;
 }
 
+int CommandParser::parseTypeName(std::string paTypeName, forte::core::TNameIdentifier &paIdentifier){
+  size_t endIndex = paTypeName.find('"');
+  if(endIndex == std::string::npos) {
+    return -1;
+  }
+  std::string result = paTypeName.substr(0, endIndex);
+  if(!paIdentifier.pushBack(CStringDictionary::getInstance().insert(result.c_str()))){
```
mx990: I would suggest using https://github.com/eclipse-4diac/4diac-forte/blob/d1afbdea0e49191ed9de1f37682ef78822eebcd0/src/core/stringdict.h#L60, so you can avoid having to create an intermediary string.

src/stdfblib/ita/CommandParser.cpp
```diff
@@ -338,6 +338,18 @@ int CommandParser::parseIdentifier(char *paIdentifierStart, forte::core::TNameId
   return -1;
 }
 
+int CommandParser::parseTypeName(std::string paTypeName, forte::core::TNameIdentifier &paIdentifier){
```
mx990: Please note that this will allocate and copy the entire buffer until the end each time this is called. Maybe you could use a [std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view/) instead?

mx990 (changes requested):

src/stdfblib/ita/CommandParser.cpp
m-meingast: Yeah, that makes sense. I replaced the strings with string_view objects. Thanks for pointing that out.

src/stdfblib/ita/CommandParser.cpp
m-meingast: This is also useful with the string_views, to set the correct string length

mx990 (approved):

azoitl (approved):

Issue #311, opened 2025-01-31
GENeric function block can not be created because cgIdentifierLength is too small.


murinsel:
In datatype.h the definition of cgIdentifierLength is set to 32.

```
const unsigned int cgIdentifierLength = 32;
typedef char TIdentifier[cgIdentifierLength + 1]:
```

This length also limits in typelib.cpp the function block type name for generic function blocks:

  ```
  TIdentifier acGenFBName = { "GEN_" };
...
      memcpy(&(acGenFBName[4]), acTypeBuf, nCopyLen);
      acGenFBName[cgIdentifierLength] = '\0';
      poToCreate = findType(CStringDictionary::getInstance().getId(acGenFBName), mFBLibStart);
```

But if the generic type is placed into a sub folder the type name contains also the folder names and gets easily longer than the 32 characters.
For example my type name created during export from the IDE is:

GEN_DAQ__Signal__SigGrpRangeReader

So this name will be truncated an findType() fails to find the type.
One solution would be to increase the default cgIdentifierLength to say 80 or is it possible to make this configurable via Cmake?

**Comment by murinsel** [comment]
The following works, but is this the propper way to do it?
I can not do it via forte_config.h.in because that already includes <datatype.h> at the top. So I added it as compile_definition in the main CMakeLists.txt like shown below:

In the main CMakeLists.txt:

```
set(FORTE_IDENTIFIER_LENGTH 32 CACHE STRING "The length an identifier may have on order that it is unique. Identifiers may be FB-Type Names, FB-Names, var names,...")
add_compile_definitions(FORTE_IDENTIFIER_LENGTH=${FORTE_IDENTIFIER_LENGTH})

```
In datatype.h:

```
#ifndef FORTE_IDENTIFIER_LENGTH
#define FORTE_IDENTIFIER_LENGTH 32
#endif
const unsigned int cgIdentifierLength = FORTE_IDENTIFIER_LENGTH; //!<This value holds the length an identifier may have on order that it is unique. Identifiers may be FB-Type Names, FB-Names, var names,...

```

azoitl:
@murinsel PR #323 should fix your problem in a hopefully nicer way. 

One thing that I wanted to add is that the package names that 4diac IDE now suggests when creating types are just a suggestion. We decided that it could be a good idea that they should follow the folder structure. But you are free to give any package structure or none at all if this fits your use case better.

PR #323, opened 2025-02-20
Fix Generic Type Creation for Packages #311

azoitl:
With package names and configurable demuxers the type names got much longer. The generic type creation didn't handle that well. This rework uses a dedicated std::string for this purpose.

Fixes: https://github.com/eclipse-4diac/4diac-forte/issues/311

### DIFF
--- a/src/core/typelib.cpp
+++ b/src/core/typelib.cpp
@@ -25,6 +25,8 @@
 #include "adapter.h"
 #include <stddef.h>
 
+using namespace std::string_literals;
+
 CTypeLib::CTypeEntry::CTypeEntry(CStringDictionary::TStringId paTypeNameId) :
   mTypeNameId(paTypeNameId),
   mNext(nullptr){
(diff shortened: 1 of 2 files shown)


mx990 (approved):

src/core/typelib.cpp
```diff
@@ -143,13 +145,10 @@ CFunctionBlock *CTypeLib::createGenericFB(CStringDictionary::TStringId paInstanc
     return nullptr;
   }
 
-  ptrdiff_t typeNameLen = underScore - acTypeBuf;
-  std::string genFBName;
-  genFBName.reserve(4 + typeNameLen);
-  genFBName += "GEN_";
-  genFBName.append(acTypeBuf, typeNameLen);
+  mGenFBName.resize(4);  //only keep the GEN_
+  mGenFBName.append(acTypeBuf, underScore - acTypeBuf);
```
mx990: You realize that using a shared buffer means this will no longer be thread-safe?

mx990: I would also like to point out that `std::string` stores small strings by-value without allocating heap memory using *small string optimization*. The threshold is implementation-specific, ranging from 15 on gcc and MSVC to 22 on clang, see also [An informal comparison of the three major implementations of std::string](https://devblogs.microsoft.com/oldnewthing/20240510-00/?p=109742) for a comparison of different implementations.

azoitl: I know that there is small string optimization. This PR is however targeting exactly the opposite. It tries to imrpove the situation for longer strings that occure when you have generic FBs in packages. 

Yes I know that it is not thread save anymore. However I think the typelib was never really that  thread save.

If you think I over optimized here I'm happy to roll it back. That is the reason I did it in a separate commit.

mx990: > I know that there is small string optimization. This PR is however targeting exactly the opposite. It tries to imrpove the situation for longer strings that occure when you have generic FBs in packages.

I agree that we should *support* longer names for generic types, but I don't really see a strong need to *optimize* for those cases, at least for now.

> Yes I know that it is not thread save anymore. However I think the typelib was never really that thread save.

In that case, I believe we should aim to make it *really* thread-safe, rather than making it even worse and thus requiring more work to do so later.

> If you think I over optimized here I'm happy to roll it back. That is the reason I did it in a separate commit.

I believe we should roll it back for now. We can also always consider this as a future optimization, should the need arise.

azoitl: you have here some points. Rolled it back.

src/core/typelib.cpp
```diff
@@ -113,54 +113,60 @@ CAdapter *CTypeLib::createAdapter(CStringDictionary::TStringId paInstanceNameId,
...
+  if(nullptr != newFB && !newFB->initialize()) {
+    deleteFB(newFB);
+    newFB = nullptr;
+  }
+
+  return newFB;
+}
+
+CFunctionBlock *CTypeLib::createGenericFB(CStringDictionary::TStringId paInstanceNameId, CStringDictionary::TStringId paFBTypeId, forte::core::CFBContainer &paContainer) {
+  const char *acTypeBuf = CStringDictionary::get(paFBTypeId);
```
MartinMelikMerkumians: here you still have a type specifier `ac`.

I think this could also be `const char *const`

src/core/typelib.cpp
```diff
@@ -113,54 +113,60 @@ CAdapter *CTypeLib::createAdapter(CStringDictionary::TStringId paInstanceNameId,
...
+  const char *underScore = getFirstNonTypeNameUnderscorePos(acTypeBuf);
```
MartinMelikMerkumians: I think this could also be `const char *const`

src/core/typelib.cpp
```diff
@@ -113,54 +113,60 @@ CAdapter *CTypeLib::createAdapter(CStringDictionary::TStringId paInstanceNameId,
...
+  if(underScore == nullptr) {
+    // We found no underscore in the type name therefore it can not be a generic type
+    mLastErrorMSG = EMGMResponse::UnsupportedType;
+    return nullptr;
+  }
+
+  ptrdiff_t typeNameLen = underScore - acTypeBuf;
+  std::string genFBName;
+  genFBName.reserve(4 + typeNameLen);
+  genFBName += "GEN_";
```
MartinMelikMerkumians: you could specify this directly as a string literal with "GEN_"s;

MartinMelikMerkumians (changes requested):
Looks good to me, I only added some minor improvement ideas

src/core/typelib.cpp
azoitl: done

src/core/typelib.cpp
azoitl: done

src/core/typelib.cpp
azoitl: thx for pointing this out. done.

MartinMelikMerkumians (approved):

PR #325, opened 2025-02-22
Generated E_TABLE_CTRL from IDE to also use EInit

MoritzO02:
Changed Event to EInit in IDE and exported to forte

Issue https://github.com/eclipse-4diac/4diac-ide/issues/943

### DIFF
--- a/src/stdfblib/events/E_TABLE_CTRL_fbt.cpp
+++ b/src/stdfblib/events/E_TABLE_CTRL_fbt.cpp
@@ -1,11 +1,12 @@
-/*******************************************************************************
- * Copyright (c) 2017 fortiss GmbH
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * http://www.eclipse.org/legal/epl-2.0.
- *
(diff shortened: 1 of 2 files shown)


diplfranzhoepfinger (changes requested):
Pls do not change this to EInit

diplfranzhoepfinger:
https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/stdfblib/events/E_TABLE_CTRL_fbt.h

this is however wrong. 
as Discussion here: 
https://github.com/eclipse-4diac/4diac-ide/pull/1074

> The standard defines this block to have INIT as the name for that event. Can you therefore not change things that are defined in the standard.

so it should be INIT also in the C++/h 
@azoitl right ? 

but INIT of type Event, not Type EInit
@MoritzO02

azoitl:
I fear it would be better to keep these blocks as they are in the standard.

diplfranzhoepfinger:
> I fear it would be better to keep these blocks as they are in the standard.

yes, so, @MoritzO02 pls change it back to INIT, as in the Standard, but Type Event not EInit.

PR #329, opened 2025-02-24
Esc storage

azoitl:
This is a rework of my old gerrit commit: https://git.eclipse.org/r/c/4diac/org.eclipse.4diac.forte/+/201360

### DIFF
--- a/src/core/util/string_utils.cpp
+++ b/src/core/util/string_utils.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013 fortiss GmbH
+ * Copyright (c) 2013, 2025 fortiss GmbH, Johannes Kepler University Linz
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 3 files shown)


src/core/util/string_utils.cpp
```diff
@@ -19,6 +18,21 @@
 
 #include <errno.h>
 #include <string.h>
+#include <array>
+#include <map>
+
+const auto scCharacters2Escape = std::array { '"', '\'', '&', '<', '>' };
```
cochicde: do  you really need these two extra arrays? can't you just have the values directly in the constructor of the map?

azoitl: we may need the oppsite map later so I would like to keep the arrays for now.

azoitl: Just noticed my answer was even wrong. Yes I need it for unescaping.

cochicde: if you needed for later is fine, but for unscaping you can iterate directly over the map:

```
for(const auto& [characters2Escape, escapeSequence] : scEscapeMap){
  if(0 == strncmp(runner, escapeSequence, strlen(escapeSequence))){
    toCopy = characters2Escape;
    toMove = strlen(escapeSequence);
    break;
   }
}
```

azoitl: After a second thought and seeing your code I think it is better to not have them for now. Removed them and used your snippet.

azoitl: While I was pressing enter on my last comment I realized that we are doing here a lot of `strlen`s on constant strings. I just pushed now a version where this string lenght calculation is pushed to compile time. What do you think about this?

src/core/util/string_utils.cpp
```diff
@@ -1,14 +1,13 @@
...
  *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
- *   Alois Zoitl
- *    - initial API and implementation and/or initial documentation
+ *   Alois Zoitl - initial API and implementation and/or initial documentation
  *******************************************************************************/
 #include "string_utils.h"
 #include <forte_dint.h>
```
MartinMelikMerkumians: You could change the include from <> to "" as this is a project include

src/core/util/string_utils.cpp
```diff
@@ -19,6 +18,18 @@
 
 #include <errno.h>
 #include <string.h>
+#include <array>
+#include <map>
+
+const auto scCharacters2Escape = std::array {"&quot;", "&apos;", "&amp;", "&lt;", "&gt;"};
```
MartinMelikMerkumians: You could try to make this constexpr

MartinMelikMerkumians (changes requested):

src/core/util/string_utils.cpp
azoitl: Thx for pointing this out. Changed it to a realtive path to be even more aligned with @jwalt updates on include paths.

src/core/util/string_utils.cpp
azoitl: good idea. Just changed.

MartinMelikMerkumians (approved):
LGTM

src/core/util/string_utils.cpp
cochicde: amazing!

azoitl: thx

PR #331, opened 2025-02-25
Reworked XML response generation to better reuse allocated memory

azoitl:
Generating the XML response used an freshly created string for each response. This resulted on platforms with little memory to out of memory issues. With this commit the memory of the final storage is used to generate this string. This reduces the number of allocations dramatically increasing speed and reduces memory consumption.

### DIFF
--- a/src/stdfblib/ita/CommandParser.cpp
+++ b/src/stdfblib/ita/CommandParser.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Jose Cabral
+ * Copyright (c) 2024, 2025 Jose Cabral
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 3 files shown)


cochicde (approved):

m-meingast (approved):

MartinMelikMerkumians (changes requested):
Hi,

as discussed via Discord, I think it would be better to change the parser to take a reference to CIEC_STRING and make the method private again, so string storage and containing object cannot be separated in a read/write fashion on the containing objects memory

azoitl:
> Hi,
> 
> as discussed via Discord, I think it would be better to change the parser to take a reference to CIEC_STRING and make the method private again, so string storage and containing object cannot be separated in a read/write fashion on the containing objects memory

thx for the feedback and your assessment. I changed it now accordingly.

MartinMelikMerkumians (approved):

PR #333, opened 2025-02-25
use std::vector/inplace_vector as backing storage for stringdict

jwalt:


src/core/stringdict.cpp
```diff
@@ -13,193 +14,121 @@
...
+#include <stringlist.cpp>
 
 DEFINE_SINGLETON(CStringDictionary)
 
-CStringDictionary::CStringDictionary(){
 #ifdef FORTE_STRING_DICT_FIXED_MEMORY
-  mStringBufAddr = scmConstStringBuf;
-  mStringIdBufAddr = scmIdList;
+#include "core/util/inplace_vector.h"
+static forte::core::util::inplace_vector<CStringDictionary::TStringId, FORTE_STRINGDICT_MAXNROFSTRINGS> scmIdList = INDEX_LIST;
```
azoitl: any reason why the lists are not members anymore?

azoitl: as you moved them out of the class the scm prefix is not valid any more as they are not const and no member.

azoitl (changes requested):

src/core/stringdict.cpp
jwalt: The lists change every time a string is added, which will happen regularly during normal FB development. In the fixedstringdict case, they will lead to recompilation of all source files due to the compile-time size changing.

jwalt: I kept the prefix as it was inaccurate anyways -- it wasn't always const, depending on compile options.

azoitl (approved):

PR #334, opened 2025-02-25
Feature/small cleanups

jwalt:


src/core/devexec.h
```diff
@@ -88,7 +89,7 @@ class CDeviceExecution {
      *
      * The element 0 is always the timer event source.
      */
-    SEventHandlerElement mRegisteredEventHandlers[cgNumberOfHandlers];
+    std::vector<SEventHandlerElement> mRegisteredEventHandlers;
```
azoitl, line 92: as we have a given fixed size we could use here either a std::array or reserve the required size in the constructor of the DeviceExecutor class.

azoitl (changes requested):

azoitl (approved):

PR #338, opened 2025-02-27
make string ID generation more explicit and robust

jwalt:
The current way of generating string IDs and strings for `core/stringdict.hpp` is quite involved, using lots of generated code and fragile parsing. This PR completely overhauls this, using explicit declarations for string IDs, eliminating the need for generated header files entirely and simplifying parsing, makign it more robust at the same time.


mx990:
If we are reworking the string id handling, have you considered [user-defined literals](https://en.cppreference.com/w/cpp/language/user_literal)? I would also like to point out another interesting approach described in [Useful Properties of String Interning in C++](https://alexpolt.github.io/intern.html).

jwalt:
Great suggestions, thanks. A major goal is better (repeated) compile times, since the strings are all over our code base. That's why this design is using just the preprocessor. However, I'll evaluate your ideas as they sound interesting.

azoitl:
While I like the idea of improving our string dict and string id handling I wanted to point out that the compile time optimization is not the main point of this features. The main point is that we get automatically an order list of all compile time knowable strings put them into a fast searchable data structure which can be extended with new strings during runtime so that during deployment strings can quickly be searched for and that an ID can be used instead of a string. I currently don't see that for methods proposed by @mx990 but I may be wrong.

jwalt:
> While I like the idea of improving our string dict and string id handling I wanted to point out that the compile time optimization is not the main point of this features.

Well, reducing compile-time complexity is an important goal of this particular change. This of course doesn't change the fact that overall, ...

>  The main point is that we get automatically an order list of all compile time knowable strings put them into a fast searchable data structure which can be extended with new strings during runtime so that during deployment strings can quickly be searched for and that an ID can be used instead of a string. I currently don't see that for methods proposed by @mx990 but I may be wrong.

Indeed. While interning strings based on types is a neat technique, our use case requires the ability to intern strings at run-time, so a pure compile-time solution can't work.

As for the suggestion of UDL, I don't see an easy way to build a compile-time efficient representation that works well with the run-time stringdict either (you can't synthesize symbol names via a UDL). The macro solution is easy to understand and fast. On the other hand, it makes future overhauls of that system pretty easy, so if a future better solution comes along, this change will make a switch easier. Therefore I propose to leave the patch as it is.

mx990:
Please have a look at the example in section *String interning* from the proposal [N3599](https://open-std.org/JTC1/SC22/WG21/docs/papers/2013/n3599.html) on how to combine the template-based approach with the ability to intern strings at run-time. I have created a working version at https://godbolt.org/z/W6anzv8YY so you can see it in action.

This is only a basic proof-of-concept, of course, and we should, at the very least:
- use a different data structure than a `std::map` for run-time lookup,
- wrap the plain `const char*` to distinguish interned strings from normal ones.

However, I see several advantages over the current approach:
- no need to define additional symbols, whether in generated files or manually,
- no special handling needed for shared libraries, since the templates are handled by the linker,
- the ability to handle compile-time strings that are not valid C++ identifiers.

jwalt:
I have fleshed this out to an example that should fulfill all requirements:

1. log-n lookup `const char *` -> `string_id` (a pointer-size object)
2. constant-time lookup `string_id` -> `const char *` (because a `string_id` is essentially a `const char *`)
3. compile-time resolving of `"foo"_strid` into a valid `string_id`
4. ability to add new `string_id`s at run-time
5. type safety (only comparisons between `string_id` possible)
6. as already mentioned, eliminating the need for generated code

https://godbolt.org/z/qnb3cYo9d

However, we can't use this. It is invalid C++: N3599 has not been standardised and seems to be forgotten by standardisation committees. While GCC and clang provide this as compiler extensions even in C++17 mode, older but still relevant GCC versions can't compile the code, and MSVC doesn't seem to support it either.

Beside this, I am unsure if we gain enough by this. I see several drawbacks:

1. the code is much more complex
2. it has higher startup time (N std::unordered_map insertions versus an N-int linear copy)
3. it has higher compile and link time (especially the latter is already quite bad in FORTE)
4. special-case code for the fixed-memory-size variant would be needed (which I only just managed to eliminate)

mx990:
> However, we can't use this. It is invalid C++: N3599 has not been standardised and seems to be forgotten by standardisation committees. While GCC and clang provide this as compiler extensions even in C++17 mode, older but still relevant GCC versions can't compile the code, and MSVC doesn't seem to support it either.

You are right, it seems the C++ standard went with a slightly different approach[^1][^2]. I have adapted your example accordingly: https://godbolt.org/z/xGWv4xTev

If we ultimately decide to keep the current implementation, I am still glad that we could explore an interesting alternate approach.

[^1]: see *string literal operator template* in https://en.cppreference.com/w/cpp/language/user_literal#Literal_operators
[^2]: supported starting with gcc 10.3, clang 12, and msvc 19.30, according to tests on Godbolt

jwalt:
> supported starting with gcc 10.3, clang 12, and msvc 19.30, according to tests on Godbolt [↩](https://github.com/eclipse-4diac/4diac-forte/pull/338#user-content-fnref-2-7b38d4221c9ef916613984c18803a46f)

Unfortunately, that feature is C++20. Also, there are mainstream LTS Linux distributions that ship gcc<10.3 and clang<12 (debian oldstable for example)

BTW, I made a mistake when stating the time complexity, it is even average constant time, better than the current implementation. Maybe we can use most of our design but leave out the UDL, using a traditional constexpr `stringid(const char*)` function or something similar?

mx990:
It seems we could make it work with C++17 using a macro, based on the approach initially presented in [Useful Properties of String Interning in C++](https://alexpolt.github.io/intern.html). I have adapted the example accordingly: https://godbolt.org/z/f9d1boej3

jwalt:
That looks great! And it compiles on really old compilers (at least down to gcc-7.5 on godbolt). I am curious how much overhead (time/code size) this generates. It will take some days, but I will try to benchmark both variants in various configurations (debug, release, LTO etc)

jwalt:
So, I made some test using all three implementations: the old stringdict, this macro-based PR, and an adaptation of the template-based godbolt code. I have measured build times (single-threaded build, full-featured FORTE), stripped binary size, and execution time of the `StringDictTests` test suite of `forte_test` (`.../tests/forte_test -t StringDictTests -l all | grep time`). Of the latter I took the median value of 10 runs, rounded to 2 significant digits.



|              | compile time | binary size | test time |
|-------|---------|--------|--------|
| old -O3 -flto | 18m11s         |  7499552 |
| old -Os         | 15m38s         |  9715712 | 1.0ms |
| macro -O3 -flto | 15m27s         |  7446336 |
| macro -Os         | 15m34s         |  9715712 | 1.0ms |
| template -O3 -flto | 18m02s         |  7601728 |
| template -Os         | 15m58s         |  9969440 | 1.5ms |

In summary: macros are strictly better than the old code, as they behave almost the same in most metrics but compile 15% faster at maximum optimisation. Templates generate ~2% bigger binaries and compile ~3% slower when not using LTO. While templates *should* scale better with the size of the dictionary (due to the unordered_map instead of a sorted list), it seems that the constant overhead is significant. Given that the tests already fill the stringdict with several thousand strings, a workload where templates perform better is probably pretty exotic. Then again, the test used for time measurement is not a good performance benchmark. Finally, templates have the advantage of no code generation needed, while macros have the advantage that fixed-size memory is implemented.

Overall, I propose to use the current PR as it is now. It is an unconditional improvement over the current state, and the macros allow easy migration if someone finds a template-based solution that further improves metrics.

mx990:
Thank you for taking the time to do a thorough comparison of each alternative implementation. I agree with your conclusions but would suggest to wait until we also have a corresponding change for the forte_ng exporter ready.

@azoitl I believe you said you were already working on a related PR? Otherwise, I could also have a look?

azoitl:
> Thank you for taking the time to do a thorough comparison of each alternative implementation. I agree with your conclusions but would suggest to wait until we also have a corresponding change for the forte_ng exporter ready.
> 
> @azoitl I believe you said you were already working on a related PR? Otherwise, I could also have a look?

@mx990 yes I already have half of it finished. I just need to think how I generate the `USE_STRING_ID` nicely. 


@jwalt also from my side many thanks for the through evaluation. I learned a lot. Could you be so kind and fix the merge conflicts while I'm looking at the exporter.

mx990:
> I just need to think how I generate the USE_STRING_ID nicely.

My suggestion would be to handle this using a `Set<String> getStringIds(Map<?, ?> options)` in both `ForteLibraryElementTemplate` and `ILanguageSupport`, similar to `getDependencies` for `#include` directives.

jwalt:
> @jwalt also from my side many thanks for the through evaluation. I learned a lot. Could you be so kind and fix the merge conflicts while I'm looking at the exporter.

I will push an update when you are ready for merging. The last commit is auto-generated from a script, I will re-run it once we are ready to go.

azoitl:
> > I just need to think how I generate the USE_STRING_ID nicely.
> 
> My suggestion would be to handle this using a `Set<String> getStringIds(Map<?, ?> options)` in both `ForteLibraryElementTemplate` and `ILanguageSupport`, similar to `getDependencies` for `#include` directives.

yes I had something like this in mind.

azoitl:
A first version of the update exporter is available in: https://github.com/eclipse-4diac/4diac-ide/pull/1154

azoitl (approved):

PR #339, opened 2025-02-27
Add a multi device

cochicde:
Allows changing the device being executed but at runtime and remotely. Instead of restarting forte with the right device, the desired one can be selected and started.

The multi device is a wrapper over all other devices, basically controlling the lifetime of another device. It offers 2 opcua methods as interface:
- Set default device -> allows setting the device to be created
- RestartDevice: restart the device with the selected default device


ernstblechaPT (approved):
closes #270

azoitl (approved):

PR #340, opened 2025-02-27
Generated embrick FBs from IDE to also use EInit

MoritzO02:

### DIFF
--- a/src/modules/embrick/types/EBBusAdapter.cpp
+++ b/src/modules/embrick/types/EBBusAdapter.cpp
@@ -1,15 +1,22 @@
-/*******************************************************************************
- * Copyright (c) 2016 - 2018 Johannes Messmer (admin@jomess.com), fortiss GmbH
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * http://www.eclipse.org/legal/epl-2.0.
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *   Johannes Messmer - initial API and implementation and/or initial documentation
- *   Jose Cabral - Cleaning of namespaces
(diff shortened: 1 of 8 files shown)


azoitl:
Why are you always changing the copyright headers and not keeping the original ones. By know you should really know how important this is. This is not acceptable.

diplfranzhoepfinger:
> Why are you always changing the copyright headers and not keeping the original ones. By know you should really know how important this is. This is not acceptable.

totally Agree. 

Especially Johannes Messmer cannot be removed. He founded all this new-IO.

MoritzO02:
> Why are you always changing the copyright headers and not keeping the original ones. By know you should really know how important this is. This is not acceptable.

When there is a difference between the info for a person in the IDE and Forte do they impact each other or not? 
If there is an organization for a person in the IDE but not in Forte should both have the same info? Which one would then be the correct info?
If a Person is stated in the IDE but not in forte does he/she belong in forte?

azoitl:
> > Why are you always changing the copyright headers and not keeping the original ones. By know you should really know how important this is. This is not acceptable.
> 
> When there is a difference between the info for a person in the IDE and Forte do they impact each other or not? If there is an organization for a person in the IDE but not in Forte should both have the same info? Which one would then be the correct info? If a Person is stated in the IDE but not in forte does he/she belong in forte?

There can be several reasons for differences between of FB types between 4diac IDE copyright header and 4diac FORTE header:

1. there was a mistake: we need to resolve that and clean it
2. It is an SIFB and the SIFB implementation was only changed in the C++ source code as no interface change was needed. Then there can be different headers for both sides

In case of 1 it would definitely better to first raise an issue so that we can discuss and find the correct solution. Often a look into the git history can help.

PR #342, opened 2025-03-01
Use PARAMS input to specify ID for I/O mapping

oellith:
For I/O read and write function blocks the PARAMS input can now be used to specify the ID of the I/O that should be accessed.
If PARAMS ist empty the full qualified instance name of the I/O read or write function block is used like before. 
The ID is defined via the I/O configuration function blocks.

### DIFF
--- a/src/core/io/processinterface.cpp
+++ b/src/core/io/processinterface.cpp
@@ -71,7 +71,7 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
 // If PARAMS is not empty use the string specified here "as is" for the IO mapping to the configuration.
 // Otherwise use the full qualified instance name of our FB.
 std::string ProcessInterface::getId() {
-  if (PARAMS().length()) {  
+  if (!PARAMS().empty()) {  
     return static_cast<std::string>(PARAMS());
   }
   else {
--- a/src/core/io/processinterface.cpp
+++ b/src/core/io/processinterface.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2023 Johannes Messmer (admin@jomess.com), fortiss GmbH,
- *                          Johannes Kepler University Linz
+ * Copyright (c) 2016, 2025 Johannes Messmer (admin@jomess.com), fortiss GmbH,
+ *                          Johannes Kepler University Linz, Thomas Öllinger
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -56,13 +56,7 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
   deinitialise();
 
   // Register interface
-  if (PARAMS().length()) {  // If PARAMS is not empty we use the string specified here "as is" for the IO mapping to the configuration. Otherwise use the full qualified instance name of our FB.
-    if (!(mIsListening = IOMapper::getInstance().registerObserver(std::string(PARAMS().c_str()), this))) {
-      STATUS() = scmFailedToRegister;
-      return false;
-    }
-  }
-  else if(!(mIsListening = IOMapper::getInstance().registerObserver(getFullQualifiedApplicationInstanceName('.'), this))) {
+  if(!(mIsListening = IOMapper::getInstance().registerObserver(getId(), this))) {
     STATUS() = scmFailedToRegister;
     return false;
   }
@@ -72,6 +66,17 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
   }
 
   return mIsReady;
+}
+
+// If PARAMS is not empty use the string specified here "as is" for the IO mapping to the configuration.
+// Otherwise use the full qualified instance name of our FB.
+std::string ProcessInterface::getId() {
+  if (PARAMS().length()) {  
+    return PARAMS().getStorage();
+  }
+  else {
+    return getFullQualifiedApplicationInstanceName('.');
+  }
 }
 
 bool ProcessInterface::deinitialise() {
--- a/src/core/io/processinterface.cpp
+++ b/src/core/io/processinterface.cpp
@@ -72,7 +72,7 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
 // Otherwise use the full qualified instance name of our FB.
 std::string ProcessInterface::getId() {
   if (PARAMS().length()) {  
-    return PARAMS().getStorage();
+    return static_cast<std::string>(PARAMS());
   }
   else {
     return getFullQualifiedApplicationInstanceName('.');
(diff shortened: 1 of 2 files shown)


src/core/io/processinterface.cpp
```diff
@@ -55,7 +56,13 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
   deinitialise();
 
   // Register interface
-  if(!(mIsListening = IOMapper::getInstance().registerObserver(getFullQualifiedApplicationInstanceName('.'), this))) {
+  if (PARAMS().length()) {  // If PARAMS is not empty we use the string specified here "as is" for the IO mapping to the configuration. Otherwise use the full qualified instance name of our FB.
```
azoitl: it would be better to introduce a method `std::string getId()` and do there the PARAMS check. With this the code duplication that we have now here can be avoided and the code is better readable.

src/core/io/processinterface.cpp
```diff
@@ -55,7 +56,13 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
...
+    if (!(mIsListening = IOMapper::getInstance().registerObserver(std::string(PARAMS().c_str()), this))) {
```
azoitl: please use `getStorage()` instead this removes the need to convert the `std::string` to a `char *` and then create a new `std::string` from it.

azoitl (commented):
As you did copyright relevant changes to this file you also need to update the first line of the copyright header. Please change the second year to the year you performed the change, and add your copyright holder (you or the company you are working for) at the end of the list.

src/core/io/processinterface.cpp
diplfranzhoepfinger: this would be Cool. we could then "almost" make a Automatic Mappinng on ID. BUT then i need a "ANY" on Params, so i can insert there a UINT as Params, and then the Mapping would be fully Automatic. 

i propose to merge this commit, then i do another PR where we cange PARAMS to ANY, and then User can insert a STRING or a UINT. 

this would be very very cool and solve one of my Main Problems at least for 50%...

oellith: @diplfranzhoepfinger How to you use the ID? Why ist this of type UDINT in your case? Are the IDs auto generated or is this some form of input number?
Usually the I/O configuration function blocks define the available IDs and the I/O access function blocks reference this IDs.

diplfranzhoepfinger: yes, they are Auto-generated. ISO-Designer der Bucher Automation AG (https://www.bucherautomation.com/iso-designer/sw10133) does output a Header File with `#define` this we convert into a global Const. 

now we do a mapping like this: 

so we make String/UINT Pairs. 

if the IX Block would directly know the UINT, it could map it 1:1

the only Question is then still IE, as this can be a UP, DOWN, LONG, so there the Information is not enough.

image

azoitl: @diplfranzhoepfinger I have several issues with your comments. First of all I would really appreciate if you could not hijack unrelated topics. What you want is something completely different and also require major changes to 4diac FORTE. For such points you should open a dedicate issue where this can be discussed and potential solutions can be derived. This brings me to my second problem. I have with the information given strong objects against turning the PARAMS input into an ANY. This would make it much harder for people to use it also from implementing the blocks. Therefore please open an issue so we can discuss there and come up with a sensible solution for this use case.

src/core/io/processinterface.h
```diff
@@ -37,6 +40,7 @@ namespace forte {
         protected:
           bool initialise(bool paIsInput, CEventChainExecutionThread *const paECET);
           bool deinitialise();
+          std::string getId();
```
azoitl: any reason why this method is protected?

azoitl (changes requested):

src/core/io/processinterface.cpp
diplfranzhoepfinger: this is what i was writing above. 
pls complete this first, 
and then i will open another PR (after Testing) 
to go for that idea.

diplfranzhoepfinger: BTW: IEC 61499 defines PARAMS as ANY, and also provide a Solution to my Question for IE: use a Structure. 
again: pls go on with this, and i make a proposal later.

src/core/io/processinterface.h
oellith: Nearly all methods in this class are protected. What do you suggest?

src/core/io/processinterface.cpp
MartinMelikMerkumians: Even better, use the explicit std::string cast operator

src/core/io/processinterface.h
azoitl: The methods are protected because they are called by child classes.

If you are not sure the default should always be private to improve encapsulation.

oellith: You are right, at the moment no child class (I/O access FB) uses this method.
But my thoughts were: If we combine the reading of the Id into a reusable getId() method (which was a good suggestion), then it should also be available for the child classes. Since all information that geItd() uses is itself public or protected, it turned out for me that getId() is also protected.
Then it can be used by a future I/O FB to request e.g. additional information from the I/O subsystem, such as calibration values from a "database" etc.

azoitl: From my experience and all the books I've read on software architecture if you currently do not need it in subclasses keep it private. If that requirement changes we can still make it protected, or in the worst case public.

diplfranzhoepfinger:
@oellith why close ?

oellith:
> @oellith why close ?

I tried what happens when I press the Close button, because I have no clue how this all works here on github. I only worked with Azure DevOps.

MartinMelikMerkumians:
The Github workflow is to keep the pull request open until a maintainer merges or declines it (e.g a maintainer accepts the merge request or closes it).
If you close a pull request, that means that you dont want your PR to be merged anymore (e.g. in the discussion a better/alternative way to meet your requirements is found and you dont need your PR merged anymore)

azoitl:
@MartinMelikMerkumians  somehow your comment got lost. Do you have an idea how that happened? Can you give an example how to do the explict cast? Is it with `static_cast<std::string>(PRAMS)`?

MartinMelikMerkumians:
> @MartinMelikMerkumians somehow your comment got lost. Do you have an idea how that happened? Can you give an example how to do the explict cast? Is it with `static_cast<std::string>(PRAMS)`?

@azoitl yes exactly. `static_cast<std::string>(PARAMS())` should work. The `CIEC_STRING` class defines `explicit operator std::string() const {
      return getStorage();
    }`

So that should be the nicer way and future robust way to get a std string

src/core/io/processinterface.cpp
```diff
@@ -67,6 +68,17 @@ bool ProcessInterface::initialise(bool paIsInput, CEventChainExecutionThread *co
   return mIsReady;
 }
 
+// If PARAMS is not empty use the string specified here "as is" for the IO mapping to the configuration.
+// Otherwise use the full qualified instance name of our FB.
+std::string ProcessInterface::getId() {
+  if (PARAMS().length()) {
```
MartinMelikMerkumians: This if should use `!PARAMS().empty()` instead of `length()`

MartinMelikMerkumians (changes requested):

MartinMelikMerkumians (approved):

azoitl (approved):

PR #345, opened 2025-03-04
Add GEN_E_SPLIT function block

franz-hoepfinger-4diac:
- Introduced a new function block to split one event into multiple events.
- Updated the CMake configuration to include the new source files.
- Implemented necessary methods for event handling and interface specification.


azoitl (changes requested):
I don't think you should  have the generated part in the copyright header. First of all its contradicting the rest of copy right header. Secondly the file contains nearly no generated code so the copyright is only attributed to you.

franz-hoepfinger-4diac:
> I don't think you should have the generated part in the copyright header. First of all its contradicting the rest of copy right header. Secondly the file contains nearly no generated code so the copyright is only attributed to you.

I agree. will Change.

franz-hoepfinger-4diac:
done

azoitl (approved):

PR #348, opened 2025-03-06
Generated PLCnext fbt files from IDE to also use EInit

MoritzO02:
Generated PLCnextAXLSEDI16 from IDE to also use EInit

Generated PLCnextAXLSEDO16 from IDE to also use EInit

Generated PLCnextAXLSESC from IDE to also use EInit

Generated PLCnextBusAdapter from IDE to also use EInit

Generated PLCnextMaster from IDE to also use EInit

new header and old header have the same information

### DIFF
--- a/src/modules/PLCnext/types/PLCnextMaster.cpp
+++ b/src/modules/PLCnext/types/PLCnextMaster.cpp
@@ -1,20 +1,34 @@
-/*******************************************************************************
- * Copyright (c) 2022 Peirlberger Juergen
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * http://www.eclipse.org/legal/epl-2.0.
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *   Peirlberger Juergen - initial API and implementation and/or initial documentation
- *******************************************************************************/
+/*************************************************************************
+ *** Copyright (c) 2022 Peirlberger Juergen
+ ***
+ *** This program and the accompanying materials are made
+ *** available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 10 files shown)


src/modules/PLCnext/types/PLCnextMaster.cpp
```diff
@@ -56,9 +74,15 @@ FORTE_PLCnextMaster::FORTE_PLCnextMaster(const CStringDictionary::TStringId paIn
...
 
+bool FORTE_PLCnextMaster::initialize() {
+  if(!var_BusAdapterOut.initialize()) { return false; }
+  var_BusAdapterOut.setParentFB(this, 0);
+  return CFunctionBlock::initialize();
+}
+
 void FORTE_PLCnextMaster::setInitialValues() {
   var_QI = 0_BOOL;
-  var_SlaveUpdateInterval = 25_UINT;
```
azoitl, line 61: Something got lost here. Can it be that the default value 25 was removed from the type?

azoitl (changes requested):

src/modules/PLCnext/types/PLCnextMaster.cpp
MoritzO02: Yes the Initial Value in IDE is 0. Is the 25 that was in Forte still correct? Should I change it back to 25?

azoitl: I feel having a default value of 25 better then zero. could you please fix that in the IDE as well and then export. Because we now generate the 25 in two places.

MoritzO02: added the 25 in IDE exported it again. https://github.com/eclipse-4diac/4diac-ide/pull/1139

azoitl (approved):

PR #349, opened 2025-03-07
Generated zephyr fbt files from IDE to also use EInit

MoritzO02:
Generated Esp32EthernetKitIO from IDE to also use EInit

Generated ZephyrIO from IDE to also use EInit


MoritzO02:
Should the Organisation be added to the Person in this case, or is it not important?

azoitl (approved):

PR #354, opened 2025-03-10
Generated wagokbus fbt files from IDE to also use EInit

MoritzO02:
Generated WagoBusAdapter from IDE to also use EInit

Generated WagoMaster from IDE to also use EInit

remove Tabs

Keep current header

### DIFF
--- a/src/modules/wagokbus/types/WagoBusAdapter.cpp
+++ b/src/modules/wagokbus/types/WagoBusAdapter.cpp
@@ -10,7 +10,7 @@
 #include "WagoBusAdapter.h"
 
 USE_STRING_ID(BOOL);
-USE_STRING_ID(Event);
+USE_STRING_ID(EInit);
 USE_STRING_ID(Index);
 USE_STRING_ID(INIT);
 USE_STRING_ID(INITO);
@@ -30,11 +30,11 @@ const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmDataOutputTypeIds[] 
 const TDataIOID FORTE_WagoBusAdapter::scmEIWith[] = {0, scmWithListDelimiter};
 const TForteInt16 FORTE_WagoBusAdapter::scmEIWithIndexes[] = {0};
 const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmEventInputNames[] = {STRID(INITO)};
-const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmEventInputTypeIds[] = {STRID(Event)};
+const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmEventInputTypeIds[] = {STRID(EInit)};
 const TDataIOID FORTE_WagoBusAdapter::scmEOWith[] = {2, 1, 0, scmWithListDelimiter};
 const TForteInt16 FORTE_WagoBusAdapter::scmEOWithIndexes[] = {0};
 const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmEventOutputNames[] = {STRID(INIT)};
-const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmEventOutputTypeIds[] = {STRID(Event), STRID(Event)};
+const CStringDictionary::TStringId FORTE_WagoBusAdapter::scmEventOutputTypeIds[] = {STRID(EInit)};
 
 const SFBInterfaceSpec FORTE_WagoBusAdapter::scmFBInterfaceSpecSocket = {
   1, scmEventInputNames, scmEventInputTypeIds, scmEIWith, scmEIWithIndexes,
@@ -46,8 +46,8 @@ const SFBInterfaceSpec FORTE_WagoBusAdapter::scmFBInterfaceSpecSocket = {
 };
 
 const SFBInterfaceSpec FORTE_WagoBusAdapter::scmFBInterfaceSpecPlug = {
-  1, scmEventOutputNames, nullptr, scmEOWith, scmEOWithIndexes,
-  1, scmEventInputNames, nullptr, scmEIWith, scmEIWithIndexes,
+  1, scmEventOutputNames, scmEventOutputTypeIds, scmEOWith, scmEOWithIndexes,
+  1, scmEventInputNames, scmEventInputTypeIds, scmEIWith, scmEIWithIndexes,
   3, scmDataOutputNames, scmDataOutputTypeIds,
   1, scmDataInputNames, scmDataInputTypeIds,
   0, nullptr,
--- a/src/modules/wagokbus/types/WagoMaster.cpp
+++ b/src/modules/wagokbus/types/WagoMaster.cpp
@@ -11,6 +11,7 @@
 
 USE_STRING_ID(BOOL);
 USE_STRING_ID(BusAdapterOut);
+USE_STRING_ID(EInit);
 USE_STRING_ID(Event);
 USE_STRING_ID(IND);
 USE_STRING_ID(INIT);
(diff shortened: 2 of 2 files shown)


src/modules/wagokbus/types/WagoMaster.cpp
```diff
@@ -115,6 +133,13 @@ CIEC_ANY *FORTE_WagoMaster::getDO(const size_t paIndex) {
   return nullptr;
 }
 
+CAdapter *FORTE_WagoMaster::getAdapterUnchecked(const size_t paIndex) {
```
moniwe: adapter changes cause build errors, for modular IOs still the "old adapter implementation" is needed

src/modules/wagokbus/types/WagoMaster.h
```diff
@@ -40,6 +78,8 @@ class FORTE_WagoMaster : public forte::core::io::IOConfigFBMultiMaster {
 
     CIEC_ANY *getDI(size_t) override;
     CIEC_ANY *getDO(size_t) override;
+
+    CAdapter *getAdapterUnchecked(size_t) override;
```
moniwe: adapter changes cause build errors, for modular IOs still the "old adapter implementation" is needed

src/modules/wagokbus/types/WagoMaster.h
```diff
@@ -26,6 +62,8 @@ class FORTE_WagoMaster : public forte::core::io::IOConfigFBMultiMaster {
     CIEC_BOOL var_QO;
     CIEC_WSTRING var_STATUS;
 
+    FORTE_WagoBusAdapter var_BusAdapterOut;
```
moniwe: adapter changes cause build errors, for modular IOs still the "old adapter implementation" is needed

src/modules/wagokbus/types/WagoMaster.h
```diff
@@ -9,15 +9,51 @@
...
+    void executeEvent(TEventID paEIID, CEventChainExecutionThread *const paECET) override;
+
+    void readInputData(TEventID paEIID) override;
+    void writeOutputData(TEventID paEIID) override;
+    void setInitialValues() override;
+
   public:
     FORTE_WagoMaster(CStringDictionary::TStringId paInstanceNameId, forte::core::CFBContainer &paContainer);
-
+    bool initialize() override;
```
moniwe: adapter changes cause build errors, for modular IOs still the "old adapter implementation" is needed

src/modules/wagokbus/types/WagoMaster.h
```diff
@@ -9,15 +9,51 @@
...
+    static const TDataIOID scmEOWith[];
+    static const TForteInt16 scmEOWithIndexes[];
+    static const CStringDictionary::TStringId scmEventOutputNames[];
+    static const CStringDictionary::TStringId scmEventOutputTypeIds[];
+    static const int scmBusAdapterOutAdpNum = 0;
+    static const SAdapterInstanceDef scmAdapterInstances[];
+
+    static const SFBInterfaceSpec scmFBInterfaceSpec;
+
+    void executeEvent(TEventID paEIID, CEventChainExecutionThread *const paECET) override;
```
moniwe: I think there should not be an executeEvent method

src/modules/wagokbus/types/WagoMaster.cpp
```diff
@@ -53,9 +65,15 @@ FORTE_WagoMaster::FORTE_WagoMaster(const CStringDictionary::TStringId paInstance
     conn_STATUS(this, 1, &var_conn_STATUS) {
 };
 
+bool FORTE_WagoMaster::initialize() {
```
moniwe: adapter changes cause build errors, for modular IOs still the "old adapter implementation" is needed

src/modules/wagokbus/types/WagoBusAdapter.h
```diff
@@ -14,10 +14,20 @@
 #include "typelib.h"
 #include "forte_bool.h"
 #include "forte_uint.h"
+#include "iec61131_functions.h"
+#include "forte_array_common.h"
+#include "forte_array.h"
+#include "forte_array_fixed.h"
+#include "forte_array_variable.h"
```
moniwe: these additional includes are not really needed

src/modules/wagokbus/types/WagoBusAdapter.cpp
```diff
@@ -12,6 +12,12 @@
 #include "WagoBusAdapter_gen.cpp"
 #endif
 
+#include "iec61131_functions.h"
+#include "forte_array_common.h"
+#include "forte_array.h"
+#include "forte_array_fixed.h"
+#include "forte_array_variable.h"
```
moniwe: these additional includes are not really needed

src/modules/wagokbus/types/WagoMaster.cpp
diplfranzhoepfinger: why ? do we need to re-Generate the Adaptors too, so everything is fresh ? or some more work to be spent on this ?

src/modules/wagokbus/types/WagoBusAdapter.cpp
diplfranzhoepfinger: do we need to file an issue in IDE? , because this comes from the Generator ?

src/modules/wagokbus/types/WagoBusAdapter.h
diplfranzhoepfinger: do we need to file an issue in IDE? , because this comes from the Generator ?

moniwe:
I updated the Wago modular IOs to integrate an additional module [here](https://github.com/eclipse-4diac/4diac-forte/pull/357), maybe it is easier to update the EInit there?

src/modules/wagokbus/types/WagoMaster.cpp
azoitl: No the problem is that the base infrastructure in the base classes is currently relying on the older API. Changing that is more work that we may do later.

src/modules/wagokbus/types/WagoBusAdapter.h
azoitl: The generator has a hard time to know if these includes are needed so it is easier to always export them. It does not really harm. However  @moniwe  is write that for these files where the C++ code is heavily adjust it would be better to not add it. So all in all it would be better to really just update the event types and remove all other changes.

azoitl:
> I updated the Wago modular IOs to integrate an additional module [here](https://github.com/eclipse-4diac/4diac-forte/pull/357), maybe it is easier to update the EInit there?

@MoritzO02 @diplfranzhoepfinger  As @moniwe changes are crutial for getting Wago support working on latest 4diac FORTE. As these changes are larger I would merge @moniwe PR first.

src/modules/wagokbus/types/WagoBusAdapter.h
moniwe: I agree with @azoitl that the generator cannot know this, since these IO implementations are always very special, and therefore it is easier when the user should remove not needed stuff.

azoitl (changes requested):
Given @moniwe's comment I would prefer if you could just commit the 3 -4 lines where the event type is changed and keep the rest as it is. These are the lines with `scmEventInputTypeIds`, `scmEventOutPutTypeIds`, and their usage in the interface spec. 

For these strongly adjusted SIFBs this is most of the time the better solution.

src/modules/wagokbus/types/WagoMaster.h
MoritzO02: Should this whole private block be kept or is it also not needed?

azoitl:
For reviewing it would be much nicer if you would not use merge commits but  rebase your commits.  Furthermore for this commit I think it is even better to just start fresh from develop and redo the changes (just the points mentioned in my previous comment) and force push it on this one.

diplfranzhoepfinger:
YES, @MoritzO02 in my Opinion this way it is much better !

azoitl (approved):

azoitl:
Thx for providing this cleaned-up version.

PR #361, opened 2025-03-15
Add to DECLARE_HANDLER's getIdentifier the 'override' specifier

gravemalte:
The macro `DECLARE_HANDLER` should include the override specifier because `getIdentifier()` is virtual and every com handler will override the method due to the inheritance of the `CExternalEventHandler` class and the usage of the `DEFINE_HANDLER` macro.

### DIFF
--- a/src/core/extevhan.h
+++ b/src/core/extevhan.h
@@ -21,7 +21,7 @@ class CFunctionBlock;
 #define DECLARE_HANDLER(TypeName)                          \
   public:                                                  \
     static const size_t mHandlerIdentifier;                \
-    virtual size_t getIdentifier() const override;         \
+    size_t getIdentifier() const override;                 \
     explicit TypeName(CDeviceExecution& paDeviceExecution);\
     ~TypeName();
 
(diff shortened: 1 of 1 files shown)


src/core/extevhan.h
```diff
@@ -18,10 +18,10 @@
...
 class CFunctionBlock;
 
-#define DECLARE_HANDLER(TypeName)                             \
-  public:                                                     \
-    static const size_t mHandlerIdentifier;              \
-    virtual size_t getIdentifier() const;               \
+#define DECLARE_HANDLER(TypeName)                          \
+  public:                                                  \
+    static const size_t mHandlerIdentifier;                \
+    virtual size_t getIdentifier() const override;         \
```
azoitl: if you add the override you can remove the virtual. Which is cleaner.

azoitl (changes requested):

src/core/extevhan.h
gravemalte: Fixed.

azoitl (approved):

Issue #362, opened 2025-03-15
Crash on replace:


franz-hoepfinger-4diac:
Crash on replace:

```
Poolstate: PoolScanning           
Poolstate: PoolWaitTransfer       
PoolTransProgress: 0.00 
Poolstate: ETPBuffLoad            
I (316618) wifi:<ba-add>idx:2 (ifx:1, 76:01:2d:a5:5f:70), tid:0, ssn:30, winSize:64
Poolstate: ETP_start              
Poolstate: ETP_Trans_running      
PoolTransProgress: 100.00 
INFO: T#318000000000: Connection established by client
Poolstate: PoolScanning           
Poolstate: PoolWaitTransfer       
PoolTransProgress: 0.00 
Poolstate: ETPBuffLoad            
Guru Meditation Error: Core  1 panic'ed (InstrFetchProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x00000000  PS      : 0x00060030  A0      : 0x820df81b  A1      : 0x3fcdda90  
A2      : 0x3c2a1924  A3      : 0x3fce8f88  A4      : 0x00000021  A5      : 0x00060023  
A6      : 0x0000002d  A7      : 0x00000021  A8      : 0x821da9a8  A9      : 0x3fcdda50  
A10     : 0x3c2a1924  A11     : 0x00000000  A12     : 0x00000001  A13     : 0x00000000  
A14     : 0x3fcd63ec  A15     : 0x00000000  SAR     : 0x00000011  EXCCAUSE: 0x00000014  
EXCVADDR: 0x00000000  LBEG    : 0x400570e8  LEND    : 0x400570f3  LCOUNT  : 0x00000000  
--- 0x400570e8: __ssprint_r in ROM
0x400570f3: __ssprint_r in ROM



[...]

PR #363, opened 2025-03-16
Fix deregistration logic and add state change method

franz-hoepfinger-4diac:
- Set mIsListening to false when deregistering
- Implement changeExecutionState method for handling commands

Background: https://github.com/eclipse-4diac/4diac-forte/issues/362


franz-hoepfinger-4diac:
at least @MartinMelikMerkumians should say if this makes sense or nonsense.

azoitl (approved):

PR #365, opened 2025-03-16
Clean-up and improvement of I/O FBs for Modular IOs

azoitl:
This PR addresses several issues on how the IO FBs where used by modular IOs. As the modular IO interface gets more and more important a dedicate implementation is much better. 

As it is a rather large change I tried to split it in individual commits.

@franz-hoepfinger-4diac FYI


src/core/io/IB_fbt.cpp
```diff
@@ -0,0 +1,146 @@
...
+      break;
+    }
+    default:
+      CProcessInterfaceFB::writeOutputData(paEIID);
+      break;
+  }
+}
+
+CIEC_ANY *FORTE_IB::getDO(const size_t paIndex) {
+  if(paIndex == 2){
```
MartinMelikMerkumians: as size_t is an unsigned type, the literal should use a literal suffix marking it as unsigned, e.g. U, to avoid compiler warnings

src/core/io/IB_fbt.cpp
```diff
@@ -0,0 +1,146 @@
...
+    return &var_IN;
+  }
+  return CProcessInterfaceFB::getDO(paIndex);
+}
+
+CEventConnection *FORTE_IB::getEOConUnchecked(const TPortId paIndex) {
+  if(paIndex == 2) {
```
MartinMelikMerkumians: as size_t is an unsigned type, the literal should use a literal suffix marking it as unsigned, e.g. U, to avoid compiler warnings

src/core/io/IB_fbt.cpp
```diff
@@ -0,0 +1,146 @@
...
+    return &conn_IND;
+  }
+  return CProcessInterfaceFB::getEOConUnchecked(paIndex);
+}
+
+CDataConnection *FORTE_IB::getDOConUnchecked(const TPortId paIndex) {
+  if(paIndex == 2) {
```
MartinMelikMerkumians: as size_t is an unsigned type, the literal should use a literal suffix marking it as unsigned, e.g. U, to avoid compiler warnings

src/core/io/ID_fbt.cpp
```diff
@@ -0,0 +1,146 @@
...
+      break;
+    }
+    default:
+      CProcessInterfaceFB::writeOutputData(paEIID);
+      break;
+  }
+}
+
+CIEC_ANY *FORTE_ID::getDO(const size_t paIndex) {
+  if(paIndex == 2){
```
MartinMelikMerkumians: as size_t is an unsigned type, the literal should use a literal suffix marking it as unsigned, e.g. U, to avoid compiler warnings

src/core/io/IB_fbt.h
```diff
@@ -0,0 +1,100 @@
...
+    void evt_REQ(const CIEC_BOOL &paQI, const CIEC_STRING &paPARAMS, CIEC_BOOL &paQO, CIEC_STRING &paSTATUS, CIEC_BYTE &paIN) {
+      var_QI = paQI;
+      var_PARAMS = paPARAMS;
+      receiveInputEvent(scmEventREQID, nullptr);
+      paQO = var_QO;
+      paSTATUS = var_STATUS;
+      paIN = var_IN;
+    }
+
+    void operator()(const CIEC_BOOL &paQI, const CIEC_STRING &paPARAMS, CIEC_BOOL &paQO, CIEC_STRING &paSTATUS, CIEC_BYTE &paIN) {
```
MartinMelikMerkumians: does it make sense to actually call evt_INIT if this FB is used like a function?

I don't see a use case where this makes sense. I would rather think this should raise an error

src/core/io/device/io_controller.cpp
```diff
@@ -73,7 +73,7 @@ void IODeviceController::addHandle(HandleDescriptor &paHandleDescriptor) {
 }
 
 void IODeviceController::fireIndicationEvent(IOObserver* paObserver) {
-  startNewEventChain((CProcessInterface*) paObserver);
+  startNewEventChain((CProcessInterfaceFB*) paObserver);
```
MartinMelikMerkumians: can you please change this to a c++ style cast?

src/core/io/processinterfacefb.cpp
```diff
@@ -0,0 +1,253 @@
...
+  // Deregister interface
+  if(mIsListening) {
+    IOMapper::getInstance().deregisterObserver(this);
+    mIsListening = false; // we are deregistered ! 
+  }
+
+  return !mIsReady;
+}
+
+CIEC_BOOL CProcessInterfaceFB::read(CIEC_ANY &paData) {
```
MartinMelikMerkumians: I think paData could be const

src/core/io/processinterfacefb.cpp
```diff
@@ -0,0 +1,253 @@
...
+  curHandle->set(paData);
+
+  return true_BOOL;
+}
+
+bool CProcessInterfaceFB::onChange() {
+  return read();
+}
+
+void CProcessInterfaceFB::onHandle(IOHandle* paHandle) {
```
MartinMelikMerkumians: I think this could be IOHandle* const paHandle

src/core/io/processinterfacefb.cpp
```diff
@@ -0,0 +1,253 @@
...
+  EMGMResponse response = CFunctionBlock::changeExecutionState(paCommand);
+  if(response == EMGMResponse::Ready && (paCommand == EMGMCommandType::Kill || paCommand == EMGMCommandType::Reset)) {
+    deinitialise();
+  }
+  return response;
+}
+
+CIEC_ANY *CProcessInterfaceFB::getDI(const size_t paIndex) {
+  switch(paIndex) {
+    case 0: return &var_QI;
```
MartinMelikMerkumians, line 183: the case literal should be suffixed with e.g. U so that we have a unsigned to unsigned comparison

src/core/io/processinterfacefb.cpp
```diff
@@ -0,0 +1,253 @@
...
+  if(mDirection == IOMapper::In) {
+    setEventChainExecutor(paECET);
+  }
+
+  return mIsReady;
+}
+
+// If PARAMS is not empty use the string specified here "as is" for the IO mapping to the configuration.
+// Otherwise use the full qualified instance name of our FB.
+std::string CProcessInterfaceFB::getId() {
```
MartinMelikMerkumians: is it necessary that the returned string is modifyable? if no, this could be made const

MartinMelikMerkumians (changes requested):
I didnt mark any instance becuase the code is repititive and it would not be of added value if I copy my comment to any instance

src/core/io/IB_fbt.h
mx990: This is a good point. We should not generate a call-operator when there are multiple events, since it is then not allowed to call the FB without specifying the event in ST anyhow.

mx990: I have created a corresponding PR for the exporter: https://github.com/eclipse-4diac/4diac-ide/pull/1166

src/core/io/IB_fbt.cpp
azoitl: I don't get any warnings. I'm anyhow changing this again. Will be fixed in a later PR also for that below.

src/core/io/processinterfacefb.cpp
azoitl: This is a copy of code generated by our exporter. @mx990  should we fix this then there as well

src/core/io/processinterfacefb.cpp
mx990: You should never use const with return-by-value, see https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/#return-types-never-const for a good explanation why.

src/core/io/processinterfacefb.cpp
mx990: The [switch labels](https://en.cppreference.com/w/cpp/language/switch#Labels) are [implicitly converted](https://en.cppreference.com/w/cpp/language/constant_expression#Converted_constant_expression) to the type of the switch condition, so I would not expect any warnings or problems here.

src/core/io/processinterfacefb.cpp
azoitl: no because we read from the process interface to store it in paData.

src/core/io/device/io_controller.cpp
azoitl: done

azoitl:
@MartinMelikMerkumians I tried to fix the most pressing things. Furthermore I added one more commit with a template for inputs. I wanted to do the same for outputs. This should allow to further improve the processinterfacefb. What do you think?

src/core/io/processinterfacefb.cpp
MartinMelikMerkumians: Oh, my bad. Thats a read from to paData? How about the write method below then? That's then setting the writing the value from paData to the IO?

src/core/io/processinterfacefb.cpp
MartinMelikMerkumians: sorry, I expressed myself perhaps bad. 
i meant to change it to `std::string CProcessInterfaceFB::getId() const` and not to make the return type const

MartinMelikMerkumians (changes requested):

src/core/io/processinterfacefb.cpp
MartinMelikMerkumians: Thanks for clarifying this. I didnt know that switch are doing that automatically. thats nice to know

MartinMelikMerkumians:
> @MartinMelikMerkumians I tried to fix the most pressing things. Furthermore I added one more commit with a template for inputs. I wanted to do the same for outputs. This should allow to further improve the processinterfacefb. What do you think?

I think that makes perfect sense, as it gathers all the boilerplate into a template. You should add a `static_assert(std::is_base_of_v<CIEC_ANY_BIT, T>);` to limit instantiation to `CIEC_ANY_BIT` types (in case you did and I just overlooked it ignore this)

azoitl:
@MartinMelikMerkumians Github somhow does not let me nicely answer to your comments. I think I addressed all of them. Some anyhow went away with later clean-up. 

@franz-hoepfinger-4diac I think it is in a state that can be tested now.

src/core/io/mapper/io_observer.cpp
```diff
@@ -18,23 +18,17 @@
 
 using namespace forte::core::io;
 
-IOObserver::IOObserver() :
-    mHandle(nullptr), mType(CIEC_ANY::e_ANY), mDirection(IOMapper::UnknownDirection) {
-
+IOObserver::IOObserver() : mHandle(nullptr) {
```
cochicde, line 21: you can initialize this in the header and set the constructor to `= default`

src/core/io/outputfb.h
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Contributors:
+ *   Alois Zoitl - adds initial implementation
+ *************************************************************************/
+
+#pragma once
+
+#include "processinterfacefb.h"
+
+namespace forte {
```
cochicde: you could have `forte::core::io {` to remove some extra indentation (also in the other files)

src/core/io/processinterfacefb.cpp
```diff
@@ -0,0 +1,227 @@
...
+const CIEC_STRING CProcessInterfaceFB::scmFailedToRegister("Failed to register observer."_STRING);
+const CIEC_STRING CProcessInterfaceFB::scmMappedWrongDirectionOutput("Mapped invalid direction. A Q block requires an output handle."_STRING);
+const CIEC_STRING CProcessInterfaceFB::scmMappedWrongDirectionInput("Mapped invalid direction. An I block requires an input handle."_STRING);
+const CIEC_STRING CProcessInterfaceFB::scmMappedWrongDataType("Mapped invalid data type."_STRING);
+
+
+CProcessInterfaceFB::CProcessInterfaceFB(forte::core::CFBContainer &paContainer, const SFBInterfaceSpec& paInterfaceSpec, const CStringDictionary::TStringId paInstanceNameId) :
+    CEventSourceFB(paContainer, paInterfaceSpec, paInstanceNameId),
+    IOObserver(),
+    var_QI(0_BOOL),
```
cochicde, line 38: you could have some of these initializations in the header where they are declared

src/core/io/processinterfacefb.cpp
```diff
@@ -0,0 +1,227 @@
...
+    var_STATUS(""_STRING),
+    var_conn_QO(var_QO),
+    var_conn_STATUS(var_STATUS),
+    conn_INITO(this, 0),
+    conn_CNF(this, 1),
+    conn_QI(nullptr),
+    conn_PARAMS(nullptr),
+    conn_QO(this, 0, &var_conn_QO),
+    conn_STATUS(this, 1, &var_conn_STATUS) {
+  mIsListening = false;
```
cochicde, line 50: also these can be set in the header where they are declared

src/core/io/processinterfacefb.h
```diff
@@ -0,0 +1,114 @@
...
+ *                      interface elements
+ *******************************************************************************/
+
+#ifndef SRC_CORE_IO_PROCESSINTERFACEFB_H_
+#define SRC_CORE_IO_PROCESSINTERFACEFB_H_
+
+#include "mapper/io_mapper.h"
+#include "mapper/io_observer.h"
+
+#include <forte_sync.h>
```
cochicde: you can remove this include now

src/core/io/processinterfacefb.cpp
azoitl: I think this is more a style thing. AFAIK we currently do all the initialization in constructors and are not using default member initializers. I'm open to both.

@MartinMelikMerkumians  @mx990 what do you think?

src/core/io/outputfb.h
azoitl: my compiler says this does not work.  I've also only seen it in using namespace statements.

src/core/io/processinterfacefb.h
azoitl: done

src/core/io/outputfb.h
azoitl: done

moniwe:
I built it with 4diac-fbe for Revolution Pi and tested it with a small IX-QX example.

azoitl:
> I built it with 4diac-fbe for Revolution Pi and tested it with a small IX-QX example.

Thx for testing this helps a lot.

src/core/io/processinterfacefb.cpp
MartinMelikMerkumians: Quite sure we do it everywhere like you do it here. If we want to change this style, imho this should be a separate commit where we add that to the styling guide and fixing it everywhere.

src/core/io/outputfb.h
```diff
@@ -0,0 +1,183 @@
...
+
+        return true_BOOL;
+      }
+
+  };
+
+  template<class T>
+  const CStringDictionary::TStringId COutputFB<T>::scmDataInputNames[] = {STRID(QI), STRID(PARAMS), STRID(OUT)};
+  template<class T>
+  const CStringDictionary::TStringId COutputFB<T>::scmDataInputTypeIds[] = {STRID(BOOL), STRID(STRING), STRID(BYTE)};
```
mx990: I believe this should now be:
```suggestion
  const CStringDictionary::TStringId COutputFB<T>::scmDataInputTypeIds[] = {STRID(BOOL), STRID(STRING), forte::CDataTypeTrait<T>::scmDataTypeName};
```

src/core/io/processinterfacefb.cpp
azoitl: I also thought a bit about this and for most of these classes we only have one single constructor. For these cases I find it a bit better to have the initialization at the constructor as then all initialization is in one single place.

mx990: In general, I believe that both styles have merit, depending on the particluar circumstances. As for the concrete case, the exporter initializes FB variables in the constructor because they may also have complex initial values, including global constants and functions. This also avoids potential additional dependencies in the header file. Furthermore, the definition of the corresponding `setInitialValues()` is also here, so we keep everything in one place. As such, I believe we should keep FB initializers in the constructor.

tl;dr I would suggest to keep it this way for now, at least for FB variables that are otherwise generated.

MartinMelikMerkumians:
The CDataTraits template is in the namespace forte, if I saw it correct, while the numeric_limits template is in the forte::templates namespace. should this be streamlined?

mx990 (changes requested):

azoitl:
> The CDataTraits template is in the namespace forte, if I saw it correct, while the numeric_limits template is in the forte::templates namespace. should this be streamlined?

the numeric_limts template is more an internal template as it is the base class for the std::numeric_limits in the specific subclasses. I found forte:.templates also a bit to long for the data type traits. 

Because of this two reasons I put it into the forte namespace. But I'm open for others.

src/core/io/outputfb.h
azoitl: Thanks for noting my copy paste mistake. Fixed in the update.

mx990 (approved):

MartinMelikMerkumians:
> > The CDataTraits template is in the namespace forte, if I saw it correct, while the numeric_limits template is in the forte::templates namespace. should this be streamlined?
> 
> the numeric_limts template is more an internal template as it is the base class for the std::numeric_limits in the specific subclasses. I found forte:.templates also a bit to long for the data type traits.
> 
> Because of this two reasons I put it into the forte namespace. But I'm open for others.

I wouldn't see numeric_limits as an internal template, as the default numeric_limits also has several use cases, or much different than type traits. 
however I can live with the current split.
I just want to point out that scattering templates associated with the data type classes over several namespaces could be hard to remember.

azoitl:
> > > The CDataTraits template is in the namespace forte, if I saw it correct, while the numeric_limits template is in the forte::templates namespace. should this be streamlined?
> > 
> > 
> > the numeric_limts template is more an internal template as it is the base class for the std::numeric_limits in the specific subclasses. I found forte:.templates also a bit to long for the data type traits.
> > Because of this two reasons I put it into the forte namespace. But I'm open for others.
> 
> I wouldn't see numeric_limits as an internal template, as the default numeric_limits also has several use cases, or much different than type traits. however I can live with the current split. I just want to point out that scattering templates associated with the data type classes over several namespaces could be hard to remember.

Ok I think I didn't expressed my self correctly. Currently the `forte::templates::numeric_limits` template is only used as base class for `std::numeric_limits` template specialization for our data types. As it is not used anywhere else and I'm not sure if it has a use outside of this use-case I assumed it is more an "internal" template.

MartinMelikMerkumians:
> > > > The CDataTraits template is in the namespace forte, if I saw it correct, while the numeric_limits template is in the forte::templates namespace. should this be streamlined?
> > > 
> > > 
> > > the numeric_limts template is more an internal template as it is the base class for the std::numeric_limits in the specific subclasses. I found forte:.templates also a bit to long for the data type traits.
> > > Because of this two reasons I put it into the forte namespace. But I'm open for others.
> > 
> > 
> > I wouldn't see numeric_limits as an internal template, as the default numeric_limits also has several use cases, or much different than type traits. however I can live with the current split. I just want to point out that scattering templates associated with the data type classes over several namespaces could be hard to remember.
> 
> Ok I think I didn't expressed my self correctly. Currently the `forte::templates::numeric_limits` template is only used as base class for `std::numeric_limits` template specialization for our data types. As it is not used anywhere else and I'm not sure if it has a use outside of this use-case I assumed it is more an "internal" template.

No, that's exactly what I meant. I know we mainly use it internally, but as for numeric_limits, there could be use in system provided functions and similar stuff, like we use numeric_limits of the C++ types for such stuff.

As said, it's fine for me, so let's merge this

MartinMelikMerkumians (approved):

PR #371, opened 2025-03-21
Change creation of Struct Demuxer interface specification to support configured Structs

m-meingast:


azoitl:
It looks like that this commit breaks the test.

m-meingast:
@azoitl I fixed the issue with the failing test.

azoitl (approved):

PR #372, opened 2025-03-21
Set configured data port to reference of respective member

m-meingast:


src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -38,8 +39,8 @@ const CStringDictionary::TStringId GEN_STRUCT_DEMUX::scmDataInputNames[] = { STR
 
 void GEN_STRUCT_DEMUX::executeEvent(TEventID paEIID, CEventChainExecutionThread *const paECET) {
   if(scmEventREQID == paEIID) {
-    for (TPortId i = 0; i < st_IN().getStructSize(); i++){
-      getDO(i)->setValue(*st_IN().getMember(i));
+    for (TPortId i = 0; i < mConfiguredDOPorts.size(); i++){
```
azoitl: wouldn't it be better to use here copyStructValuesToOutputs

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -73,9 +75,9 @@ void GEN_STRUCT_DEMUX::setInitialValues() {
 }
 
 void GEN_STRUCT_DEMUX::copyStructValuesToOutputs() {
-  for (TPortId i = 0; i < st_IN().getStructSize(); i++) {
-    getDO(i)->setValue(st_IN().getMember(i)->unwrap());
-  }
+  for (TPortId i = 0; i < mConfiguredDOPorts.size(); i++){
+      getDO(i)->setValue(*mConfiguredDOPorts[i]);
```
azoitl: to be sure you should use unwrap here.

azoitl (changes requested):
Could you please move that PR to freeze branch.

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
m-meingast: That's true. Fixed it.

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
m-meingast: Done

azoitl (approved):
Please ignore my move it to freeze. I mixed up PRs.

PR #375, opened 2025-03-21
Rework CFB If2In and In2If connections

mx990:
Rework CFB interface-to-internal and internal-to-interface connections to directly sample between interface connections and internal FBs. Also remove connected check in write data as this would currently result in a stale (sampled) value in the connection if it became connected later.

### DIFF
--- a/src/core/cfb.cpp
+++ b/src/core/cfb.cpp
@@ -27,8 +27,6 @@
 CCompositeFB::CCompositeFB(forte::core::CFBContainer &paContainer, const SFBInterfaceSpec& paInterfaceSpec,
                            CStringDictionary::TStringId paInstanceNameId, const SCFB_FBNData & paFBNData) :
         CFunctionBlock(paContainer, paInterfaceSpec, paInstanceNameId),
-        mIf2InDConns(nullptr),
-        mIn2IfDConns(nullptr),
         cmFBNData(paFBNData),
         mEventConnections(nullptr),
         mDataConnections(nullptr),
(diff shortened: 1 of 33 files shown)


src/core/cfb.cpp
```diff
@@ -76,46 +74,21 @@ CCompositeFB::~CCompositeFB() {
...
 
 bool CCompositeFB::configureGenericDO(TPortId paDOPortId, const CIEC_ANY &paRefValue){
-  bool bRetVal = CFunctionBlock::configureGenericDO(paDOPortId, paRefValue);
-  CDataConnection *dataCon = mIn2IfDConns[paDOPortId];
-  if(bRetVal && dataCon != nullptr){
-    //issue a reconfiguration attempt so that all connection end points in this connection are also correctly configured
-    dataCon->handleAnySrcPortConnection(paRefValue);
-  }
-  return bRetVal;
+  // TODO remove?
```
azoitl, line 90: no we still have to call handlyAnySrcPortconnection to configure into the cfb.

src/core/cfb.cpp
```diff
@@ -76,46 +74,21 @@ CCompositeFB::~CCompositeFB() {
...
-    if (nullptr != mIn2IfDConns) {
-      delete[] mIn2IfDConns;
-    }
   }
 }
 
 bool CCompositeFB::connectDI(TPortId paDIPortId, CDataConnection *paDataCon){
-  bool retVal = false;
-
+  // TODO remove?
```
azoitl, line 81: if can remove the connectDI calls from the internal to interface connection creation as we don't need it anymore we could get rid of this method.

azoitl (approved):

PR #376, opened 2025-03-22
Reworked FB Test Fixture to not use CInterface2InternalDataConnection

azoitl:
As preperation for getting rid of the mValue member of DataConnection and simplifying inheriting from CDataConnection mValue access in DataConnection was reworked.

### DIFF
--- a/src/core/dataconn.h
+++ b/src/core/dataconn.h
@@ -74,20 +74,20 @@ class CDataConnection : public CConnection {
     *   Get class member variable mValue.
     *   \return pointer to class member variable mValue
     */
-    CIEC_ANY* getValue() {
-      return mValue;
+    virtual CIEC_ANY& getValue() {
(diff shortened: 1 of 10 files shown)


src/core/dataconn.h
```diff
@@ -109,6 +109,10 @@ class COutDataConnection final : public CDataConnection {
       paValue.setValue(mValue.unwrap());
     }
 
+    CIEC_ANY& getValue() override {
```
mx990: Please return a `T&` here to take advantage of the known type.

mx990 (changes requested):

src/core/dataconn.h
azoitl: thx for noting. Fixed.

mx990 (approved):

PR #378, opened 2025-03-23
Reworked connection lists in CFBs

azoitl:
With these two commits the connection creation and management of CFBs is reworked.


src/core/cfb.cpp
```diff
@@ -118,58 +94,43 @@ CIEC_ANY *CCompositeFB::getVar(CStringDictionary::TStringId *paNameList, unsigne
...
 
-void CCompositeFB::createEventConnections(){
-  prepareIf2InEventCons();  //the interface to internal event connections are needed even if they are not connected therefore we have to create them correctly in any case
-  if(0 != cmFBNData.mNumEventConnections){
-    mEventConnections = new CEventConnection *[cmFBNData.mNumEventConnections]; //TODO for a major revison this list could be ommited but requires a change in the faned out connections
-
-    for(size_t i = 0; i < cmFBNData.mNumEventConnections; ++i){
+void CCompositeFB::createEventConnections() {
+  prepareIf2InEventCons(); //the interface to internal event connections are needed even if they are not connected therefore we have to create them correctly in any case
+  if(0 != cmFBNData.mNumEventConnections) {
```
mx990: I believe this check is now redundant, since the following `for` loop would not be entered anyhow.

src/core/cfb.cpp
```diff
@@ -118,58 +94,43 @@ CIEC_ANY *CCompositeFB::getVar(CStringDictionary::TStringId *paNameList, unsigne
...
-      establishConnection(mEventConnections[currentFannedConn->mConnectionNum], dstFB, currentFannedConn->mDstId);
-    }
   }
 }
 
 void CCompositeFB::prepareIf2InEventCons(){
-  if(0 != getFBInterfaceSpec().mNumEIs){
-    mInterface2InternalEventCons = new TEventConnectionPtr[getFBInterfaceSpec().mNumEIs];
-    //FIXME find a way to avoid that each connection has to be allocated separately
+  if(getFBInterfaceSpec().mNumEIs != 0){
```
mx990: I believe this check is now redundant, since reserving `0` elements would not matter and the following `for` loop would also not be entered anyhow.

src/core/cfb.cpp
```diff
@@ -185,29 +146,19 @@ void CCompositeFB::establishConnection(CConnection *paCon, CFunctionBlock *paDst
 }
 
 void CCompositeFB::createDataConnections(){
-  if(cmFBNData.mNumDataConnections){
-    mDataConnections = new CDataConnection *[cmFBNData.mNumDataConnections];
-
-    for(size_t i = 0; i < cmFBNData.mNumDataConnections; ++i){
+  if(cmFBNData.mNumDataConnections) {
```
mx990: I believe this check is now redundant, since the following `for` loop would not be entered anyhow.

mx990 (changes requested):

src/core/cfb.cpp
azoitl: I was not sure about the reserving but I get what you mean here. will change.

src/core/cfb.cpp
azoitl: done

src/core/cfb.cpp
azoitl: done

mx990 (approved):

PR #381, opened 2025-03-24
Reworked Query Connection: Data Connections

azoitl:
For building the query connection response string for data connections the input side of FBs is used. This avoids the need for storing destination information into the connections.

### DIFF
--- a/src/core/resource.cpp
+++ b/src/core/resource.cpp
@@ -366,7 +366,7 @@ EMGMResponse CResource::queryFBs(std::string &paValue, const CFBContainer &conta
       const CFunctionBlock &fb = static_cast<const CFunctionBlock &>(*itRunner);
       createFBResponseMessage(fb, prefix + fb.getInstanceName(), paValue);
     } else {
-      queryFBs(paValue, *itRunner, prefix + itRunner->getInstanceName() + ".");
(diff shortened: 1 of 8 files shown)


src/core/resource.cpp
```diff
@@ -417,65 +413,38 @@ void CResource::createAOConnectionResponse(const CFunctionBlock& paFb, std::stri
...
-  paReqResult.append(fullName);
+  paValue += "<FB name=\""s;
+  paValue += fullName;
+  paValue += "\" type=\""s;
+  paValue += paFb.getFBTypeName();
+  paValue += "\"/>\n"s;
+}
+
+void CResource::createConnectionResponseMessage(const CFunctionBlock& paDstFb, const CStringDictionary::TStringId srcId,
+    const CFunctionBlock& paSrcFb, const CStringDictionary::TStringId dstId, std::string& paReqResult) {
```
mx990: I looks to me like the order of `paDstFb` and  `paSrcFb` should be flipped? Also, I believe `srcId` and `dstId` should be renamed to `paSrcId` and `paDstId`, respectively.

mx990 (changes requested):

src/core/resource.cpp
azoitl: thx for noting. fixed

mx990 (approved):

PR #383, opened 2025-03-24
Reworked Connect and Disconnect of Adapter Connections

azoitl:
As adapter connections are point to point connections no information on destionation lists are needed.

Furthermore as more fields and methods are moved to references.


mx990 (approved):

MartinMelikMerkumians (approved):
LGTM

PR #385, opened 2025-03-25
Remove value in data connection base class

mx990:
Remove value in data connection base class and add generic data connection for generic FBs.


azoitl (changes requested):
Could you please remove the `;` that is after the writeData method. I forgot to remove it when I did my data connection clean.

Please also rebase the PR.

azoitl (approved):

PR #386, opened 2025-03-25
Replay algorithm and device

cochicde:
This PR is based on a current research being done about reproducing all events triggered in a resource based on only the output events of Service Function Blocks. 
The idea here is that we trace all events, and then, only with the subset of output events of SIFB, generate the rest of the traces. For that, we need a counter for each resource which increases on each event being triggered. The counter is also traced.

The PR is divided in 4 main parts/commits:
1. Add the counter of events and trace the output events together with their data (tracing was previously done separately between data and events). 
2. Add resource and device replayers: Base on the counter and the output events of SIFB, these classes implement the algorithm to generate the rest of the traces
3. A Replay Device is added which leverages the previous classes and offers the replay functionality as OPC UA methods to the outside of the device
4. Tests are modified and added for the specific new features 

P.S.: There's no rush to merge this, so I can wait for the feature branch to be ready first.


src/core/funcbloc.h
```diff
@@ -298,9 +298,7 @@ class CFunctionBlock : public forte::core::CFBContainer {
     void receiveInputEvent(TEventID paEIID, CEventChainExecutionThread *paExecEnv) {
       FORTE_TRACE("InputEvent: Function Block (%s) got event: %d (maxid: %d)\n", CStringDictionary::get(getInstanceNameId()), paEIID, getFBInterfaceSpec().mNumEIs - 1);
 
-      #ifdef FORTE_TRACE_CTF
-        traceInputEvent(paEIID);
-      #endif
+
```
azoitl: Moving this down changes the order of generated traces breaking trace analysis of existing users. Can you please restore that again.

src/core/funcbloc.cpp
```diff
@@ -636,11 +636,30 @@ void CFunctionBlock::traceInputEvent(TEventID paEIID){
...
-void CFunctionBlock::traceOutputEvent(TEventID paEOID){
+void CFunctionBlock::traceOutputEvent(TEventID paEOID, CEventChainExecutionThread * const paECET){
   if(auto& tracer = getResource()->getTracer(); tracer.isEnabled()){
+ 
+    // don't trace events that are not connected since they don't increase the event counter
+    if(getEOConUnchecked(static_cast<TPortId>(paEOID))->getDestinationList().size() == 0) {
+      return;
+    }
+
+    std::vector<std::string> outputs(getFBInterfaceSpec().mNumDOs);
```
azoitl: with your change data outputs are now logged twice. is your change needed? can't you use the existing data output logging?

src/core/funcbloc.h
```diff
@@ -468,12 +470,11 @@ class CFunctionBlock : public forte::core::CFBContainer {
     void sendOutputEvent(TEventID paEO, CEventChainExecutionThread * const paECET){
       FORTE_TRACE("OutputEvent: Function Block sending event: %d (maxid: %d)\n", paEO, getFBInterfaceSpec().mNumEOs - 1);
 
-      #ifdef FORTE_TRACE_CTF
-        traceOutputEvent(paEO);
-      #endif
-
       if(paEO < getFBInterfaceSpec().mNumEOs) {
         writeOutputData(paEO);
+#ifdef FORTE_TRACE_CTF
```
azoitl: tracing the output event after write Outputdata is not changing the output of your event trace as you are tracing the values of the FB ports and not the connections. However moving that down changes the order of the generated traces for other users. Therefore I would kindly asking to move the output event tracing back where it was.

azoitl (changes requested):

src/core/trace/config.yaml
ernstblechaPT, line 1: Right now the dependencies to regenerate the cpp and h files from the config.yaml of barectf are not available in FBE - trying to build with trace ctf will result in a failing build. In order to work the build toolchain would need to include the barectf python package; alternativly generating the config file is done by the developer and the generated files are stored under version control.

Do you have any plan to tackle this?

cochicde: you're right! The good thing about barectf is that the generated files are self-contained and not extra libraries are needed. I think it makes more sense to have the generated file also commited in the repo. what do think @azoitl ?

If we go for this approach, I'd however open a new issue and have the changes separated from this one as to keep the intent of this focussed on only one thing

azoitl: I find this a good idea as it makes compiling 4diac FORTE for our users simpler. Especially as these files will rarely change. We follow a similar strategy with the Java files generated from the EMF models in 4diac IDE.

Also doing this in a separate issue/PR is good.

src/core/funcbloc.h
cochicde: I reverted it back

src/core/funcbloc.h
cochicde: I reverted it back

src/core/funcbloc.cpp
cochicde: I separated the two tracing methods: one for the classic one, and a new one for replay debugging. The replay debugging one changes the interface of the `sendOutputEvent` so I added a CMake variable and a `define` to separate both properly. For now, the replay debugging one still traces as much as the regular one,  but eventually the I will remove the rest of the traces for that configuration.

I had to move some files around to leave it cleaner, but basically I added some folders for it. The `internal` and `reader` folders are used only for testing and for the replay debugging device.

src/core/trace/config.yaml
ernstblechaPT: Thank you! I also think that this should be done in a seperate issue/PR. Up to now the files have not changed since October 2023 - So I agree with the assumption that those files will not change regularly.

azoitl (approved):

Issue #379, opened 2025-03-24
[CI] Formatting check


cochicde:
We need a formatting check added to the CI to have the code always consistent. We have the `[fortestyle.xml](https://github.com/eclipse-4diac/4diac-forte/blob/release/fortestyle.xml)` which are derived from what's specified here [forte_coding_rules.tex](https://github.com/eclipse-4diac/4diac-forte/blob/release/doc/coding_rules/src/forte_coding_rules.tex).

I'm not sure if we can incorporate a formatting check based on `fortestyle.xml`, and if not, we could add the same rules in clang which I know is easier to incorporate to the CI.

Any suggestions/ideas are welcomed

mx990:
I would prefer `.clang-format` in general, since that also works with other IDEs, such as CLion or Visual Studio (Code). It seems this is also used by Eclipse CDT itself, see https://github.com/eclipse-cdt/cdt/blob/main/.clang-format.

azoitl:
In general I'm open to anything. But I didn't find an option to use `.clang-format` in CDT. Maybe I have to check my setup.

MartinMelikMerkumians:
> In general I'm open to anything. But I didn't find an option to use `.clang-format` in CDT. Maybe I have to check my setup.

I think you need a plugin for that https://marketplace.eclipse.org/free-tagging/clang-format

azoitl:
@MartinMelikMerkumians I will also recheck the latest CDT 12.0 and install it from scratch. There is now a clangd integrated. I guess for that I can use a clang format file.

azoitl:
I've tested with ned CDT 12.0 and its LSP based setup. There I can us a clang format file. So I'm all in. How do we get such a file?

mx990:
I will create a `.clang-format` file based on the coding rules.

azoitl:
> I will create a `.clang-format` file based on the coding rules.

Thanks a lot. Not sure if I'm opening a pandoras box but if we have rules that are by now outdated or do not make to much sense anymore I would consider changing them. @MartinMelikMerkumians  what is your opinion here?

mx990:
I have created a PR https://github.com/eclipse-4diac/4diac-forte/pull/389 with a `.clang-format` file based on the coding rules and what I would consider sane defaults.

gravemalte:
I think we can close this issue. DO you agree @cochicde and @mx990?

PR #389, opened 2025-03-26
Replace Eclipse code style file with .clang-format

mx990:
Replace the Eclipse code style file with .clang-format and update the coding rules.

Contributes to https://github.com/eclipse-4diac/4diac-forte/issues/379


gravemalte:
Maybe some additions.

I'm a big fan of adding comments to the specific clangd format option, to quickly look it up. Also, I think the following options would be beneficial as well:

```.clang-format
# https://clang.llvm.org/docs/ClangFormatStyleOptions.html#derivepointeralignment
DerivePointerAlignment: false

# https://clang.llvm.org/docs/ClangFormatStyleOptions.html#pointeralignment
PointerAlignment: Left

# https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignarrayofstructures
AlignArrayOfStructures: Left
```

mx990:
I have added the suggested comments to the `.clang-format` file. As for the additional options:
- `DerivePointerAlignment: false`
  This is already `false` by default in the LLVM base style, see also `clang-format -style=llvm -dump-config`.
- `PointerAlignment: Left`
  I have personally become used to `*` and `&` aligned to the *right* instead (e.g., `int *p;` and `int &r;`), but I would be interested to hear the opinion of @azoitl and @MartinMelikMerkumians.
- `AlignArrayOfStructures: Left`
  The downside of aligning code or comments in columns is that this can produce large diffs when columns are re-aligned by the formatter due to editing a single line. I would therefore suggest to keep this as `None`.

MartinMelikMerkumians:
Derived Pointer alignments set to false seems to be a mindful setting. 

I am used to right alignments, and it has the benefit that it is consistent with order how the compiler applies modifiers, especially those `const *const`  things 

On the alignment point, I see the potential benefit, and I used it on OpENer, but this is also why I also see the issues @mx990 is bringing up on that point. Additionally, we would need to ensure that this is consistent with the code exporter in all cases, which adds some extra complexity. 
As a last comment, thinking about our test code base, which probably has the highest amount of such aligned code, does this work well with all the macros used in boost.test?
Overall I am not sure if it's worth it, and would probably go for `None` now too. 

Thinking of that, is there the possibility to use the formatter as part of the export process?

gravemalte:
> Thinking of that, is there the possibility to use the formatter as part of the export process?

The simplest and also the dirtiest solution which come to my mind is to spawn a `clang-format` process on the exported  code before (and then piping it to a file) or after writing it to a file. But this would require that `clang-format` is present in the system path (which is highly unlikely, at least on Windows).

Maybe @azoitl has a much better idea.

cochicde:
I just ran clang-format with the provided style to see how it looks. Here are some small examples:

```
CEventConnection *CFunctionBlock::getEOConnection(CStringDictionary::TStringId paEONameId) {
  CEventConnection *retVal = nullptr;
  TPortId portId = getPortId(paEONameId, getFBInterfaceSpec().mNumEOs, getFBInterfaceSpec().mEONames);
  if (cgInvalidPortId != portId) {
    retVal = getEOConUnchecked(portId);
  }
  return retVal;
}

const CEventConnection *CFunctionBlock::getEOConnection(CStringDictionary::TStringId paEONameId) const {
  const CEventConnection *retVal = nullptr;
  TPortId portId = getPortId(paEONameId, getFBInterfaceSpec().mNumEOs, getFBInterfaceSpec().mEONames);
  if (cgInvalidPortId != portId) {
    retVal = const_cast<CFunctionBlock *>(this)->getEOConUnchecked(portId);
  }
  return retVal;
}
```

The main thing that's different are the spaces, for example between `if` and `(`, beween the function parameters and the opening `{`. I would expect 
`if(cgInvalidPortId != portId){` instead of 
`if (cgInvalidPortId != portId) {` or maybe:
`if(cgInvalidPortId != portId) {` (I'm not sure anymore, but just putting it here to decide on what makes everyone happier)

Regarding the pointer on the left or right, I don't mind, but having it on the right, leads to some things looking kind of weird to me, for example:
casting looks like this:
 `retVal = const_cast<CFunctionBlock *>(this)->getEOConUnchecked(portId);` instead of 
`retVal = const_cast<CFunctionBlock*>(this)->getEOConUnchecked(portId);`

A method definition without a variable name, it looks like this:
`CBaseCommFB(const CBaseCommFB &) = delete;` instead of
`CBaseCommFB(const CBaseCommFB&) = delete;`

or when returning a reference it looks like
`CTimerHandler &CFunctionBlock::getTimer() {` instead of 
`CTimerHandler& CFunctionBlock::getTimer() {`



Regarding the exporter, I don't think we should not include it in the process. As I see the usual workflow is to export from the IDE, and then submit it to this repo. Applying the formating would be a responsability of the person comitting the new function blocks, which shouldn't be too much trouble.

gravemalte:
I agree with you regarding the pointers to the left. But I think is really a matter of personal preferences.


> Regarding the exporter, I don't think we should not include it in the process. As I see the usual workflow is to export from the IDE, and then submit it to this repo. Applying the formating would be a responsability of the person comitting the new function blocks, which shouldn't be too much trouble.

That’s a valid point. In theory, we could run a pipeline job for exported artifacts within the repository to ensure they are properly formatted. However, if we prefer the pipeline not to modify any files (which I personally lean toward), we can require contributors to format their files according to the ` .clang-format` rules, with the pipeline simply verifying compliance.

cochicde:
> That’s a valid point. In theory, we could run a pipeline job for exported artifacts within the repository to ensure they are properly formatted. However, if we prefer the pipeline not to modify any files (which I personally lean toward), we can require contributors to format their files according to the ` .clang-format` rules, with the pipeline simply verifying compliance.

exactly. That is actually the issue being tackled here. Once the style file is submitted, an action checking files in the PR for formatting needs to be added. At some point, when nobody is working on big PRs and to avoid conflicts, a complete formatting can be done to all files.

gravemalte:
> exactly. That is actually the issue being tackled here. Once the style file is submitted, an action checking files in the PR for formatting needs to be added. At some point, when nobody is working on big PRs and to avoid conflicts, a complete formatting can be done to all files.

`clang-format` features a `--dry-run` flag to not touch the orignal file (https://clang.llvm.org/docs/ClangFormat.html). An example setup would be:
 `clang-format <<my_file.cxx>> --dry-run --Werror`

That way, `clang-format` will output to stderr if something is odd in the formatting. The style is automatically picked up from the `.clang-format` file. If everything is alright, 0 is returned to stdout

azoitl:
Sorry for my late response. To add also here my opinions:

Regarding pointer and reference placement: Both options have their merit and therefore it would only be a matter of taste. Up till now 4diac FORTE used the right style and I would like to keep it like that. 

Regarding formatting of generated code: This is generated code based on xtend templates. We have full control how this code shall look like. So I don't think it is a smart idea to come up with some additional formaters and checkers when we can simply add the right format to your exporter. Especially as the generated code should already pretty well follow our style. Furthermore it is generated code. Such code is most of the time not intended to be read or changed by humans.

MartinMelikMerkumians:
> Sorry for my late response. To add also here my opinions:
> 
> Regarding pointer and reference placement: Both options have their merit and therefore it would only be a matter of taste. Up till now 4diac FORTE used the right style and I would like to keep it like that.
> 
> Regarding formatting of generated code: This is generated code based on xtend templates. We have full control how this code shall look like. So I don't think it is a smart idea to come up with some additional formaters and checkers when we can simply add the right format to your exporter. Especially as the generated code should already pretty well follow our style. Furthermore it is generated code. Such code is most of the time not intended to be read or changed by humans.

My comment on that was less to use two code formatters, but that we need to think about this if the code is checked in the pipeline against the clang-format, it needs to match, which would mean we need to autoformat it the same way as well.
With options like `AlignArrayOfStructures: Left` that will probably be a PitA to deal within xtend templates (or Java Template Strings), which I would like to keep such options disabled.

cochicde:
> My comment on that was less to use two code formatters, but that we need to think about this if the code is checked in the pipeline against the clang-format, it needs to match, which would mean we need to autoformat it the same way as well. 

as there no automatic exporter submitting to this repo, the problem of generated code being flagged as not formatted in the CI will get the person trying to submit the exported function blocks. As every person submitting will have anyways a formatter installed, I don't see any issue with this.

gravemalte:
> > My comment on that was less to use two code formatters, but that we need to think about this if the code is checked in the pipeline against the clang-format, it needs to match, which would mean we need to autoformat it the same way as well.
> 
> as there no automatic exporter submitting to this repo, the problem of generated code being flagged as not formatted in the CI will get the person trying to submit the exported function blocks. As every person submitting will have anyways a formatter installed, I don't see any issue with this.
I agree.

We could also just state in the README under contribution, that you should set up a git hook so that on commit `clang-format` will be automatically format your check-in files. Then a pipeline will just check if there are any formatting errors (which should not occur because of the git hook). And we do not need to touch any component of the code generation of the IDE.

MartinMelikMerkumians:
Our own code exporter not sticking to our own coding rules still strikes me as weird, and actually a lot of the currently included standard FBs have been generated with the IDE's code exporter.

azoitl:
For testing purposes I added the `clang.format` file to my project and enabled auto formating on save.  You can see the result as an example in this commit: https://github.com/eclipse-4diac/4diac-forte/pull/393/commits/203151deb1d88c9c8e6e3dbf4477002a7fabbb3b

It is still a bit noise but most of it I already like. Where I'm not sure is how function parameters are broken into new lines.

azoitl:
> Our own code exporter not sticking to our own coding rules still strikes me as weird, and actually a lot of the currently included standard FBs have been generated with the IDE's code exporter.

I also would find that extremely akquward.

cochicde:
> It is still a bit noise but most of it I already like. Where I'm not sure is how function parameters are broken into new lines.
 
It seems that it breaks only when it reaches the maximum column length of 120, otherwise it keeps it in the same line

`EMGMResponse CFBContainer::deleteFB(NameIterator &paNameListIt, NameIterator paNameListEnd) {`
 
For me it looks fine as it is.

azoitl:
After using it now for a bit of 4diac FORTE development I find it fine as it is and would merge it for now. If special cases are yet undetected we can fine tune it in further PRs, or did I miss anything important?

PR #391, opened 2025-03-26
Reworked Connection Code to Use FB refrences for Connection Endpoints

azoitl:
In order to reduce the need for null checks and make the code saver ConnectionPoint now only accepts FB references and provides FB references.


src/core/adapter.cpp
```diff
@@ -21,19 +21,20 @@ USE_STRING_ID(ANY_ADAPTER);
 #include "ecet.h"
 
 CAdapter::CAdapter(forte::core::CFBContainer &paContainer, const SFBInterfaceSpec& paInterfaceSpecSocket, const CStringDictionary::TStringId paInstanceNameId, const SFBInterfaceSpec& paInterfaceSpecPlug, bool paIsPlug) :
-  CGenFunctionBlock<CFunctionBlock>(paContainer, (paIsPlug) ? paInterfaceSpecPlug : paInterfaceSpecSocket, paInstanceNameId),
-  mParentAdapterListEventID(0),
-  mIsPlug(paIsPlug),
-  mPeer(nullptr),
-  mLocalDIs(mDIs),
-  mAdapterConn(nullptr){
+    CGenFunctionBlock<CFunctionBlock>(paContainer, (paIsPlug) ? paInterfaceSpecPlug : paInterfaceSpecSocket, paInstanceNameId),
```
MartinMelikMerkumians: I would suggest to remove the parantheses around paIsPlug, as it makes the code harder to read

MartinMelikMerkumians (changes requested):

src/core/adapter.cpp
azoitl: done

src/core/conn.h
```diff
@@ -43,6 +40,22 @@ class CConnectionPoint {
...
+    CFunctionBlock &getFB() const {
+      return *mFB;
+    }
+    
+    TPortId getPortId() const {
+      return mPortId;
+    }
+    
+  private:    
+    CFunctionBlock *mFB;
```
cochicde, line 57: can't this be a reference?

src/core/conn.h
```diff
@@ -43,6 +40,22 @@ class CConnectionPoint {
     bool operator!=(const CConnectionPoint & paRight) const {
       return !(*this == paRight);
     }
+    
+    CFunctionBlock &getFB() {
+      return *mFB;
+    }
+    
+    CFunctionBlock &getFB() const {
```
cochicde: shouldn't this return a const&?

src/core/conn.h
```diff
@@ -27,11 +27,8 @@ class CFunctionBlock;
...
-    CFunctionBlock *mFB;
-    TPortId mPortId;
-
-    CConnectionPoint(CFunctionBlock *paFB, TPortId paPortId) :
-      mFB(paFB), mPortId(paPortId) {
+    CConnectionPoint(CFunctionBlock &paFB, const TPortId paPortId) :
+      mFB(&paFB), mPortId(paPortId) {
     }
 
     CConnectionPoint() = default;
```
cochicde, line 34: is this constructor needed? since now the members cannot be set without the other constructor

src/core/conn.h
azoitl: Unfortunately no. This struct is used in vectors and lists and only if this is a pointer asignment and moving is possible.

src/core/conn.h
azoitl: No, we need to be able to send input events to the FB returned by this method or disconnect connections.

src/core/conn.h
azoitl: This constructor is needed by the ringbuffer of the event chain execution thread. Only that should be able to create uninitialized instances. I haven't yet found a way to better express that in code

src/core/conn.h
cochicde: but if the non-const method is needed, the above one will be used, right?

src/core/conn.h
cochicde: could we add assign and move operators then to this class to overcome that?

src/core/conn.h
cochicde: the ringbuffer only use the assign operator to store the data in the buffer, so if we had the assign operator for this class it should work (this analysis is done only from checking the code, didn't try compiling it yet without the constructor)

cochicde: ok, I tried doing it locally and I see the problem now regarding `std::array`. Forget about this comment

azoitl: I was offline and about to write that I tried a few things. But the problem is that references are not assignable.  STL has the `std::reference_wrapper` for this task but in the end it is doing the same thing that we are doing here. And as this is in addition a very performance sensitive class I don't think it is worth the effort.

src/core/conn.h
azoitl: Only if the SConnectionPoint is non const. We have one use case where the SConnectionPoint should not change but we need to update the contained FB. So SConnectionPoint is const but the FB itself not. How to implement this?

cochicde: For that particular case, I think this solution is the only one. But I'd change the caller actually to iterate over non-const connections. The caller looks like:

```
      for(const auto& it : mDestinationIds){
         it.getFB().connectDI(it.getPortId(), this);
       }
```

which to me seems counter-intuitive that `const` is specified to perform a non-const operation. The idea of  a `const CConnectionPoint` doesn't make much sense otherwise.

azoitl: Fixed this. For other use cases I will see where this whole thing leads us.

mx990 (approved):

cochicde (approved):

MartinMelikMerkumians (approved):

PR #393, opened 2025-03-28
First Version of Delegating Data Connection 

azoitl:


diplfranzhoepfinger:
> _ No description provided. _

what does this do ?

azoitl:
> > _ No description provided. _
> 
> what does this do ?

This is the base work to support negated boolean connections also in 4diac FORTE.

franz-ms-muc:
like NOT-AND aka NAND ?

azoitl:
> like NOT-AND aka NAND ?

Only at the input side but something like that.

franz-ms-muc:
Cooool

src/core/resource.cpp
```diff
@@ -188,85 +196,112 @@ EMGMResponse CResource::handleExecutionStateCmd(EMGMCommandType paCMD, forte::co
...
+EMGMResponse CResource::deleteConnection(forte::core::TNameIdentifier &paSrcNameList,
+                                         forte::core::TNameIdentifier &paDstNameList) {
+
+  auto srcIt = paSrcNameList.cbegin();
+  CFunctionBlock *srcFB = getFB(srcIt, paSrcNameList.cend());
+
+  auto dstIt = paDstNameList.cbegin();
+  CFunctionBlock *dstFB = getFB(dstIt, paDstNameList.cend());
+
+  if (!isConnPort(srcFB, srcIt, paSrcNameList.cend()) || isConnPort(dstFB, dstIt, paDstNameList.cend())) {
```
mx990:
```suggestion
  if (!isConnPort(srcFB, srcIt, paSrcNameList.cend()) || !isConnPort(dstFB, dstIt, paDstNameList.cend())) {
```

src/core/resource.cpp
```diff
@@ -188,85 +196,112 @@ EMGMResponse CResource::handleExecutionStateCmd(EMGMCommandType paCMD, forte::co
...
   }
 
   return retVal;
 }
 
-EMGMResponse CResource::deleteConnection(forte::core::TNameIdentifier &paSrcNameList, forte::core::TNameIdentifier &paDstNameList){
-  EMGMResponse retVal = EMGMResponse::NoSuchObject;
+namespace {
+  bool isConnPort(CFunctionBlock *const paFB,
+                  forte::core::TNameIdentifier::const_iterator &paNameListIt,
```
mx990: Why do you pass the iterator by reference?

src/core/typelib.cpp
```diff
@@ -171,6 +168,7 @@ CFunctionBlock *CTypeLib::createGenericFB(CStringDictionary::TStringId paInstanc
 }
 
 bool CTypeLib::deleteFB(CFunctionBlock *paFBToDelete) {
+  paFBToDelete->deinitialize();
```
mx990, line 171: I believe we should then just use `delete` in `createFB` and `createGenericFB` when the FB cannot be initialized/configured, in order to make sure we do not call `deinitialize` on an uninitialized/-configured FB.

mx990 (changes requested):

src/core/resource.cpp
azoitl: bad habit. Fixed.

src/core/typelib.cpp
azoitl: good point done

src/core/resource.cpp
azoitl: thx for noting this.

mx990 (approved):

PR #398, opened 2025-04-03
Move destination list to event connection

mx990:


src/core/eventconn.h
```diff
@@ -40,6 +40,39 @@ class CEventConnection : public CConnection{
      */
     void triggerEvent(CEventChainExecutionThread *paExecEnv) const;
 
+    /*! \brief Check if there are destinations added to this connection
+     *
+     * \return TRUE if there is no destination in the connection.
+     */
+    bool isEmpty() const {
+      return mDestinationIds.empty();
+    }
```
azoitl: I would not bring the isEmpty method to the EventConnection

azoitl (changes requested):

src/core/eventconn.h
mx990: Makes sense, removed.

azoitl (approved):

Issue #399, opened 2025-04-04
Upgrade 4diac FORTE to C++ 20


azoitl:
In our work towards the Eclipe 4diac 3.0 release with many major changes and major improvements we noticed that for implementing 4diac FORTE more correct and saver an upgrade to C++ 20 would be beneficial. In recent discussion we, for example, identified the following C++20 features:

- virtual constexpr methods and destructors, particularly for CIEC datatypes
  - enables constexpr literal operators for CIEC datatypes
  - enables usage of CIEC datatypes in switch labels, see https://github.com/eclipse-4diac/4diac-ide/issues/1106#issuecomment-2683165360
- literal operators for string ids, see https://github.com/eclipse-4diac/4diac-forte/pull/338#issuecomment-2691184906
- designated initializers, particularly for interface spec, see https://github.com/eclipse-4diac/4diac-forte/issues/392#issuecomment-2764716525
- concepts to make templates more type-safe, particularly for CIEC datatypes and 61131 functions
- aliasing shared_ptr construktor, see (8) in https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr, for better handling of connections
- std::span for improved interface spec handling: https://github.com/eclipse-4diac/4diac-forte/issues/392

And I guess there several more features that would be helpful.

Therefore I wanted to propose to upgrade the minimum required C++ version for 4diac FORTE to C++ 20.

This has some implication as maybe more exotic devices do not have an up to date compiler.

Therefore with this issue I want to assess explicitly with our users what compiler versions they use and what limitations are out there.

diplfranzhoepfinger:
Thanks for bring this up, i asked https://github.com/espressif if there are any issues.

ruspl-afed:
Perhaps it would be helpful to specify what kind of migration effort is expected from suppliers

azoitl:
> Perhaps it would be helpful to specify what kind of migration effort is expected from suppliers

@ruspl-afed the migration effort will be to use a compiler version that provides the according features. For GCC it would be somthing around v 12 or 14 (see https://en.cppreference.com/w/cpp/compiler_support/20 for details). If you use 4diac FBE no further steps should be needed. You would see it rather quick as 4diac FORTE would not compile with a wrong compiler.

jwalt:
I think this needs highlighting:

[...]

PR #404, opened 2025-04-10
Change minimum required standard to C++20

mx990:
Fixes https://github.com/eclipse-4diac/4diac-forte/issues/399

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -362,14 +362,15 @@ set_property(CACHE FORTE_ARCHITECTURE PROPERTY STRINGS None ${architectures})
 #######################################################################################
 
 # setting the C++ version has to be done directly before defining the targets only then modules can override this setting 
-if(NOT DEFINED CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 17)
-elseif(CMAKE_CXX_STANDARD LESS 17)
-  set(CMAKE_CXX_STANDARD 17)
-endif()
-if(NOT DEFINED CMAKE_CXX_STANDARD_REQUIRED)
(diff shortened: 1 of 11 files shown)


src/CMakeLists.txt
```diff
@@ -363,9 +363,9 @@ set_property(CACHE FORTE_ARCHITECTURE PROPERTY STRINGS None ${architectures})
 
 # setting the C++ version has to be done directly before defining the targets only then modules can override this setting 
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 17)
-elseif(CMAKE_CXX_STANDARD LESS 17)
-  set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD 20)
+elseif(CMAKE_CXX_STANDARD LESS 20)
+  set(CMAKE_CXX_STANDARD 20)
```
MartinMelikMerkumians, line 366: Should we put there also a warning log as we set the C++ standard to a higher version than previously set?

MartinMelikMerkumians (changes requested):

src/CMakeLists.txt
mx990: Good idea, changed it accordingly.

MartinMelikMerkumians (approved):

azoitl (approved):
As utf8 gets mandatory as it looks should I set the default encoding for the Eclipse CDT project that comes a long to UTF8 then?

mx990:
Yes, I believe that would make sense.

cochicde (approved):

azoitl (approved):

PR #406, opened 2025-04-15
Add connection reference tracking

azoitl:
FBs are now only deletable if they are not connected anymore.

### DIFF
--- a/src/core/cfb.cpp
+++ b/src/core/cfb.cpp
@@ -212,6 +212,11 @@ void CCompositeFB::createDataConnections(){
 
     if((srcFB != nullptr) && (dstFB != nullptr)) {
       establishConnection(getDataConn(srcFB, currentConn.mSrcId), *dstFB, currentConn.mDstId);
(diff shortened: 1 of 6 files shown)


src/core/cfb.cpp
```diff
@@ -212,6 +212,9 @@ void CCompositeFB::createDataConnections(){
 
     if((srcFB != nullptr) && (dstFB != nullptr)) {
       establishConnection(getDataConn(srcFB, currentConn.mSrcId), *dstFB, currentConn.mDstId);
+      if(srcFB == this){
+        decConnRefCount();
```
mx990, line 218: I would suggest adding a comment explaining why this is necessary.

mx990 (changes requested):

src/core/cfb.cpp
azoitl: good point done.

mx990 (approved):

PR #409, opened 2025-04-16
Remove connected check in OPCUA layer

mx990:
Remove connected check in OPCUA layer, since that information is no longer available.


mx990:
@azoitl I noticed that we still had an `isConnected()` check for data connections in the OPC UA layer. I hope this is safe to remove?

azoitl:
I had that on my todo list. I think there even more can be cleaned. Will look into it and PR on top of yours.

azoitl (approved):

PR #415, opened 2025-04-16
Add delegating connection deletion and query support

mx990:


src/core/util/inplace_vector.h
```diff
@@ -168,4 +169,15 @@ class inplace_vector {
...
 };
 
+template<typename T, std::size_t U, std::size_t V>
+[[nodiscard]]
+bool operator==(const inplace_vector<T, U> &paFirst, const inplace_vector<T, V> &paSecond) {
+    return paFirst.size() == paSecond.size() && std::equal(paFirst.cbegin(), paFirst.cend(), paSecond.cbegin());
+}
+
+template<typename T, std::size_t U, std::size_t V>
+[[__nodiscard__]]
```
azoitl: why are here the underscores?

src/core/adapterconn.h
```diff
@@ -30,6 +30,10 @@ class CAdapterConnection : public CConnection{
 
     EMGMResponse disconnect(CFunctionBlock &paDstFB, CStringDictionary::TStringId paDstPortNameId) override;
 
+    void getSourceName(char paSepChar, std::string &paResult) const override;
```
azoitl: as for most implementations the only difference is the source pin name. Wouldn't it be nicer to move the getSourcePin name into a pure virtual function so that we have less duplicated code. This would also not really effect the negated connection.

azoitl (changes requested):

src/core/util/inplace_vector.h
mx990: Removed the extra underscores. Thanks for noting!

src/core/adapterconn.h
mx990: Good point. I also added a join utility function to avoid duplicated code for assembling either a string or a list.

azoitl (approved):

PR #417, opened 2025-04-18
Add generated tracing files to the repo and fix some compilation errors 

cochicde:


azoitl (commented):
Looks good to me. But as we are now incorporating generated code and code from barectf into our repo we would need to fill an IP request.

cochicde:
> Looks good to me. But as we are now incorporating generated code and code from barectf into our repo we would need to fill an IP request.

right! I filed a request just now 

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/20782

azoitl:
> > Looks good to me. But as we are now incorporating generated code and code from barectf into our repo we would need to fill an IP request.
> 
> right! I filed a request just now
> 
> https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/20782

Thx

tests/core/trace/ctfTracerTest.cpp
ernstblechaPT, line 1: As this test is disabled via the `CMakeLists.txt` above whenever `FORTE_TRACE_CTF_REPLAY_DEBUGGING` is set, I don't see why the `#ifdef FORTE_TRACE_CTF_REPLAY_DEBUGGING` sections are added in this pull request?

cochicde: yes, 100% true. The idea is that the `ctfTracerTest` is used by both configurations. While cleaning everything up and testing, I realized that the replay debugging is tracing slightly differently and the test doesn't apply as it is now (the replay debuging one does not trace output events which are not connected to anything).

I'll unifty both approaches and have the `ctfTracerTest` then be valid for both cases. But I didn't want to do that in this PR and I'd rather leave the test as it is since it's a test that should pass for the replay debugging too

mx990 (approved):

ernstblechaPT (approved):

cochicde:
> > Looks good to me. But as we are now incorporating generated code and code from barectf into our repo we would need to fill an IP request.
> 
> right! I filed a request just now
> 
> https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/20782

this was done automatically. @azoitl is this enough and can we merge this? Or is there any other process needed?

azoitl:
@cochicde yes I think this is what is needed. What still would be great if you could add an entry to the root spdx file so that the dependency is also documented there.

azoitl (approved):

PR #419, opened 2025-04-18
Add cpp-linter action to GitHub workflow

mx990:
Fixes https://github.com/eclipse-4diac/4diac-forte/issues/379

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,5 @@
+src/core/trace/classic/generated/**
+src/core/trace/replay_debugging/generated/**
+src/stdfblib/**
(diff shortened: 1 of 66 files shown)


mx990:
For reference, the result of a failed check due to unformatted files is in https://github.com/eclipse-4diac/4diac-forte/actions/runs/15829275047.

gravemalte (approved):

mx990:
As per @ernstblechaPT's suggestion, changed `.clang-format-ignore` to use globstar, since we are explicitly requesting clang-format version 20, which has the necessary support.

ernstblechaPT (approved):

azoitl (approved):

PR #425, opened 2025-04-29
[cleanup] Removed CTOR Defines

azoitl:
The CTOR defines was original added to hide CTOR setups to make 4diac FORTE easier upgradeable. With the latest changes they are not realy used anymore and do not make sense. Therefore they are removed and the last FBs that still had them where ported to the latest setup. However, please note that these FBs have not tested in quite some while and my still not work.

### DIFF
--- a/src/modules/ads/ADS_SERVER_CONFIG_fbt.h
+++ b/src/modules/ads/ADS_SERVER_CONFIG_fbt.h
@@ -85,7 +85,6 @@ class FORTE_ADS_SERVER_CONFIG final : public CFunctionBlock {
     void setInitialValues() override;
 
   public:
-    FUNCTION_BLOCK_CTOR(FORTE_ADS_SERVER_CONFIG) {};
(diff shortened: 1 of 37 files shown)


mx990 (approved):

src/modules/ads/ADS_SERVER_CONFIG_fbt.h
```diff
@@ -85,7 +85,6 @@ class FORTE_ADS_SERVER_CONFIG final : public CFunctionBlock {
     void setInitialValues() override;
 
   public:
-    FUNCTION_BLOCK_CTOR(FORTE_ADS_SERVER_CONFIG) {};
```
cochicde, line 88: what about the constructor of this one now?

azoitl: This is the same as the one below. This was anyhow an artifact left from some clean-up. Which means ADS was not compiling.

PR #426, opened 2025-04-30
Force

azoitl:


src/core/funcbloc.cpp
```diff
@@ -40,8 +41,7 @@ CFunctionBlock::CFunctionBlock(forte::core::CFBContainer &paContainer,
     CFBContainer(paInstanceNameId, paContainer),
     mInterfaceSpec(paInterfaceSpec),
 #ifdef FORTE_SUPPORT_MONITORING
-    mEOMonitorCount(nullptr),
-    mEIMonitorCount(nullptr),
+    mEventMonitorCount(),
 #endif
```
mx990: Explicitly calling the `std::vector` default constructor should not be necessary.

mx990 (changes requested):

src/core/funcbloc.cpp
azoitl: good point. Fixed.

mx990 (approved):

PR #427, opened 2025-05-01
Rework set initial value

azoitl:


src/core/basicfb.cpp
```diff
@@ -26,12 +27,12 @@ CBasicFB::CBasicFB(forte::core::CFBContainer &paContainer,
 }
 
 void CBasicFB::setInitialValues() {
```
mx990, line 30: We should probably also move this implementation into `CGenFunctionBlock`. We could maybe put it inside an `if constexpr` that checks if the base class is a `CBasicFB`.

src/core/interfacespec.h
mx990, line 1: This seems to belong in a different PR?

src/core/interfacespec.cpp
mx990, line 1: This seems to belong in a different PR?

src/core/basicfb.cpp
```diff
@@ -26,12 +27,12 @@ CBasicFB::CBasicFB(forte::core::CFBContainer &paContainer,
...
-  CFunctionBlock::setInitialValues();
+  mECCState = CIEC_STATE(0);
```
mx990, line 32: Should we really always reset the ECC state when the FB is reset? Do we need a separate RETAIN flag for this?

src/core/basicfb.cpp
azoitl: I was about to do that. How would I do this?

src/core/interfacespec.h
azoitl: yes adeed accidentaly

src/core/interfacespec.cpp
azoitl: yes adeed accidentaly

src/core/basicfb.cpp
azoitl: nevermind found it.

mx990: It should work using:
```cpp
if constexpr (std::is_base_of_v<CBasicFB, T>) {
  // BasicFB-specific code here (qualify access to variables and methods from CBasicFB with T::)
}
```

mx990: Sorry, didn't see your latest reply before posting mine :sweat_smile:

src/core/basicfb.cpp
azoitl: How I understand the standard yes.

azoitl:
@mx990 do you know why this is not building her but locally?

mx990:
It works for me if I change `CBasicFB::` to `T::` in `genfb.h`.

src/core/genfb.h
```diff
@@ -238,6 +241,45 @@ bool CGenFunctionBlock<T>::configureFB(const char *paConfigString) {
...
+    genericInitializeDataPin(dataTypeIds, *getDI(i));
+  }
+
+  dataTypeIds = interfaceSpec.mDODataTypeNames;
+  for (TPortId i = 0; i < interfaceSpec.mNumDOs; ++i) {
+    genericInitializeDataPin(dataTypeIds, *getDO(i));
+  }
+
+  if constexpr (std::is_base_of_v<CBasicFB, T>) {
+    CBasicFB::setInitialValues();
```
mx990: I would suggest providing an empty default implementation in `CFunctionBlock` and then moving this line to the beginning as `T::setInitialValues();`. That way, it should work consistently with different base classes.

azoitl: done

src/core/cfb.cpp
```diff
@@ -142,6 +142,10 @@ EMGMResponse CCompositeFB::changeExecutionState(EMGMCommandType paCommand) {
   return nRetVal;
 }
 
+void CCompositeFB::setInitialValues() {
+  // currently nothing to do
```
mx990, line 148: I believe this should also call `CFunctionBlock::setInitialValues();` to be consistent?

mx990 (changes requested):

src/core/cfb.cpp
azoitl: Yes. Thx for noting.

azoitl: done

mx990 (approved):

PR #322, opened 2025-02-20
Add LOGIBUS_IO integration support for ESP32

franz-hoepfinger-4diac:
Add LOGIBUS_IO integration support for ESP32

dependencies documented in readme.txt

PR #430, opened 2025-05-05
Implementation of an eIO concept into Modular IO

scharf-m:
Implementation of a event-triggered IO (eIO) feature.
This system uses IO function blocks, which are similar to the standard IO FBs, except one additional adapter port. This adapter port is used to connect the FB to a configuration FB, which allows the user to configure trigger-events such as
- rising edge
- falling edge
- upper and lower limits and
- triggers on changing values (gradient).


franz-ms-muc:
this is interesting, but quite redundant to what i did with the 
https://components.espressif.com/components/espressif/button
https://github.com/espressif/esp-iot-solution/tree/master/components/button
and https://github.com/eclipse-4diac/4diac-forte/pull/322

With the IE Block we get Event Input. 
with the IX Block we get normal ( Raising & Falling for Bool)

IE can map to any Event, e.g. 

- Button pressed
- Button released
- Button pressed repeat
- Button press repeat done
- Button single click
- Button double click
- Button multiple click
- Button long press start
- Button long press hold
- Button long press up
- Button Press end

same we do for Buttons and Softkeys and AUX Input on ISOBUS. 

there we have only Pressed, released and held.

franz-ms-muc:
@scharf-m pls also publish your Blocks (*.fbt) TNX.

franz-ms-muc:
i shortly looked over, and i think at least your implementation is missing Long press and all the Click: like single click and double click.

scharf-m:
> @scharf-m pls also publish your Blocks (*.fbt) TNX.

The FBs: https://github.com/eclipse-4diac/4diac-ide/pull/1277

@franz-ms-muc Thx for the feedback!
About the missing button features: The here introduced eIO system is not really meant to serve just as button interface, but could be used as one. The idea is, to offer different trigger possibilities, not only for digital inputs (IX), but also for analog inputs (IW etc.). So each eIO FB has a config FB. In the `eIX` case, this would be `eIXconfig`. `eIXconfig` can register trigger on rising and/or falling edges. Using analog sensor values, for example an ADC,  `eIW` can be combined with  `eIWconfig`, to register trigger on changing values, or when thresholds are reached. (eIO and config blocks are connected via an adapter interface)

So the button feature is not implemented - yet! To achieve this, a new eIO config block could be used (lets say `eButtonConfig`), which gets connected to the `eIX` block and registers event-triggers on some of the conditions (button pressed, long press, double press ...) you suggested!

franz-ms-muc:
so, frankly speaking: 
i do not really like it. 

the Concept of "Mapping" is OK so far ( not perfect, not maximum User Friendly, but OK)

Mapping is the central Point of Configuration. 

if there are additional configs they should be added to the mapping place, 
and IX, IE and such should be as slim an tidy as possible. 
(i would like them even smaller)

azoitl:
@franz-ms-muc you don't need to use it it is an additional optional feature. Because of several reasons this concept has some advantages for other use cases therefore we would like to keep it.

franz-ms-muc:
> @franz-ms-muc you don't need to use it it is an additional optional feature. Because of several reasons this concept has some advantages for other use cases therefore we would like to keep it.

Think of the User ! should we really have another and another concept for the same things ? 
just with different Blocks ? 

making things harder and harder to understand for the newcomers :-(

If there are advantages, they should be explained very good, so it is visible.

src/core/io/device/io_controller.cpp
```diff
@@ -61,15 +61,51 @@ void IODeviceController::addHandle(HandleDescriptor &paHandleDescriptor) {
...
+      if (!paId.empty() && IOMapper::getInstance().registerHandle(paId, paHandle))
+        DEVLOG_INFO("[addHandle] %ss direction is `UnknownDirection`. Registered successfully.\r\n", paId.c_str());
+      break;
+    default:
+      DEVLOG_DEBUG("[addHandle] Not able to detect direction for %s.\r\n", paId.c_str());
+      break;
+  }
+}
+
+void IODeviceController::updateHandleList(std::string const &paId, IOHandle *paHandle) {
```
azoitl: this is not used in any code in this PR. Could you please remove it and add it in a pr where you need it if you really need it.

azoitl (changes requested):

src/core/io/device/io_controller.cpp
scharf-m: Sure - already fixed!

franz-ms-muc:
> so, frankly speaking: i do not really like it.
> the Concept of "Mapping" is OK so far ( not perfect, not maximum User Friendly, but OK)
> Mapping is the central Point of Configuration.

I ask the Opposite Question: 
removing the "Mapping" totally, and make all Configs on this Adapter ?

src/core/io/eIO/io/einputfb.h
```diff
@@ -0,0 +1,151 @@
...
+        return false;
+      };
+
+    private:
+      void executeEvent(TEventID paEIID, CEventChainExecutionThread *const paECET) final override {
+        switch(paEIID){
+          case cgExternalEventID:
+            if(this->var_QI) {
+              this->var_QO = this->read();
+              if (this->var_QO && peerFB && peerFB->checkEventTriggersConf(&var_IN)) {
```
azoitl: this should be moved to on change so that you are only generating an external event when the even ttrigger says so.

src/core/io/eIO/io/einputfb.h
```diff
@@ -0,0 +1,151 @@
...
+      virtual bool checkEventTriggers() {
+        return false;
+      };
+
+    private:
+      void executeEvent(TEventID paEIID, CEventChainExecutionThread *const paECET) final override {
+        switch(paEIID){
+          case cgExternalEventID:
+            if(this->var_QI) {
+              this->var_QO = this->read();
```
azoitl: this is already done in on chnage not needed here.

src/core/io/eIO/EConfig.h
```diff
@@ -0,0 +1,75 @@
...
+#include "devlog.h"
+
+class FORTE_eGenAdapter;
+/**
+ * This class handles the registration of event-triggers on the controller
+ * and maintains a record of all registered event-triggers from this instance, within this instance.
+ */
+class EConfig {
+  private:
+    CProcessInterfaceFB *peer;
```
azoitl: please ad m prefix  according to the codeing guidelines

azoitl: please for all members

src/core/io/eIO/EConfig.h
```diff
@@ -0,0 +1,75 @@
...
+
+  protected:
+    std::forward_list<ESpecBase*> eventGenList;
```
azoitl: a forwardlist seems a bit an overkill here. even if the list is changing once in a while a vector would be more efficient and performant.

src/core/io/eIO/EConfig.h
```diff
@@ -0,0 +1,75 @@
...
+
+   void registerEventTrigger(ESpecBase* event) {
```
azoitl: please add to all paramters a pa prefix according to the codeing guidelines

azoitl (changes requested):

azoitl:
> > @franz-ms-muc you don't need to use it it is an additional optional feature. Because of several reasons this concept has some advantages for other use cases therefore we would like to keep it.
> 
> Think of the User ! should we really have another and another concept for the same things ? just with different Blocks ?
> 
> making things harder and harder to understand for the newcomers :-(

Yes wee need to think of the users. But also we need a better interface. I think this are two separate things and how we offer different users different libraries is a something to be discussed and fixed elswehere.

src/core/io/eIO/io/einputfb.h
```diff
@@ -72,15 +72,23 @@ namespace forte::core::io {
...
-          peerFB = static_cast<EConfigFB *>(eGenAdapter->parentFB);
-          peerFB->eventGen();
+          mPeerFB = static_cast<EConfigFB *>(eGenAdapter->parentFB);
+          mPeerFB->eventGen();
         }
         return true;
       }
 
+      bool onChange() override {
+        if (!CInputFB<T>::onChange())
```
azoitl: please no ifs without {}

azoitl (changes requested):

src/core/io/eIO/io/einputfb.h
scharf-m: okay - fixed!

azoitl: please in all files and also for while and for loops.

scharf-m: hope I've got all of them now :)

azoitl (approved):

PR #434, opened 2025-05-05
Intermediate changes in toString methods of CIEC types

m-meingast:
Classes done: 

- CIEC_ANY_ELEMENTARY
- CIEC_ARRAY
- CIEC_STRUCT

Classes started:
- CIEC_STRING: `dollarEscapeChar()` needs Change for std::string& paValue

I tested the monitoring by creating a CFB with the corresponding datatype. In `appendDataWatches()` I included the CIEC type into the upper switch, where the toString(std::string &paValue) is called. It is there just for testing and should be replaced with just the toString(std::string &paValue) call, just as in the original version of the method. 
image

When I tested this, it was not possible to export a freeze CFB to a develop Forte, so I opted to switch to the Forte freeze branch for testing the CFB. The base version of this Draft is develop


src/core/basefb.cpp
```diff
@@ -56,51 +58,29 @@ CIEC_ANY *CBaseFB::getInternalVar(CStringDictionary::TStringId paInternalName) {
...
 
-int CBaseFB::toString(char *paValue, size_t paBufferSize) const {
-  int usedBuffer = CFunctionBlock::toString(paValue, paBufferSize);
-  if (usedBuffer < 1 || cmVarInternals == nullptr || (cmVarInternals != nullptr && cmVarInternals->mNumIntVars == 0)) {
-    return usedBuffer; // nothing to do
+void CBaseFB::toString(std::string &paValue) const {
+  auto startSize = paValue.size();
+  CFunctionBlock::toString(paValue);
+
+  if (cmVarInternals == nullptr || (cmVarInternals != nullptr && cmVarInternals->mNumIntVars == 0)) {
```
mx990:
```suggestion
  if (cmVarInternals == nullptr || cmVarInternals->mNumIntVars == 0) {
```

src/core/basefb.cpp
```diff
@@ -56,51 +58,29 @@ CIEC_ANY *CBaseFB::getInternalVar(CStringDictionary::TStringId paInternalName) {
...
+    return; // nothing to do
   }
 
-  --usedBuffer; // move the pointer to the position of the closing )
-  if (usedBuffer > 1) { // not only ()
-    strncpy(paValue + usedBuffer, csmToStringSeparator, paBufferSize - usedBuffer);
-    usedBuffer += sizeof(csmToStringSeparator) - 1;
+  paValue.resize(paValue.size() - 1); // move the pointer to the position of the closing )
+
+  if (paValue.size() - startSize > 1) { // not only ()
```
mx990:
```suggestion
  if (paValue.back() != '(') { // not only ()
```

src/core/datatypes/forte_any_real.cpp
```diff
@@ -14,23 +14,20 @@
...
+    return;
   }
-  // found none, check if a decimal point can be added
-  if (paUsedBytes + 2 >= static_cast<int>(paBufferSize)) {
-    return -1;
+
+  auto dotPos = paValue.find('.', paStartPos);
+  auto expPos = paValue.find('e', paStartPos);
+
+  if (dotPos == std::string::npos && expPos == std::string::npos) {
```
mx990, line 31: What if the string is `3e20`? We would also need to insert a `.0` before the exponent?

src/core/datatypes/forte_any_signed_variant.cpp
```diff
@@ -78,23 +80,10 @@ int CIEC_ANY_SIGNED_VARIANT::fromString(const char *paValue) {
   return nRetVal;
 }
 
-int CIEC_ANY_SIGNED_VARIANT::toString(char *paValue, size_t paBufferSize) const {
+void CIEC_ANY_SIGNED_VARIANT::toString(std::string &paValue) const {
   int result = -1;
```
mx990: I believe this line is no longer necessary?

src/core/datatypes/forte_any_string.cpp
```diff
@@ -138,10 +139,7 @@ bool CIEC_ANY_STRING::handleDollarEscapedChar(const char **paSymbol, bool paWide
   return bRetVal;
 }
 
-int CIEC_ANY_STRING::dollarEscapeChar(char *paValue,
-                                      char paSymbol,
-                                      unsigned int paBufferSize,
-                                      const EDataTypeID paTypeID) {
+void CIEC_ANY_STRING::dollarEscapeChar(std::string &paValue, char paSymbol, const EDataTypeID paTypeID) {
```
mx990: If we move this to `CIEC_ANY_CHARS`, could we also use this for `CHAR`?

src/core/datatypes/forte_any_string.cpp
```diff
@@ -188,28 +186,15 @@ int CIEC_ANY_STRING::dollarEscapeChar(char *paValue,
...
+    paValue += '$';
+    std::format_to(std::back_inserter(paValue), "{:02X}", paSymbol);
   } else if (nUsedBytes == 2) {
-    *paValue = '$';
-    *(++paValue) = cVal;
+    paValue += '$';
+    paValue += cVal;
   } else {
-    *paValue = cVal;
+    paValue += cVal;
```
mx990: I would suggest to append the corresponding character or escape sequence directly in the switch above.

src/core/datatypes/forte_array.h
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2022 Primetals Technologies Austria GmbH
- *               2023 Martin Erich Jobst
+ * Copyright (c) 2022, 2023 Primetals Technologies Austria GmbH,
```
mx990:
```suggestion
 * Copyright (c) 2022, 2025 Primetals Technologies Austria GmbH,
```

src/core/datatypes/forte_char.cpp
```diff
@@ -20,25 +21,28 @@ USE_STRING_ID(CHAR);
...
+    case '\n': paValue += "CHAR#'$N'"s; break;
+    case '\f':
+      paValue += "CHAR#'$P'"s; // page aka form feed
+      break;
+    case '\r': paValue += "CHAR#'$R'"s; break;
+    case '\t': paValue += "CHAR#'$T'"s; break;
+    default:
+      paValue += "CHAR#'"s;
+      paValue += symbol;
+      paValue += '\'';
```
mx990: It seems this has been missing a check for non-printable characters, which should produce hex-escapes.

src/core/datatypes/forte_char.cpp
```diff
@@ -20,25 +21,28 @@ USE_STRING_ID(CHAR);
...
-      case '\n': return snprintf(paValue, longestStringSize, "CHAR#'$N'");
-      case '\f': return snprintf(paValue, longestStringSize, "CHAR#'$P'"); // page aka form feed
-      case '\r': return snprintf(paValue, longestStringSize, "CHAR#'$R'");
-      case '\t': return snprintf(paValue, longestStringSize, "CHAR#'$T'");
-      default: return snprintf(paValue, longestStringSize, "CHAR#'%c'", symbol);
-    }
+void CIEC_CHAR::toString(std::string &paValue) const {
+  const unsigned char symbol = this->operator TForteChar();
+  switch (symbol) {
+    case '\0': paValue += "CHAR#''"s; break;
```
mx990: I believe this should produce `CHAR#'$00'`, since an empty character literal is invalid.

src/core/datatypes/forte_wchar.cpp
```diff
@@ -20,30 +23,32 @@ USE_STRING_ID(WCHAR);
...
-      case '\t': return snprintf(paValue, longestStringSize, "WCHAR#\"$T\"");
-      default: {
-        if (symbol < 256) {
-          return snprintf(paValue, longestStringSize, "WCHAR#\"%c\"", symbol);
-        }
-        return snprintf(paValue, longestStringSize, "WCHAR#\"$%X\"", symbol);
+void CIEC_WCHAR::toString(std::string &paValue) const {
+  const TForteWChar symbol = this->operator TForteWChar();
+  switch (symbol) {
+    case '\0': paValue += "WCHAR#\"\""s; break;
```
mx990: same regarding `WCHAR#'$0000'`

src/core/datatypes/forte_wchar.cpp
```diff
@@ -20,30 +23,32 @@ USE_STRING_ID(WCHAR);
...
+    case '"': paValue += "WCHAR#\"$\"\""s; break;
+    case '\n': paValue += "WCHAR#\"$N\""s; break;
+    case '\f':
+      paValue += "WCHAR#\"$P\""s; // page aka form feed
+      break;
+    case '\r': paValue += "WCHAR#\"$R\""s; break;
+    case '\t': paValue += "WCHAR#\"$T\""s; break;
+    default:
+      paValue += "WCHAR#\""s;
+      if (symbol < 256) {
```
mx990: I believe this should also check for printable characters.

src/core/funcbloc.cpp
```diff
@@ -480,113 +484,39 @@ void CFunctionBlock::setForce(TAbsDataPortNum paAbsDataPortNum, bool paForceValu
...
-    const CIEC_ANY *const variable = getDIO(i);
-    const CStringDictionary::TStringId nameId = getFBInterfaceSpec().mDIONames[i];
-    const char *varName = CStringDictionary::get(nameId);
-    bufferSize +=
-        strlen(varName) + 4 + variable->getToStringBufferSize(); // compensation for := and , for every variable
+    writeToStringNameValuePair(paValue, getFBInterfaceSpec().mDIONames[i], getDIO(i));
+    paValue += csmToStringSeparator;
+  }
+
+  if (paValue.size() - startSize > 3) {
```
mx990:
```suggestion
  if (paValue.back() != '(') {
```

src/core/funcbloc.h
```diff
@@ -604,10 +603,9 @@ class CFunctionBlock : public forte::core::CFBContainer {
     CFunctionBlock(const CFunctionBlock &) = delete;
 
   protected:
-    int writeToStringNameValuePair(char *paValue,
-                                   size_t paBufferSize,
-                                   const CStringDictionary::TStringId variableNameId,
-                                   const CIEC_ANY *const variable) const;
+    void writeToStringNameValuePair(std::string &paValue,
```
mx990: I believe this can be made static? Maybe we could even make this into a utility function (e.g., in `string_utils.h`) and also use it for structs?

src/core/monitoring.cpp
```diff
@@ -234,9 +229,7 @@ namespace {
     CIEC_UDINT udint(paEventWatchEntry.mEventDataBuf);
 
     paResponse += "<Data value=\""s;
-    char buf[21]; // the biggest number in an ulint is 18446744073709551616, TODO directly use paResponse
-    udint.toString(buf, sizeof(buf));
-    paResponse += buf;
+    udint.toString(paResponse);
     paResponse += "\"/>\n</Port>"s;
```
mx990: Maybe we could directly format the `TForteUInt32` using `std::format_to` without taking the detour via `CIEC_UDINT`?

src/core/resource.cpp
```diff
@@ -304,59 +305,24 @@ void CResource::setInitialValues() {
...
+  switch (var->getDataTypeID()) {
+    case CIEC_ANY::e_STRING:
+    case CIEC_ANY::e_WSTRING: {
+      auto start = paValue.size();
+      static_cast<CIEC_ANY_STRING &>(*var).toUTF8(paValue, false);
+      if (start != paValue.size()) {
+        forte::core::util::transformNonEscapedToEscapedXMLText(paValue, start);
+      }
+    } break;
+    default: var->toString(paValue); break;
```
mx990, line 331: What happens if there is a struct or array containing a string with invalid XML characters?

src/core/util/string_utils.cpp
```diff
@@ -225,23 +228,19 @@ size_t forte::core::util::getExtraSizeForXMLEscapedChars(const char *paString) {
...
+  for (size_t runner = paString.size() - 1; runner >= paStart && runner != static_cast<size_t>(-1); runner--) {
+    auto escapeChar = scEscapeMap.find(paString[runner]);
     if (escapeChar != scEscapeMap.end()) {
       const char *const escapSeq = std::get<0>(escapeChar->second);
-      size_t toMove = std::get<1>(escapeChar->second);
-      memmove(&runner[toMove], runner + 1, originalEnd - runner + retVal);
-      memcpy(runner, escapSeq, toMove);
-      retVal += toMove - 1;
+      paString.replace(runner, 1, escapSeq, 1);
+      paString.insert(runner + 1, escapSeq + 1);
```
mx990:
```suggestion
      paString.replace(runner, 1, escapSeq);
```

tests/core/datatypes/CIEC_ARRAY_test.cpp
```diff
@@ -175,79 +174,82 @@ BOOST_AUTO_TEST_CASE(Array_assignment_test_INT_with_IEC_type_indices) {
...
   BOOST_CHECK_EQUAL(static_cast<CIEC_INT::TValueType>(static_cast<CIEC_INT &>(nTest[2])), 3);
   BOOST_CHECK_EQUAL(static_cast<CIEC_INT::TValueType>(static_cast<CIEC_INT &>(nTest[3])), 4);
   BOOST_CHECK_EQUAL(static_cast<CIEC_INT::TValueType>(static_cast<CIEC_INT &>(nTest[4])), 5);
-  BOOST_CHECK_EQUAL(nTest.toString(acBuffer, 2), -1);
-  BOOST_CHECK_EQUAL(nTest.toString(acBuffer, 30), 11);
-  BOOST_CHECK_EQUAL(strcmp(acBuffer, "[1,2,3,4,5]"), 0);
+  nTest.toString(buffer);
+  BOOST_CHECK_EQUAL(buffer, "[1,2,3,4,5]");
 
+  buffer.clear();
```
mx990, line 211: Why do you keep clearing the buffer when it is not used? I would suggest to clear the buffer either directly before `toString` or after checking the result to keep this as one block.

tests/core/datatypes/CIEC_REAL_test.cpp
```diff
@@ -111,106 +113,104 @@ BOOST_AUTO_TEST_CASE(Equality_test) {
...
+  testVal = CIEC_REAL(-INFINITY);
+  testVal.toString(buffer);
+  BOOST_CHECK_EQUAL(buffer, "-inf");
+
+  buffer.clear();
+  testVal = CIEC_REAL(NAN);
+  testVal.toString(buffer);
+  BOOST_CHECK_EQUAL(buffer, "nan");
+
+#ifndef _MSC_VER // FIXME enable tests for MSVC after switching to std::format
```
mx990: Please enable this test for MSVC now that we have switched to `std:format_to`.

tests/core/datatypes/CIEC_LREAL_test.cpp
```diff
@@ -109,106 +111,107 @@ BOOST_AUTO_TEST_CASE(Equality_test) {
...
+  testVal = CIEC_LREAL(-INFINITY);
+  testVal.toString(buffer);
+  BOOST_CHECK_EQUAL(buffer, "-inf");
+
+  buffer.clear();
+  testVal = CIEC_LREAL(NAN);
+  testVal.toString(buffer);
+  BOOST_CHECK_EQUAL(buffer, "nan");
+
+#ifndef _MSC_VER // FIXME enable tests for MSVC after switching to std::format
```
mx990: Please enable this test for MSVC now that we have switched to `std:format_to`.

mx990 (changes requested):

mx990:
I believe we also need to update `4diac-fortebuildcontainer` with a more recent compiler.

azoitl:
> I believe we also need to update `4diac-fortebuildcontainer` with a more recent compiler.

Who knows how to do that?

src/core/datatypes/forte_any_real.cpp
azoitl: according to the tests this is handled by the g format option.

mx990:
> > I believe we also need to update `4diac-fortebuildcontainer` with a more recent compiler.
> 
> Who knows how to do that?

The image should (hopefully) build automatically after merging https://github.com/eclipse-4diac/4diac-fortebuildcontainer/pull/4.

MartinMelikMerkumiansElectrify:
> > > I believe we also need to update `4diac-fortebuildcontainer` with a more recent compiler.
> > 
> > 
> > Who knows how to do that?
> 
> The image should (hopefully) build automatically after merging [eclipse-4diac/4diac-fortebuildcontainer#4](https://github.com/eclipse-4diac/4diac-fortebuildcontainer/pull/4).

It didn't last time I checked, as the whole Jenkins workflow didn't work/start anymore

src/core/datatypes/forte_any_real.cpp
mx990: I don't see any tests to that effect. I have created a PR https://github.com/m-meingast/4diac-forte/pull/1 that fixes the problem and also adds the missing tests.

mx990:
> > > > I believe we also need to update `4diac-fortebuildcontainer` with a more recent compiler.
> > > 
> > > 
> > > Who knows how to do that?
> > 
> > 
> > The image should (hopefully) build automatically after merging [eclipse-4diac/4diac-fortebuildcontainer#4](https://github.com/eclipse-4diac/4diac-fortebuildcontainer/pull/4).
> 
> It didn't last time I checked, as the whole Jenkins workflow didn't work/start anymore

I got the workflow to start but then it failed because it is missing the `containertools` container. After checking the Wiki,  it seems building custom container images is no longer supported:
> [I want to build a custom Docker image (with docker build), but it does not work. What should I do?](https://wiki.eclipse.org/Hudson#I_want_to_build_a_custom_Docker_image_.28with_docker_build.29.2C_but_it_does_not_work._What_should_I_do.3F)
> > You cannot currently build images on the cluster (i.e. docker build does not work). We plan to address this shortcoming shortly.

I have therefore created a PR https://github.com/eclipse-4diac/4diac-forte/pull/453 to replace the fortebuildcontainer with the default ubuntu image in the system tests.

src/core/util/string_utils.cpp
azoitl: done

tests/core/datatypes/CIEC_ARRAY_test.cpp
azoitl: I guess this are remains of the previous error test cases that I had to remove. Fixed

tests/core/datatypes/CIEC_LREAL_test.cpp
azoitl: done

src/core/basefb.cpp
azoitl: done

src/core/datatypes/forte_any_signed_variant.cpp
azoitl: yes, done

src/core/datatypes/forte_any_string.cpp
azoitl: changed makes a much nicer code.

src/core/datatypes/forte_array.h
azoitl: fixed

src/core/funcbloc.cpp
azoitl: done

src/core/monitoring.cpp
azoitl: good idea, this code comes from a time where we wanted to use things we allready had.

src/core/resource.cpp
azoitl: harmonized with the watches

tests/core/datatypes/CIEC_REAL_test.cpp
azoitl: done

src/core/datatypes/forte_any_real.cpp
azoitl: Interesting. When I implemented the new version I had exactly your code but then saw that the tests had someting in it. I may need new glasses. Thx for the fix.

src/core/basefb.cpp
azoitl: done

src/core/datatypes/forte_any_string.cpp
azoitl: yes this is a good idea, done

src/core/datatypes/forte_char.cpp
azoitl: with the move to dollarEscapeChar this is fixed

src/core/datatypes/forte_char.cpp
azoitl: As we now use dollarEscapeChar this is fixed as well.

src/core/datatypes/forte_wchar.cpp
azoitl: done

src/core/datatypes/forte_wchar.cpp
azoitl: good point. Done.

src/core/funcbloc.h
azoitl: good idea did that.

src/core/datatypes/forte_any_chars.cpp
```diff
@@ -0,0 +1,50 @@
...
+  switch (paSymbol) {
+    case '$': paValue += "$$"s; break;
+    case '\'':
+      if (paTypeID == CIEC_ANY::e_STRING) {
+        paValue += "$'"s;
+      } else {
+        paValue += paSymbol;
+      }
+      break;
+    case 0x10: // line feed
```
mx990: `0x10` is *Data Link Escape* and not *Line Feed*[^1]. I believe this is supposed to be decimal `10` or `0x0a`, but that is already handled by `\n` below.

As far as I am aware, both `$L` and `$N` should, in fact, map to the same character, see also https://github.com/eclipse-4diac/4diac-forte/blob/0f1a914bfaa87d6c64dd0470daf530eb9fdaf81a/src/core/datatypes/forte_char.cpp#L74-L75

[^1]: https://en.wikipedia.org/wiki/ASCII#Control_code_table

src/core/datatypes/forte_any_chars.h
```diff
@@ -33,11 +32,11 @@ class CIEC_ANY_CHARS : public CIEC_ANY_ELEMENTARY {
       return CIEC_ANY::e_ANY;
     }
 
-    // int toString(char* paValue, size_t paBufferSize) const override;
-    // int fromString(const char *paValue) override;
-
   protected:
     CIEC_ANY_CHARS() = default;
+
+    // Use null as destination for just determining the need of escaping
```
mx990: I believe this comment does not make sense anymore.

mx990 (changes requested):

src/core/datatypes/forte_any_real.cpp
azoitl: I think we should merge this one here as it provides the same functionality as before and then take your PR https://github.com/m-meingast/4diac-forte/pull/1 and move it here.

mx990: I agree. I'll rebase and create a new PR here once this is merged.

src/core/datatypes/convert/convert_functions.h
```diff
@@ -64,21 +65,15 @@
...
-  size_t bufferSize = paVal.getToStringBufferSize();
-
-  paString.reserve(static_cast<TForteUInt16>(bufferSize));
-  char *pacBuffer = paString.getValue();
-  int nWrittenBytes = paVal.toString(pacBuffer, bufferSize);
-  nWrittenBytes = nWrittenBytes > -1 ? nWrittenBytes : 0;
-  paString.assign(pacBuffer, static_cast<TForteUInt16>(nWrittenBytes));
+  std::string buffer;
+  paVal.toString(buffer);
+  paString = CIEC_WSTRING(buffer.c_str());
```
MartinMelikMerkumiansElectrify, line 70: Just to be sure, you are aware that this is still doing the wrong thing (as the original code did the wrong thing), because WSTRING would need to store it in a 2-byte format?

src/core/datatypes/forte_any.cpp
```diff
@@ -9,11 +10,11 @@
...
- *    Alois Zoitl, Ingo Hegny, Martin Melik Merkumians, Stanislav Meduna,
- *    Monika Wenger, Matthias Plasch
- *      - initial implementation and rework communication infrastructure
- *    Martin Jobst
- *      - add support for data types with different size
+ *   Alois Zoitl, Ingo Hegny, Martin Melik Merkumians, Stanislav Meduna,
+ *     Monika Wenger, Matthias Plasch
+ *                - initial implementation and rework communication infrastructure
+ *   Martin Jobst - add support for data types with different size
+ *   Alois Zoitl  - migrated data type toString to std::string
```
MartinMelikMerkumiansElectrify: The only change I see here, is the update of the copyright header. is this on purpose?

src/core/datatypes/forte_any_bit_variant.cpp
```diff
@@ -80,23 +82,9 @@ int CIEC_ANY_BIT_VARIANT::fromString(const char *paValue) {
   return nRetVal;
 }
 
-int CIEC_ANY_BIT_VARIANT::toString(char *paValue, size_t paBufferSize) const {
-  int result = -1;
+void CIEC_ANY_BIT_VARIANT::toString(std::string &paValue) const {
```
MartinMelikMerkumiansElectrify: is paValue the correct name to use here? because its not the value to be process, but a buffer to be written to

(as a general remark for this method)

src/core/datatypes/forte_any_char.h
```diff
@@ -15,7 +15,6 @@
 #define _ANY_CHAR_H_
 
 #include "forte_any_chars.h"
-#include <ctype.h>
 #include <string_utils.h>
```
MartinMelikMerkumiansElectrify: as you already change this file, the string_utils.h should be using "" and not <>

src/core/datatypes/forte_any_chars.cpp
```diff
@@ -0,0 +1,47 @@
...
+using namespace std::literals::string_literals;
+
+void CIEC_ANY_CHARS::dollarEscapeChar(std::string &paValue, char paSymbol, const EDataTypeID paTypeID) {
+  switch (paSymbol) {
+    case '$': paValue += "$$"s; break;
+    case '\'':
+      if (paTypeID == CIEC_ANY::e_STRING) {
+        paValue += "$'"s;
+      } else {
+        paValue += paSymbol;
```
MartinMelikMerkumiansElectrify: I find this (and line 36) confusing, because you always replace the symbol, but here you use the original one. I think for simplicity's sake and making the code easier to understand for newcomers, also replace here with the corresponding non-escaped symbol

src/core/datatypes/forte_any_chars.cpp
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Contributors:
+ *   Alois Zoitl - extracted and reimplemnted from forte_any_string
+ *******************************************************************************/
+#include <format>
+#include "forte_any_chars.h"
+
+using namespace std::literals::string_literals;
+
+void CIEC_ANY_CHARS::dollarEscapeChar(std::string &paValue, char paSymbol, const EDataTypeID paTypeID) {
```
MartinMelikMerkumiansElectrify: also here, I dont think paValue is a fitting name, as this is some kind of provided memory in the end, and not something which is used as an input

src/core/datatypes/forte_any_elementary.cpp
```diff
@@ -150,16 +150,12 @@ int CIEC_ANY_ELEMENTARY::toString(char *paValue, size_t paBufferSize) const {
       nDivisor = 10000000000000000000ULL;
       bSigned = false;
       break;
-    default: DEVLOG_ERROR("Attempt to call CIEC_ANY::toString in CIEC_ANY_ELEMENTARY\n"); return -1;
+    default: DEVLOG_ERROR("Attempt to call CIEC_ANY::toString in CIEC_ANY_ELEMENTARY\n"); return;
```
MartinMelikMerkumiansElectrify, line 153: is this allowed according to the coding guidelines?

src/core/datatypes/forte_any_elementary.cpp
```diff
@@ -150,16 +150,12 @@ int CIEC_ANY_ELEMENTARY::toString(char *paValue, size_t paBufferSize) const {
...
   }
 
   if (true == bSigned) {
     if (nSBuffer < 0) {
-      if (nRetVal >= static_cast<int>(paBufferSize)) {
-        return -1;
-      }
-      paValue[nRetVal] = '-';
-      nRetVal++;
+      paValue += '-';
```
MartinMelikMerkumiansElectrify: shouldn't you use "-"s here? because now it is a char literal

src/core/datatypes/forte_any_elementary.cpp
```diff
@@ -173,28 +169,12 @@ int CIEC_ANY_ELEMENTARY::toString(char *paValue, size_t paBufferSize) const {
...
-    if (nRetVal >= static_cast<int>(paBufferSize)) {
-      return -1;
-    }
-
-    paValue[nRetVal] = static_cast<char>(static_cast<char>(nUBuffer / nDivisor) + '0');
-    nUBuffer = nUBuffer - (paValue[nRetVal] - '0') * nDivisor;
+    paValue += static_cast<char>(static_cast<char>(nUBuffer / nDivisor) + '0');
+    nUBuffer = nUBuffer - (paValue.back() - '0') * nDivisor;
     nDivisor /= 10;
-    nRetVal++;
```
MartinMelikMerkumiansElectrify, line 183: this is still using Hungarian notation

src/core/datatypes/forte_any_elementary.cpp
```diff
@@ -173,28 +169,12 @@ int CIEC_ANY_ELEMENTARY::toString(char *paValue, size_t paBufferSize) const {
     } else {
       bLeadingZeros = false;
```
MartinMelikMerkumiansElectrify: Hungarian notiation

src/core/datatypes/forte_any_elementary.cpp
```diff
@@ -150,16 +150,12 @@ int CIEC_ANY_ELEMENTARY::toString(char *paValue, size_t paBufferSize) const {
...
 
   if (true == bSigned) {
     if (nSBuffer < 0) {
-      if (nRetVal >= static_cast<int>(paBufferSize)) {
-        return -1;
-      }
-      paValue[nRetVal] = '-';
-      nRetVal++;
+      paValue += '-';
       nSBuffer *= -1;
```
MartinMelikMerkumiansElectrify: Hungarian notiation

src/core/datatypes/forte_array.cpp
```diff
@@ -39,168 +40,51 @@ void CIEC_ARRAY::setValue(const CIEC_ANY &paValue) {
...
-      if (lastElement != nullptr && !lastElement->equals(operator[](i))) {
-        int usedBytesByElement = toCollapsedElementString(*lastElement, count, nBytesUsed > 1, paValue, paBufferSize);
-        if (usedBytesByElement < 0) {
-          return -1;
-        }
-        paValue += usedBytesByElement;
-        paBufferSize -= static_cast<size_t>(usedBytesByElement);
-        nBytesUsed += usedBytesByElement;
-        count = 0;
+void CIEC_ARRAY::toCollapsedString(std::string &paValue) const {
```
MartinMelikMerkumiansElectrify: again, not sure if paValue is a good name for a memory buffer kind-of object

src/core/datatypes/forte_bool.cpp
```diff
@@ -55,19 +59,8 @@ int CIEC_BOOL::fromString(const char *paValue) {
...
-      nRetval = 4;
-    } else {
-      strncpy(paValue, "FALSE", paBufferSize);
-      nRetval = 5;
-    }
-    paValue[paBufferSize - 1] = '\0';
-  }
-  return nRetval;
+void CIEC_BOOL::toString(std::string &paValue) const {
+  paValue += (getTBOOL8()) ? "TRUE"s : "FALSE"s;
```
MartinMelikMerkumiansElectrify: the other parentheses are superfluous

src/core/datatypes/forte_date.cpp
```diff
@@ -70,18 +68,11 @@ int CIEC_DATE::fromString(const char *paValue) {
   return static_cast<unsigned int>(acBuffer - paValue);
 }
 
-int CIEC_DATE::toString(char *paValue, size_t paBufferSize) const {
-  int nRetVal = -1;
+void CIEC_DATE::toString(std::string &paValue) const {
   struct tm ptm;
```
MartinMelikMerkumiansElectrify: perhaps I am wrong, but the `struct` keyword is not needed in C++

src/core/datatypes/forte_ltime_of_day.cpp
```diff
@@ -91,17 +90,12 @@ int CIEC_LTIME_OF_DAY::fromString(const char *paValue) {
   return static_cast<int>(acBuffer - paValue);
 }
 
-int CIEC_LTIME_OF_DAY::toString(char *paValue, size_t paBufferSize) const {
+void CIEC_LTIME_OF_DAY::toString(std::string &paValue) const {
   TForteUInt64 ntoStingBuffer = getTUINT64();
-  time_t t = static_cast<time_t>(ntoStingBuffer / (1000ULL * 1000000ULL));
+  time_t t = static_cast<time_t>(ntoStingBuffer / 1000000000);
```
MartinMelikMerkumiansElectrify: from my experience it is favorable to use the type suffix (LL) to be sure that the C++ compiler is generating the correct operations. I had strange results when not using the suffix

src/core/datatypes/forte_ltime_of_day.cpp
```diff
@@ -91,17 +90,12 @@ int CIEC_LTIME_OF_DAY::fromString(const char *paValue) {
...
 
-  int nRetVal =
-      forte_snprintf(paValue, paBufferSize, "LTOD#%02d:%02d:%02d.%03d", (int) (t / 3600), (int) ((t % 3600) / 60),
-                     (int) (t % 60), (int) ((ntoStingBuffer / 1000000ULL) % 1000ULL));
-  if ((nRetVal < -1) || (nRetVal >= static_cast<int>(paBufferSize))) {
-    nRetVal = -1;
-  }
-  return nRetVal;
+  std::format_to(std::back_inserter(paValue), "LTOD#{:02}:{:02}:{:02}.{:03}", (int) (t / 3600), (int) ((t % 3600) / 60),
```
MartinMelikMerkumiansElectrify: please change the C-style casts to C++ style casts, as written in the coding guidelines

src/core/monitoring.cpp
```diff
@@ -95,100 +92,28 @@ namespace {
...
         case CIEC_ANY::e_ANY:
         case CIEC_ANY::e_WSTRING:
         case CIEC_ANY::e_STRING:
         case CIEC_ANY::e_CHAR:
         case CIEC_ANY::e_WCHAR:
         case CIEC_ANY::e_ARRAY:
-        case CIEC_ANY::e_STRUCT:
-          consumedBytes += static_cast<int>(forte::core::util::transformNonEscapedToEscapedXMLText(acDataValue));
-          break;
+        case CIEC_ANY::e_STRUCT: forte::core::util::transformNonEscapedToEscapedXMLText(paResponse, start); break;
```
MartinMelikMerkumiansElectrify, line 110: shouldnt that be on two lines? Not exactly sure what the coding guideline says, but usually we have one statement per line

MartinMelikMerkumiansElectrify (changes requested):
For the tests, could you please change your modified tests to use `BOOST_TEST` instead of `BOOST_TEST_EQUAL` to streamline the usage. you even changed one test from `BOOST_TEST` to `BOOST_TEST_EQUAL`

src/core/monitoring.cpp
mx990: We changed the formatting to allow short case labels on a single line, see https://github.com/eclipse-4diac/4diac-forte/pull/402.

src/core/datatypes/convert/convert_functions.h
azoitl: yes I'm aware. This pr already got totally out of hand just to provide the same or a bit better functionality then before. Fixing bigger things is for me definitely off for another PR.

src/core/datatypes/forte_ltime_of_day.cpp
azoitl: thx for noting I think this is a copy paste issue from the non long time version. Fixed.

src/core/datatypes/forte_any_elementary.cpp
azoitl: As I want to append a single char I found it clearer to use just the char here and not a string literal.

src/core/datatypes/forte_any_elementary.cpp
azoitl: yes. If not please comment on the clang formater issue.

src/core/datatypes/forte_any_bit_variant.cpp
azoitl: changed all paValues of toString and related functions to paTargetBuf

src/core/datatypes/forte_any.cpp
azoitl: I think this was an accident. thx for noting

src/core/datatypes/forte_any_char.h
azoitl: even removed as not needed

src/core/datatypes/forte_any_chars.cpp
azoitl: good point. changed

src/core/datatypes/forte_any_chars.cpp
azoitl: see above

src/core/datatypes/forte_any_elementary.cpp
azoitl: fixed

src/core/datatypes/forte_any_elementary.cpp
azoitl: fixed

src/core/datatypes/forte_any_elementary.cpp
azoitl: fixed

src/core/datatypes/forte_array.cpp
azoitl: see above

src/core/datatypes/forte_bool.cpp
azoitl: removed

src/core/datatypes/forte_date.cpp
azoitl: yes it is possible. fixed.

src/core/datatypes/forte_ltime_of_day.cpp
azoitl: fixed

azoitl:
@MartinMelikMerkumiansElectrify  I tried to fix all things you reported. Put it into a separate commit for easier cross checking.

MartinMelikMerkumiansElectrify (approved):

azoitl:
@mx990 @MartinMelikMerkumiansElectrify  I'm a bit confused by the test fail. We have a test string that is `"\x10\x10"_STRING` but when I debug the toString method I get `0x20` values for each char is this correct? How should we handle it?

mx990:
When I run the tests locally, I see decimal `16`, octal `'\020'`, or hex `0x10`. However, those values make little sense in the first place, see my comment https://github.com/eclipse-4diac/4diac-forte/pull/434#discussion_r2093781641 above.

azoitl:
So I just remove the test?

mx990:
Yes, I believe it should be removed.

azoitl:
I checked all test and our fromStringCode as well as the standard again. The problem is that the standard demands that both `$n` and `$l` needs to be distinguishable. Therefore I think the original functionality should come back, even if it is not perfectly correct.

mx990:
> I checked all test and our fromStringCode as well as the standard again. The problem is that the standard demands that both `$n` and `$l` needs to be distinguishable.

Where does it say that?

> Therefore I think the original functionality should come back, even if it is not perfectly correct.

So we should just assign an arbitrary character value for `$l` then?

MartinMelikMerkumiansElectrify:
I checked the 2003 version of the standard Table 6 Note 1 says $N is vendor-specific (who would have thought that? :P)

azoitl:
Interesting. I think as this is not totally straight forward I would bring back the previous implementation and move this to a separate issue and PR.

mx990:
I agree that we should better separate this. Since I am the one who originally raised this issue here, I would be happy to create the separate issue and PR.

mx990:
I have created a separate issue https://github.com/eclipse-4diac/4diac-forte/issues/461.

azoitl:
> I have created a separate issue #461.

thx

azoitl:
@mx990 @MartinMelikMerkumiansElectrify I now brought back the original behavior cleaned the commit. All tests run through and I hope that I haven't forgotten anything else.

mx990 (approved):

PR #436, opened 2025-05-05
position detection for Wago 750-636, update supported IOs

moniwe:
This PR fixes the current and target position of 750-636, that did not work within the previous PR. It relates to https://github.com/eclipse-4diac/4diac-ide/pull/1278 

There seems to be a problem with the adapters for the modular IOs, during compiling I get:
```
[01m[K/.../forte/src/modules/wagokbus/types/WagoBusAdapter.cpp:108:19:[m[K [01;31m[Kerror: [m[K'[01m[KscmFBInterfaceSpec[m[K' was not declared in this scope; did you mean '[01m[KSFBInterfaceSpec[m[K'?
  108 |         writeData([01;31m[KscmFBInterfaceSpec[m[K.mNumDIs + 2, *mDOs[2], mDOConns[2]);
      |                   [01;31m[K^~~~~~[m[K
      |                   [32m[KSFBInterfaceSpec[m[K
```

at least revolution pi and wago seem to use _scmFBInterfaceSpec_ within the cpp but within the header there is _scmFBInterfaceSpecSocket_ and _scmFBInterfaceSpecPlug_. Not sure if this should be changed within this PR or there should be a separate one.

### DIFF
--- a/src/modules/wagokbus/WagoDeviceController.cpp
+++ b/src/modules/wagokbus/WagoDeviceController.cpp
@@ -105,14 +105,12 @@ WagoDeviceController::createIOHandle(forte::core::io::IODeviceController::Handle
       inputOffset = mTerminalInfos[desc.mSlaveIndex].OffsetInput_bits + (2 * desc.mChannel);
       break;
     case CIEC_ANY::e_DWORD:
-      outputOffset = mTerminalInfos[desc.mSlaveIndex].OffsetOutput_bits + (4 * desc.mChannel);
-      inputOffset = mTerminalInfos[desc.mSlaveIndex].OffsetInput_bits + (4 * desc.mChannel);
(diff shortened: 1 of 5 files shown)


azoitl:
with the latest changes of writeData you need to provide the absolute datapoint ID. this is for outputs the numInputs + outputnum.

as this depends on if plug or socket we need to be more specific. See [ARTimeOut_adp.cpp::writeData](https://github.com/eclipse-4diac/4diac-forte/blob/bee2335f9dea47132e285c9e9e74416ab9f678ee/src/stdfblib/events/ARTimeOut_adp.cpp#L92C23-L92C38) as refrence

cochicde (approved):
Looks good!

azoitl (approved):

PR #439, opened 2025-05-06
Rework monitoring data to not use CSingleLinkedList

azoitl:


src/core/monitoring.h
```diff
@@ -20,124 +20,140 @@
...
+
+      private:
+        CStringDictionary::TStringId mPortId;
+    };
+
+    class CDataWatchEntry : public CWatchEntry {
+      public:
+        CDataWatchEntry(CStringDictionary::TStringId paPortId, CIEC_ANY &paDataValue, TAbsDataPortNum paForceIndex) :
+            CWatchEntry(paPortId),
+            mDataBuffer(std::unique_ptr<CIEC_ANY>(paDataValue.clone(nullptr))),
```
mx990: The intermediate temporary should not be necessary:
```suggestion
            mDataBuffer(paDataValue.clone(nullptr)),
```

src/core/monitoring.h
```diff
@@ -20,124 +20,140 @@
...
-
-        static size_t getExtraSizeForEscapedCharsArray(const CIEC_ARRAY &paDataValue);
-
-        static size_t getExtraSizeForEscapedCharsStruct(const CIEC_STRUCT &paDataValue);
-
-        //! List storing all FBs which are currently monitored
-        TFBMonitoringList mFBMonitoringList;
-        CResource &mResource; //!< The resource this monitoring handler manages
+        TAbsDataPortNum mForceIndex;
+        CIEC_ANY *mDataValueRef; //!< reference to the data point to watch
```
mx990: I would suggest to use a [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) or rename the variable and change the comment to reflect that it is in fact *not* a reference.

src/core/monitoring.h
```diff
@@ -20,124 +20,140 @@
...
+
+        TForteUInt32 mEventDataBuf; //!< buffer for the event count
+
+        CEventWatchEntry(const CEventWatchEntry &) = delete;
+        CEventWatchEntry(CEventWatchEntry &&) = default;
+        CEventWatchEntry &operator=(const CEventWatchEntry &) = delete;
+        CEventWatchEntry &operator=(CEventWatchEntry &&) = default;
+
+      private:
+        TForteUInt32 *mEventDataRef; //!< reference to the event counter of the watched event pin
```
mx990: same here regarding the reference

src/core/monitoring.h
```diff
@@ -20,124 +20,140 @@
...
     };
 
-  } // namespace core
-} // namespace forte
+    struct SFBMonitoringEntry {
+        SFBMonitoringEntry(std::string paFullFBName, CFunctionBlock *paFB) : mFullFBName(paFullFBName), mFB(paFB) {
+        }
+
+        const std::string &getFullFBName() {
```
mx990: I believe this can also be const itself?

src/core/monitoring.cpp
```diff
@@ -12,19 +12,234 @@
...
+#include "datatypes/forte_struct.h"
 
 using namespace std::string_literals;
 using namespace forte::core;
+using namespace forte::core::internal;
 
 const std::string cgClosingXMLTag = "\">"s;
 
+namespace {
+  constexpr auto watchEntryComperator = [](const CWatchEntry &paItem, CStringDictionary::TStringId paPortId) {
```
mx990:
```suggestion
  constexpr auto watchEntryComparator = [](const CWatchEntry &paItem, CStringDictionary::TStringId paPortId) {
```

src/core/monitoring.cpp
```diff
@@ -12,19 +12,234 @@
...
+    return paItem.getPortId() < paPortId;
+  };
+
+  constexpr auto monitoringEntryComperator = [](const SFBMonitoringEntry &paItem, const CFunctionBlock *const paFB) {
```
mx990:
```suggestion
  constexpr auto monitoringEntryComparator = [](const SFBMonitoringEntry &paItem, const CFunctionBlock *const paFB) {
```

src/core/monitoring.cpp
```diff
@@ -12,19 +12,234 @@
...
+    auto &dataWatches = paFBMonitoringEntry.mWatchedDataPoints;
+    auto it = std::lower_bound(dataWatches.begin(), dataWatches.end(), paPortId, watchEntryComperator);
+
+    if (it != dataWatches.end() && it->getPortId() == paPortId) {
+      // the data point is already in the watch list
+      return;
+    }
+
+    dataWatches.emplace(it,
+                        CDataWatchEntry(paPortId, paDataVal, paFBMonitoringEntry.getFB().getAbsDataPortNum(paPortId)));
```
mx990: I would suggest to either use `insert` when there is already a (temporary) instance or supply the individual arguments to actually construct the element in-place with `emplace`:
```suggestion
    dataWatches.emplace(it, paPortId, paDataVal, paFBMonitoringEntry.getFB().getAbsDataPortNum(paPortId));
```

src/core/monitoring.cpp
```diff
@@ -12,19 +12,234 @@
...
+  }
+
+  bool removeDataWatch(internal::SFBMonitoringEntry &paFBMonitoringEntry, CStringDictionary::TStringId paPortId) {
+    auto &dataWatches = paFBMonitoringEntry.mWatchedDataPoints;
+    auto it = std::lower_bound(dataWatches.begin(), dataWatches.end(), paPortId, watchEntryComperator);
```
mx990: It may be better to search for the element directly with `std::find_if`:
```suggestion
    auto it = std::find_if(dataWatches.begin(), dataWatches.end(), [paPortId](auto watch) { return watch.getPortId() == paPortId; });
```

mx990 (changes requested):

src/core/monitoring.cpp
azoitl: as I do sorted inserts the vector maintains a sorted state. Using lower bound would take advantage of this. Not sure if this is needed.

mx990: I guess we also have some pretty large blocks out there, so on further thought let's keep it this way.

src/core/monitoring.cpp
azoitl: done

src/core/monitoring.cpp
azoitl: done

src/core/monitoring.cpp
azoitl: done, didn't know this works.

src/core/monitoring.h
azoitl: done

src/core/monitoring.h
azoitl: was not sure, and somehow it is still a kind of reference. but you are right a refernce_wrapper is much better, changed.

src/core/monitoring.h
azoitl: done

src/core/monitoring.h
azoitl: yes, fixed

mx990 (approved):

PR #440, opened 2025-05-06
Removed singlylinkedList from IO infra

azoitl:


src/modules/wagokbus/WagoDeviceController.h
moniwe, line 1: I think a minor issue but, there is a warning for the following methods due to paIndex not being unsinged:
```
bool isSlaveAvailable(int paIndex);
bool checkSlaveType(int paIndex, int paType);
```
maybe paIndex should be size_t instead of int

src/modules/wagokbus/WagoDeviceController.cpp
moniwe, line 1:
```
void WagoDeviceController::addSlaveHandle(int, forte::core::io::IOHandle *paHandle) {
  CCriticalRegion criticalRegion(mHandleMutex);
  paHandle->isInput() ? mInputHandles.pushBack(paHandle) : mOutputHandles.pushBack(paHandle);
}
```

`pushBack(paHandle)` causes an error, due to `using THandleList = std::vector<std::unique_ptr<IOHandle>>;`

src/modules/revolutionPi/RevPiController.cpp
moniwe, line 1:
```
void RevPiController::addSlaveHandle(int, IOHandle *paHandle) {
  CCriticalRegion criticalRegion(mHandleMutex);
  paHandle->isInput() ? mInputHandles.pushBack(paHandle) : mOutputHandles.pushBack(paHandle);
}
```

I think `pushBack` will cause the same problems than within the `WagoDeviceController`.

src/modules/wagokbus/WagoDeviceController.h
azoitl: I agree but I  think this should be a separate commit as this would need changes through out the IODeviceMultiController and all its children and using classes. Do you want to do that?

src/modules/wagokbus/WagoDeviceController.cpp
azoitl: thx for finding this. Fixed.

src/modules/revolutionPi/RevPiController.cpp
azoitl: thx for finding this. Fixed.

src/core/io/configFB/io_controller_split.cpp
```diff
@@ -30,18 +30,13 @@ IOConfigFBSplitController::IOConfigFBSplitController(const TForteUInt8 *const pa
...
-  TControllerList::Iterator itEnd = mInstances.end();
-  int i = 0;
-  for (TControllerList::Iterator it = mInstances.begin(); it != itEnd; ++it, i++) {
-    if (paId == i && *it != nullptr) {
-      return *it;
-    }
-  }
-  return nullptr;
+  auto it = std::find_if(mInstances.begin(), mInstances.end(),
+                         [&](const IOConfigFBSplitController *obj) { return obj->mId == paId; });
```
cochicde: you just need `paId` inside the lambda, so no need to catch all by reference.

src/core/io/configFB/io_master_multi.cpp
```diff
@@ -18,26 +18,21 @@ using namespace forte::core::io;
...
-  TMasterList::Iterator itEnd = mInstances.end();
-  int i = 0;
-  for (TMasterList::Iterator it = mInstances.begin(); it != itEnd; ++it, i++) {
-    if (paId == i && *it != nullptr) {
-      return *it;
-    }
-  }
-  return nullptr;
+  auto it = std::find_if(mInstances.begin(), mInstances.end(),
+                         [&](const IOConfigFBMultiMaster *obj) { return obj->mId == paId; });
```
cochicde: same here

src/core/io/configFB/io_controller_split.cpp
azoitl: Thx for pointing that out. I'm using these kind of clean-up for brushing up my C++ and missed it. Fixed now

src/core/io/configFB/io_master_multi.cpp
azoitl: Also fixed here.

moniwe (approved):

PR #443, opened 2025-05-09
Marked CSinglyLinkedListNode as Deprecated

azoitl:
With good C++ support on all of our platforms doing containers by ourself is not sensible anymore. Now code should therefor not use our CSinglyLinkedListNode. Also existing code should be gradualy moved to STL containers.

As part of this fix not used includes of this class.


mx990 (approved):

MartinMelikMerkumians (approved):
Good old Singly Linked List served us well for over 15 years

PR #444, opened 2025-05-12
Add support for non-finite values in to/fromString of REAL and LREAL

mx990:
There was a check to ensure that the parsed value is finite in the fromString method of REAL and LREAL. This meant that no infinite or Nan values were accepted. This removes the checks to also allow infinite and NaN values to be parsed.


mx990:
I additionally had to change `fromString` of `REAL` to use `strtof` on all platforms, since the replacement of casting the result of `strtod` does not work properly when allowing for non-finite values.

mx990:
I further fixed the *normalization*, that is, adding a decimal point where necessary, to not add one for non-finite values.

azoitl:
could you add a fromstring test with Nan and Inf?

mx990:
> could you add a fromstring test with Nan and Inf?

done

azoitl (approved):

Issue #392, opened 2025-03-27
[Proposal] Rework FB Interface Spec Data Handling


azoitl:
We are currently working on the final clean-ups and API breaking changes for the Eclipse 4diac 3.0 release. As there are anyhow major changes for 4diac FORTE happening I was reconsidering the way how we store the data for FB interface specs. As we have here many raw pointers this was always a place which I was not happy about. 

The first proposal I have is to store all the arrays with the interface data in `std::array`s and move them out of the class so that they are only in the cpp file of a FB. For a FB (I took `E_CTU` as example) this would then look like as follows:

```
namespace FORTE_E_CTU_data {
  const auto cEventInputNames    = std::array{STRID(CU), STRID(R)};
  const auto cEventInputTypeIds  = std::array{STRID(Event), STRID(Event)};
  const auto cEIWith             = std::array<const TDataIOID, 2>{0, CFunctionBlock::scmWithListDelimiter};
  const auto cEIWithIndexes      = std::array<const TForteInt16, 2>{0, -1};
  const auto cEventOutputNames   = std::array{STRID(CUO), STRID(RO)};
  const auto cEventOutputTypeIds = std::array{STRID(Event), STRID(Event)};
  const auto cEOWith             = std::array<const TDataIOID, 6>{0, 1, CFunctionBlock::scmWithListDelimiter, 0, 1, CFunctionBlock::scmWithListDelimiter};
  const auto cEOWithIndexes      = std::array<const TForteInt16, 2>{0, 3};
  const auto cDataInputNames     = std::array{STRID(PV)};
  const auto cDataInputTypeIds   = std::array{STRID(UINT)};
  const auto cDataOutputNames    = std::array{STRID(Q), STRID(CV)};
  const auto cDataOutputTypeIds  = std::array{STRID(BOOL), STRID(UINT)};
}
```

This change would still be fully backwards compatible and could just be done in the exporter. The initialisation of the interface spec struct would then be:

```
const SFBInterfaceSpec FORTE_E_CTU::scmFBInterfaceSpec = {
  FORTE_E_CTU_data::cEventInputNames.size(), FORTE_E_CTU_data::cEventInputNames.data(), FORTE_E_CTU_data::cEventInputTypeIds.data(), FORTE_E_CTU_data::cEIWith.data(), FORTE_E_CTU_data::cEIWithIndexes.data(),
[...]

PR #454, opened 2025-05-16
Rework CommFB to test how future GenFB API could look like.

azoitl:
This is  a first step to see how we could rework generic FB generation improving especially input output data handling. I started with CommFB as one of our most important generic FB. However this is a bit special so I don't find it to conclusive. I would do more FBs as separate commits on top of this. 

Feedback is always welcome.


cochicde:
I'm having trouble understanding what actually changed or what is the problem with the current implementation. Also, it seems for me more complicated and the generic generations seems to be more dispersed. 

`DEV_MGR::createGenInputData` is allocating memory which I assume its deallocated in the parent, but I don't think is clean to do it like this.

Is `commfb` now offering a default implementation of `createGenInputData` but not used for example by `DEV_MGR`?

azoitl:
@cochicde sorry it may was a bit early to share this pr. The main reason for this PR is to do more preperational work for #392. Because when I started working on #392 I noticed that the way we are currently representing data type names and create the pins for generic fbs is not supporting #392. With this PR I'm trying to move some of the data driven part of gen fbs to the sub-classes. This will result in more code in the subclasses but will simplify interface spec generation and also allows more optimized generation of the members. 

I started with comm fb because it is one of the more important gen FBs. During the work I noticed that I did quite some special optimizations with DEV_MGR. So it is not an ideal example for the goals I had in mind. 

Let me try I few more gen fbs to see where this is leading. Because currently I'm not happy with the datastructures for the data pins and connections. But my vision is that we can move some of the code I#m developing to gen fb as common infrastructure for all gen fbs.

src/core/cominfra/commfb.h
```diff
@@ -57,6 +81,38 @@ namespace forte {
...
+        CIEC_ANY *getDO(size_t) override;
+        CEventConnection *getEOConUnchecked(TPortId) override;
+        CDataConnection **getDIConUnchecked(TPortId) override;
+        CDataConnection *getDOConUnchecked(TPortId) override;
+
+        void createGenInputData() override;
+        void createGenOutputData() override;
+
+        std::unique_ptr<CIEC_ANY *[]> mGenDIs;
+        std::unique_ptr<CIEC_ANY *[]> mGenDOs;
```
mx990, line 105: How about using a `std::vector<std::unique_ptr<CIEC_ANY>>` instead?

Maybe we could even use a `std::vector<CIEC_ANY_VARIANT>` with some additional changes? If we need a particular type in a subclass, could we not just typify them by assigning a value or using `CIEC_ANY_VARIANT::setDefaultValue`?

src/core/genfb.h
```diff
@@ -156,24 +145,77 @@ class CGenFunctionBlock : public T {
...
+      if (getGenDONums()) {
+        mGenDOConns.reserve(getGenDONums());
+        for (size_t i = getGenDOOffset(); i < getGenInterfaceSpec().mNumDOs; ++i) {
+          mGenDOConns.emplace_back(*this, i, *getDO(i));
+        }
+      }
+    }
+
+    virtual size_t getGenEONums() {
+      return getGenInterfaceSpec().mNumEOs;
```
mx990: Is this not *always* `getGenInterfaceSpec().mNumEOs - getGenEOOffset()`?

src/core/genfb.h
```diff
@@ -156,24 +145,77 @@ class CGenFunctionBlock : public T {
...
+        }
+      }
+    }
+
+    void createGenDICons() {
+      if (getGenDINums()) {
+        mGenDIConns = std::unique_ptr<CDataConnection *[]>(new CDataConnection *[getGenDINums()]);
+        for (std::size_t i = 0; i < getGenDINums(); ++i) {
+          mGenDIConns[i] = nullptr;
+        }
```
mx990: I would suggest using the array overload of `std::make_unique`, which also performs value initialization[^1]:
```suggestion
        mGenDIConns = std::make_unique<CDataConnection *[]>(getGenDINums());
```

[^1]: section 2 of https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

src/modules/IEC61131-3/Arithmetic/GEN_FORTE_F_MOVE_fbt.cpp
```diff
@@ -53,30 +59,30 @@ void GEN_FORTE_F_MOVE::executeEvent(TEventID paEIID, CEventChainExecutionThread
...
+    writeData(1 + 0, *mDOs[0], mGenDOConns[0]);
   }
 }
 
 bool GEN_FORTE_F_MOVE::createInterfaceSpec(const char *paConfigString, SFBInterfaceSpec &paInterfaceSpec) {
-  const CTypeLib::CTypeEntry *poToCreate;
   if (strcmp(paConfigString, "F_MOVE") == 0) {
-    poToCreate = &CIEC_ANY_VARIANT::csmFirmwareDataTypeEntry_ANY_VARIANT;
+    mIn = std::unique_ptr<CIEC_ANY>(new CIEC_ANY_VARIANT());
+    mOut = std::unique_ptr<CIEC_ANY>(new CIEC_ANY_VARIANT());
```
mx990:
```suggestion
    mIn = std::make_unique<CIEC_ANY_VARIANT>();
    mOut = std::make_unique<CIEC_ANY_VARIANT>();
```

src/modules/IEC61131-3/Arithmetic/GEN_FORTE_F_MOVE_fbt.cpp
```diff
@@ -53,30 +59,30 @@ void GEN_FORTE_F_MOVE::executeEvent(TEventID paEIID, CEventChainExecutionThread
...
   if (paEIID == scmEventREQID) {
-    readData(0, *mDIs[0], mDIConns[0]);
+    readData(0, *mIn, conn_IN);
   }
 }
 
 void GEN_FORTE_F_MOVE::writeOutputData(TEventID paEOID) {
   if (paEOID == scmEventCNFID) {
-    writeData(1 + 0, *mDOs[0], mDOConns[0]);
+    writeData(1 + 0, *mDOs[0], mGenDOConns[0]);
```
mx990:
```suggestion
    writeData(1 + 0, mOut, mGenDOConns[0]);
```

src/core/cominfra/commfb.cpp
```diff
@@ -57,10 +58,24 @@ const CStringDictionary::TStringId CCommFB::scmEventOutputTypeIds[2] = {STRID(Ev
...
+    conn_QI(nullptr),
+    conn_ID(nullptr),
+    conn_QO(*this, 0, var_QO),
+    conn_STATUS(*this, 1, var_STATUS) {
 }
 
-CCommFB::~CCommFB() = default;
+CCommFB::~CCommFB() {
+  for (size_t i = 0; i < getFBInterfaceSpec().mNumDIs - 2; ++i) {
+    std::destroy_at(mGenDIs[i]);
```
mx990: You should use `delete` instead, since you create them with `new` below. `std::destroy_at` will just call the destructor and not free the underlying memory.

src/core/cominfra/commfb.cpp
```diff
@@ -114,13 +129,15 @@ void CCommFB::executeEvent(TEventID paEIID, CEventChainExecutionThread *const pa
...
+      readData(1, var_ID, conn_ID);
       break;
     }
     case scmSendNotificationEventID: {
-      for (TPortId i = 0; i < getFBInterfaceSpec().mNumDIs; ++i) {
-        readData(i, *mDIs[i], mDIConns[i]);
+      readData(0, var_QI, conn_QI);
+      readData(1, var_ID, conn_ID);
+      for (TPortId i = 2; i < getFBInterfaceSpec().mNumDIs; ++i) {
+        readData(i, *mGenDIs[i], mGenDIConns[i]);
```
mx990: I believe those arrays start at `0` instead of `2`?

src/core/genfb.h
```diff
@@ -243,12 +292,11 @@ bool CGenFunctionBlock<T>::configureFB(const char *paConfigString) {
...
 namespace {
 
-  void genericInitializeDataPin(const CStringDictionary::TStringId *&paDataTypeIds, CIEC_ANY &paPin) {
-    TForteByte *varsData = nullptr;
-    CIEC_ANY *value = CTypeLib::createDataPoint(paDataTypeIds, varsData);
-    if (value != nullptr) {
+  void genericInitializeDataPin(CIEC_ANY &paPin) {
+    if (auto value = std::unique_ptr<CIEC_ANY>(CTypeLib::createDataTypeInstance(paPin.getTypeNameID(), nullptr));
+        value) {
+      // TODO handle array
```
mx990: Maybe we should consider adding a virtual method `reset()` to `CIEC_ANY` that re-initializes the value? That way, we could also avoid creating a temporary instance on the heap.

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -44,22 +43,29 @@ void GEN_STRUCT_DEMUX::executeEvent(TEventID paEIID, CEventChainExecutionThread
...
                                    forte::core::CFBContainer &paContainer) :
-    CGenFunctionBlock<CFunctionBlock>(paContainer, paInstanceNameId) {
+    CGenFunctionBlock<CFunctionBlock>(paContainer, paInstanceNameId),
+    conn_CNF(*this, 0),
+    conn_IN(nullptr) {
+}
+
+GEN_STRUCT_DEMUX::~GEN_STRUCT_DEMUX() {
+  for (size_t i = 0; i < getFBInterfaceSpec().mNumDIs; ++i) {
+    std::destroy_at(mGenDOs[i]);
```
mx990:
```suggestion
    delete mGenDOs[i];
```

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -106,23 +112,20 @@ bool GEN_STRUCT_DEMUX::createInterfaceSpec(const char *paConfigString, SFBInterf
   }
 
   // we could find the struct
-  auto structInstance = static_cast<CIEC_STRUCT *>(data.get());
+  var_IN = std::unique_ptr<CIEC_STRUCT>(static_cast<CIEC_STRUCT *>(data.release()));
```
mx990, line 90: I would suggest directly assigning to `var_IN` above and removing `data`. Otherwise, please move assign the `unique_ptr` instead:
```suggestion
  var_IN = std::move(data);
```

mx990 (changes requested):

src/core/cominfra/commfb.h
azoitl: > How about using a std::vector<std::unique_ptr<CIEC_ANY>> instead?

was not sure regarding the overhead of vectors but I'm open. I think the current design is here rather flexible.

> Maybe we could even use a std::vector<CIEC_ANY_VARIANT> with some additional changes? If we need a particular type in a subclass, could we not just typify them by assigning a value or using CIEC_ANY_VARIANT::setDefaultValue?

The reason I didn't do that for now is not the subclasses but this would require an update to all communication layers. I wanted to avoid this at this point.

src/core/genfb.h
azoitl: Yes. Interesting point. Subclasses that want to indicate that there are no generic interface elements would then need to implement `getGenEOOffset() { return getGenInterfaceSpec().mNumEOs; }`. Would this be confusing?

src/core/genfb.h
azoitl: I like that a lot. This could also make some initialize methods cleaner.

But I guess this would then be a separate commit/PR or?

azoitl:
@mx990 thx for your feedback this helps a lot. Do you think that I'm in general on track with this? I was very ensure about the specific data structures the number of allocations and their overhead. Codewise I like it.

src/core/cominfra/commfb.h
mx990: That makes sense. We can always change that in a future PR.

src/core/genfb.h
mx990: In that case, there is a fixed number of pins, so they could just return that (literal) value.

src/core/genfb.h
mx990: Yes, we should probably separate that into its own PR and then rebase this one.

mx990:
> Do you think that I'm in general on track with this?

I think so, yes. I believe this gives us the opportunity to optimize for a particular FB, while retaining the flexibility of the original implementation.

> I was very ensure about the specific data structures the number of allocations and their overhead.

We certainly lose some locality since we no longer allocate one large chunk in advance. We also gain some locality since we now have "fixed" variables directly in the FB. The latter, in particular, saves us several indirect lookups.

In terms of memory overhead, these extra allocations will certianly cost some additional space. However, there are also cases where we save space. `F_MOVE`, for example, no longer needs arrays of pointers with length one for its single data input and output.

src/modules/convert/GEN_STRUCT_DEMUX_fbt.h
```diff
@@ -51,26 +64,24 @@ class GEN_STRUCT_DEMUX : public CGenFunctionBlock<CFunctionBlock> {
...
     std::vector<std::string_view> getConfiguredMemberNames(std::string_view paMemberNameString);
-    void setConfiguredDOPorts();
     CIEC_ANY *getNestedMember(const CStringDictionary::TStringId paNameId, CIEC_STRUCT *paStructType);
-    size_t calcConfiguredStructTypeNameSize(CIEC_STRUCT *paStructType,
-                                            std::vector<std::string_view> &paConfiguredMemberNames);
+
+    CEventConnection conn_CNF;
+    std::unique_ptr<CIEC_STRUCT> var_IN;
+    CDataConnection *conn_IN;
+    std::unique_ptr<CIEC_ANY *[]> mGenDOs;
```
mx990: Instead of allocating the struct members individually and copying from the struct, could we maybe just directly return the members of `var_IN` in `getDO`?

src/core/genfb.h
```diff
@@ -156,24 +145,77 @@ class CGenFunctionBlock : public T {
     static size_t getDataPointSpecSize(const CIEC_ANY &paValue);
     static void fillDataPointSpec(const CIEC_ANY &paValue, CStringDictionary::TStringId *&paDataTypeIds);
 
-    static size_t calculateFBConnDataSize(const SFBInterfaceSpec &paInterfaceSpec);
+    virtual void setupFBInterface();
 
-    static size_t calculateFBVarsDataSize(const SFBInterfaceSpec &paInterfaceSpec);
+    virtual void createGenInputData() {
+      // TODO turn into pure virtual as soon as all gen fbs are converted
```
cochicde: maybe it's fine as it is now, since not all subclasses have intput AND output generic data, right? Otherwise you'll also spread a lot of empty implementations in all generic FBs

src/core/genfb.h
```diff
@@ -368,98 +413,20 @@ void CGenFunctionBlock<T>::fillDataPointSpec(const CIEC_ANY &paValue, CStringDic
...
 template<class T>
 void CGenFunctionBlock<T>::setupFBInterface() {
-  freeFBInterfaceData();
 
-  const SFBInterfaceSpec &interfaceSpec = T::getFBInterfaceSpec();
-  size_t connDataSize = calculateFBConnDataSize(interfaceSpec);
-  size_t varsDataSize = calculateFBVarsDataSize(interfaceSpec);
-  mFBConnData = connDataSize ? operator new(connDataSize) : nullptr;
-  mFBVarsData = varsDataSize ? operator new(varsDataSize) : nullptr;
+  createGenEOCons();
```
cochicde, line 394: is there a special reason to have all these separated functions? couldn't you just have one and the subclass takes care then of doing what it needs?

at least in `commfb` `createGenInputData` and `createGenOutputData` it seems that one function would be enough

cochicde (commented):
I haven't look into the connection code for a while, so I'm not up to date with the latest implementation, so my view here is more from an outsider perspective. 

But I have to say this seems more complicated and I see a lot of duplicated code. The subclasses are doing a lot more than they were doing before, which is against maintainability of the code. 

It's also confusing what is the difference between `createInterfaceSpec` and the new methods such as `createGenInputData`. 

as the new proposed structure from #392 is still not used, it's hard to see what's been fixed or improved with this PR.

But as said, this is from an outsider view. If you think the improvements compensate the negatives side, I trust your expertise, so my only comment would be: are you 100% sure this is the right approach?

P.S.: If you want to avoid large PRs and changing a lot of files, you could create a parallel `genFB` class and have both classes live together for a while and make  the "testing" subclasses derive from these new parallel class. That way, the code will still compile and you'll get feedback also from the CI

src/core/genfb.h
azoitl: I think you are right here. I'll remove the comment

src/core/genfb.h
azoitl: Yes in many subclasses only one of the functions need to be override. In some even none. Also there is a required order of these methods. With this template method pattern subclasses are forced into a certain structre.

azoitl:
@cochicde thx for your feedback. Yes it moveing some of the generic code to the respective subclasses. Themain reason is that the subclasses know better how certain stuff should be done. Even some elements now can be members of classes and doe not have to created dynamically as before (i.e., this are very often output event connections. This also allows subclasses to be implemented more efficiently. As you will see with the new versions of F_MOVE and struct mux that I'll do in a minute. 

the interface spec is the information needed to create connections and to perform monitoring. The other function are now handling the data creation for the FB. 

> But as said, this is from an outsider view. If you think the improvements compensate the negatives side, I trust your expertise, so my only comment would be: are you 100% sure this is the right approach?

When I started I was not sure. Now that I'm further I think it is the right way. Many gen FBs can be implemented better cleaner and more important better optimizeable for the compiler.  But with the tradeoff that more code in more functions has to be written.

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
azoitl: no I can't use std::move because I need to cast here. Doing this where data is created leads for non struct types to wrong results.

mx990: Ah, I somehow overlooked the cast in there.

src/core/genfb.h
```diff
@@ -368,98 +389,20 @@ void CGenFunctionBlock<T>::fillDataPointSpec(const CIEC_ANY &paValue, CStringDic
...
-  }
-
-  const CStringDictionary::TStringId *datarTypeIds;
-  if (interfaceSpec.mNumDIs) {
-    mDIConns = reinterpret_cast<CDataConnection **>(connData);
-    connData += sizeof(CDataConnection *) * interfaceSpec.mNumDIs;
-
-    mDIs = reinterpret_cast<CIEC_ANY **>(varsData);
-    varsData += interfaceSpec.mNumDIs * sizeof(CIEC_ANY *);
+  createGenInputData();
```
mx990: I believe we should create the data before the connections?

src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
```diff
@@ -27,81 +26,101 @@ USE_STRING_ID(GEN_ADD);
...
-    mDInputs = static_cast<unsigned int>(forte::core::util::strtoul(pcPos, nullptr, 10));
-    DEVLOG_DEBUG("DIs: %d;\n", mDInputs);
-  } else {
+  if (pcPos == nullptr) {
     return false;
   }
 
+  pcPos++;
+  // we have an underscore and it is the first underscore after AND
+  unsigned int mDInputs = static_cast<unsigned int>(forte::core::util::strtoul(pcPos, nullptr, 10));
```
mx990: Please remove the `m` prefix, since this is no longer a member.

src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
```diff
@@ -27,81 +26,101 @@ USE_STRING_ID(GEN_ADD);
...
+
+  return true;
+}
+
+CEventConnection *GEN_ADD::getEOConUnchecked(TPortId paEONum) {
+  return (paEONum == 0) ? &conn_CNF : nullptr;
+}
+
+void GEN_ADD::createGenInputData() {
+  mGenDIs = std::unique_ptr<CIEC_ANY_MAGNITUDE_VARIANT[]>(new CIEC_ANY_MAGNITUDE_VARIANT[getFBInterfaceSpec().mNumDIs]);
```
mx990:
```suggestion
  mGenDIs = std::make_unique<CIEC_ANY_MAGNITUDE_VARIANT[]>(getFBInterfaceSpec().mNumDIs);
```

src/modules/IEC61131-3/Arithmetic/GEN_FORTE_F_MOVE_fbt.cpp
```diff
@@ -39,44 +43,41 @@ const CStringDictionary::TStringId GEN_FORTE_F_MOVE::scmEventOutputTypeIds[] = {
...
   if (paEIID == scmEventREQID) {
-    readData(0, *mDIs[0], mDIConns[0]);
+    readData(0, *mIn, conn_IN);
   }
 }
 
 void GEN_FORTE_F_MOVE::writeOutputData(TEventID paEOID) {
   if (paEOID == scmEventCNFID) {
-    writeData(1 + 0, *mDOs[0], mDOConns[0]);
+    writeData(1 + 0, *mIn, mGenDOConns[0]);
```
mx990: Could you not also define a `std::unique_ptr<CGenDataConnection> conn_OUT` in this class instead of having a separate array with length one?

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -140,32 +117,27 @@ bool GEN_STRUCT_DEMUX::createInterfaceSpec(const char *paConfigString, SFBInterf
...
-  auto doDataTypeNames = mDoDataTypeNames.get();
+void GEN_STRUCT_DEMUX::fillConfiguredInterfaceSpec(std::vector<std::string_view> &paConfiguredMemberNames) {
   for (size_t i = 0; i < paConfiguredMemberNames.size(); ++i) {
     CStringDictionary::TStringId memberNameId =
         CStringDictionary::insert(paConfiguredMemberNames[i].data(), paConfiguredMemberNames[i].length());
     mDoNames[i] = memberNameId;
-    CIEC_ANY *member = getNestedMember(memberNameId, paStructType);
-    fillDataPointSpec(*member, doDataTypeNames);
+    CIEC_ANY *member = getNestedMember(memberNameId, var_IN.get());
+    mConfiguredDOPorts.emplace_back(member);
```
mx990, line 133: I believe we should reserve with `paConfiguredMemberNames.size()` before.

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
```diff
@@ -140,32 +117,27 @@ bool GEN_STRUCT_DEMUX::createInterfaceSpec(const char *paConfigString, SFBInterf
...
-  auto doDataTypeNames = mDoDataTypeNames.get();
-  for (size_t i = 0; i < paStructType->getStructSize(); ++i) {
-    const CIEC_ANY &member = *paStructType->getMember(i);
-    mDoNames[i] = paStructType->elementNames()[i];
-    fillDataPointSpec(member, doDataTypeNames);
+void GEN_STRUCT_DEMUX::fillInterfaceSpec() {
+  for (size_t i = 0; i < var_IN->getStructSize(); ++i) {
+    CIEC_ANY *member = var_IN->getMember(i);
+    mDoNames[i] = var_IN->elementNames()[i];
+    mConfiguredDOPorts.emplace_back(member);
```
mx990, line 143: same here with `var_IN->getStructSize()`

src/modules/rt_events/RT_Bridge_fbt.cpp
```diff
@@ -80,14 +78,37 @@ bool FORTE_GEN_RT_Bridge::createInterfaceSpec(const char *paConfigString, SFBInt
...
+CIEC_ANY *FORTE_GEN_RT_Bridge::getDO(size_t paIndex) {
+  return &mGenDOs[paIndex];
+}
+
+CEventConnection *FORTE_GEN_RT_Bridge::getEOConUnchecked(TPortId paIndex) {
+  return (paIndex == 0) ? &conn_RDO : nullptr;
+}
+
+void FORTE_GEN_RT_Bridge::createGenInputData() {
+  mGenDIs = std::unique_ptr<CIEC_ANY_VARIANT[]>(new CIEC_ANY_VARIANT[getFBInterfaceSpec().mNumDIs]);
```
mx990:
```suggestion
  mGenDIs = std::make_unique<CIEC_ANY_VARIANT[]>(getFBInterfaceSpec().mNumDIs);
```

src/modules/rt_events/RT_Bridge_fbt.cpp
```diff
@@ -80,14 +78,37 @@ bool FORTE_GEN_RT_Bridge::createInterfaceSpec(const char *paConfigString, SFBInt
...
+}
+
+void FORTE_GEN_RT_Bridge::createGenOutputData() {
+  mGenDOs = std::unique_ptr<CIEC_ANY_VARIANT[]>(new CIEC_ANY_VARIANT[getFBInterfaceSpec().mNumDOs]);
```
mx990:
```suggestion
  mGenDOs = std::make_unique<CIEC_ANY_VARIANT[]>(getFBInterfaceSpec().mNumDOs);
```

src/modules/utils/GEN_ARRAY2VALUES_fbt.cpp
```diff
@@ -58,65 +59,68 @@ void GEN_ARRAY2VALUES::executeEvent(TEventID paEIID, CEventChainExecutionThread
...
+  paInterfaceSpec.mEONames = scmEventOutputNames;
+  paInterfaceSpec.mNumDIs = 1;
+  paInterfaceSpec.mDINames = scmDataInputNames;
+  paInterfaceSpec.mNumDOs = numDOs;
+  paInterfaceSpec.mDONames = mDataOutputNames.get();
+  return true;
+}
+
+void GEN_ARRAY2VALUES::createGenOutputData() {
+  mGenDOs = std::unique_ptr<CIEC_ANY_VARIANT[]>(new CIEC_ANY_VARIANT[getFBInterfaceSpec().mNumDOs]);
```
mx990:
```suggestion
  mGenDOs = std::make_unique<CIEC_ANY_VARIANT[]>(getFBInterfaceSpec().mNumDOs);
```

src/modules/utils/GEN_CSV_WRITER_fbt.cpp
```diff
@@ -218,8 +220,57 @@ void GEN_CSV_WRITER::writeCSVFileLine() {
...
-    DEVLOG_ERROR("[GEN_CSV_WRITER]: Can't write to file %s since it is not opened\n", FILE_NAME().getStorage().c_str());
+    var_QO = false_BOOL;
+    var_STATUS = scmFileNotOpened;
+    DEVLOG_ERROR("[GEN_CSV_WRITER]: Can't write to file %s since it is not opened\n",
+                 var_FILE_NAME.getStorage().c_str());
+  }
+}
+
+void GEN_CSV_WRITER::createGenInputData() {
+  mGenDIs = std::unique_ptr<CIEC_ANY_VARIANT[]>(new CIEC_ANY_VARIANT[getGenDINums()]);
```
mx990:
```suggestion
  mGenDIs = std::make_unique<CIEC_ANY_VARIANT[]>(getGenDINums());
```

src/modules/utils/GEN_VALUES2ARRAY_fbt.cpp
```diff
@@ -41,84 +41,84 @@ GEN_VALUES2ARRAY::GEN_VALUES2ARRAY(const CStringDictionary::TStringId paInstance
...
+  paInterfaceSpec.mEONames = scmEventOutputNames;
+  paInterfaceSpec.mNumDIs = numDIs;
+  paInterfaceSpec.mDINames = mDataInputNames.get();
+  paInterfaceSpec.mNumDOs = 1;
+  paInterfaceSpec.mDONames = scmDataOutputNames;
+  return true;
+}
+
+void GEN_VALUES2ARRAY::createGenInputData() {
+  mGenDIs = std::unique_ptr<CIEC_ANY_VARIANT[]>(new CIEC_ANY_VARIANT[getFBInterfaceSpec().mNumDIs]);
```
mx990:
```suggestion
  mGenDIs = std::make_unique<CIEC_ANY_VARIANT[]>(getFBInterfaceSpec().mNumDIs);
```

tests/core/cominfra/fbdkasn1layerdeser_test.cpp
```diff
@@ -104,9 +104,21 @@ class CDeserTestMockCommFB : public forte::com_infra::CCommFB {
       setupFBInterface();
     }
 
-    virtual ~CDeserTestMockCommFB() {
-      freeFBInterfaceData();
-      getGenInterfaceSpec() = {};
+    virtual ~CDeserTestMockCommFB() = default;
```
mx990:
```suggestion
    ~CDeserTestMockCommFB() override = default;
```

tests/core/cominfra/fbdkasn1layerdeser_test.cpp
```diff
@@ -104,9 +104,21 @@ class CDeserTestMockCommFB : public forte::com_infra::CCommFB {
...
+
+  protected:
+    void createGenOutputData() override {
+      size_t numGenDOs = getFBInterfaceSpec().mNumDOs - 2;
+      const CStringDictionary::TStringId *datarTypeIds = getFBInterfaceSpec().mDODataTypeNames + 2;
+      TForteByte *varsData = nullptr;
+      mGenDOs = std::unique_ptr<CIEC_ANY *[]>(new CIEC_ANY *[numGenDOs]);
```
mx990:
```suggestion
      mGenDOs = std::make_unique<CIEC_ANY *[]>(numGenDOs);
```

mx990 (changes requested):

src/core/genfb.h
azoitl: for input connections it does not matter but for consistency I think it is better. Will change.

src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
azoitl: done

src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
azoitl: done

src/modules/IEC61131-3/Arithmetic/GEN_FORTE_F_MOVE_fbt.cpp
azoitl: done

src/core/genfb.h
azoitl: done

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
azoitl: Done

src/modules/convert/GEN_STRUCT_DEMUX_fbt.cpp
azoitl: Done

src/modules/rt_events/RT_Bridge_fbt.cpp
azoitl: Done

src/modules/rt_events/RT_Bridge_fbt.cpp
azoitl: Done

src/modules/utils/GEN_ARRAY2VALUES_fbt.cpp
azoitl: Done

src/modules/utils/GEN_CSV_WRITER_fbt.cpp
azoitl: Done

src/modules/utils/GEN_VALUES2ARRAY_fbt.cpp
azoitl: Done

tests/core/cominfra/fbdkasn1layerdeser_test.cpp
azoitl: Done

tests/core/cominfra/fbdkasn1layerdeser_test.cpp
azoitl: Done

azoitl:
@mx990 @cochicde  I think now it is in a stage that it could be merged. It is the basis for reworking adapters as well as reworking the typelib which I need for the type hashes.

src/modules/IEC61131-3/Arithmetic/GEN_FORTE_F_MOVE_fbt.cpp
mx990: I believe we should also use `conn_OUT` now, as well as override `getGenDOOffset` and `getDOConUnchecked`?

src/modules/utils/GEN_ARRAY2VALUES_fbt.h
```diff
@@ -18,24 +18,35 @@
 #ifndef _GEN_ARRAY2VALUES_H_
 #define _GEN_ARRAY2VALUES_H_
 
+#include <boost/test/utils/xml_printer.hpp>
```
mx990: It seems the Windows build is failing because of this. As far as I can tell, this is not needed. At the very least, it builds on my machine without it.

mx990 (changes requested):

src/modules/utils/GEN_ARRAY2VALUES_fbt.h
azoitl: I have no clue how that got in. Thx for finding.

src/core/genfb.h
```diff
@@ -368,98 +389,20 @@ void CGenFunctionBlock<T>::fillDataPointSpec(const CIEC_ANY &paValue, CStringDic
...
-    result += CTypeLib::getDataPointSize(dataTypeIds) * 2; // * 2 for connection buffer value
-  }
-
-  result += paInterfaceSpec.mNumAdapters * sizeof(TAdapterPtr);
-  return result;
-}
-
 template<class T>
 void CGenFunctionBlock<T>::setupFBInterface() {
-  freeFBInterfaceData();
```
mx990, line 399: Both `initialize()` and `configureFB(...)` are currently calling `setupFBInterface()`, so we get here and allocate the data twice but no longer free the first round.

azoitl: thx for noting this. I think this should only be done once. I personally would prefere the configureFB because then the FB is in a valid configured state for any other initialisations that may happen.

mx990: Yes, I agree.

mx990 (approved):

PR #455, opened 2025-05-16
`IODeviceController` support for `IOHandle`s  which direction is not set at initialization + Update eLITeBoard

scharf-m:
The current implementation of `IODeviceController` wraps the adds `IOHandle`s into a `std::unique_ptr` and adds them into the controllers `mInputHandles` or `mOutputHandles` std::vector. If the handles direction is neither set to  `IN` nor `OUT`, it will not getting added and the live-time of the unique-pointer expires, which causes a de-registration of the corresponding IO-handle.

To catch this, this PR introduces an additional std::vector. Objects with `UnknownDirection` or `InOut` are now temporarily stored into `mDiverseHandles`.
`IODeviceController::updateHandleList(std::string const &paId, IOHandle *paHandle)` makes it possible to move a handle from the list `mDiverseHandles` into the right vector, according to the currently set direction.

An example of the use is shown in the updated eLITeBoard module. GPIOs are initialized with `UnknownDirection`. After the observer is registered, the IOHandle (see `IOHandleGPIO::onObserver`) updates its direction and calls `updateHandleList`.

### DIFF
--- a/src/core/io/device/io_controller.cpp
+++ b/src/core/io/device/io_controller.cpp
@@ -66,13 +66,46 @@ void IODeviceController::addHandle(HandleDescriptor &paHandleDescriptor) {
                    paHandleDescriptor.mId.c_str());
     return;
   }
+  addHandle(paHandleDescriptor.mId, std::move(handle));
+}
 
-  switch (handle->getDirection()) {
-    case IOMapper::In: addHandle(mInputHandles, paHandleDescriptor.mId, std::move(handle)); break;
-    case IOMapper::Out: addHandle(mOutputHandles, paHandleDescriptor.mId, std::move(handle)); break;
+void IODeviceController::addHandle(std::string const &paId, std::unique_ptr<IOHandle> paHandle) {
+  switch (paHandle->getDirection()) {
+    case IOMapper::In: addHandle(mInputHandles, paId, std::move(paHandle)); break;
+    case IOMapper::Out: addHandle(mOutputHandles, paId, std::move(paHandle)); break;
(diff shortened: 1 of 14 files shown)


src/core/io/device/io_controller.cpp
```diff
@@ -66,14 +66,47 @@ void IODeviceController::addHandle(HandleDescriptor &paHandleDescriptor) {
...
 
+void IODeviceController::updateHandleList(std::string const &paId, IOHandle *paHandle) {
+  for (auto it = mDiverseHandles.begin(); it != mDiverseHandles.end(); ++it) {
+    if (it->get() == paHandle) {
+      auto handleDirection = (*it)->getDirection();
+      switch (handleDirection) {
+        case IOMapper::In: {
+          CCriticalRegion criticalRegion(mHandleMutex);
+          mInputHandles.push_back(std::move(*it));
+          mDiverseHandles.erase(it);
```
azoitl: this will invalidate the iterator. you need to use the return value of erease as the new iterator. But then you can not use ++it in the for loop

src/core/io/device/io_controller.cpp
```diff
@@ -66,14 +66,47 @@ void IODeviceController::addHandle(HandleDescriptor &paHandleDescriptor) {
...
+          break;
+        }
+        case IOMapper::Out: {
+          CCriticalRegion criticalRegion(mHandleMutex);
+          mOutputHandles.push_back(std::move(*it));
+          mDiverseHandles.erase(it);
```
azoitl: same

azoitl (changes requested):

azoitl (approved):

PR #458, opened 2025-05-19
Fix compilation errors and add ctf trace to the github CI

cochicde:


src/core/basefb.cpp
```diff
@@ -147,10 +148,12 @@ void CBasicFB::traceInstanceData() {
...
   }
 
-  getResource()->getTracer().traceInstanceData(
-      getFBTypeName() ?: "null", getFullQualifiedApplicationInstanceName('.').c_str() ?: "null",
-      static_cast<uint32_t>(inputs.size()), inputs_c_str.data(), static_cast<uint32_t>(outputs.size()),
-      outputs_c_str.data(), static_cast<uint32_t>(internals.size()), internals_c_str.data(),
-      static_cast<uint32_t>(internalFbs.size()), internalFbs_c_str.data());
+  auto typeName = getFBTypeName();
+  auto fullName = getFullQualifiedApplicationInstanceName('.').c_str();
+  getResource()->getTracer().traceInstanceData(typeName ? typeName : "null", fullName ? fullName : "null",
```
mx990: Should we change `getFBTypeName()` to never return a nullptr? Maybe we should instead return an empty string if the type name is invalid or not set?

Furthermore, I believe `fullName` can never be `nullptr`?

mx990 (changes requested):

src/core/basefb.cpp
azoitl: I like this idea. But I think this should be then done in a separate PR. @cochicde would you like to look into that?

cochicde: yes, that makes sense. I will update this PR with the check against `nullptr` of `fullName` and then create another PR that takes care of `getFBTypeName` so it returns something nicer

mx990 (approved):

cochicde:
I ran the tests and there was an error when tracing the output values.

azoitl (approved):

azoitl (approved):

cochicde:
thank you for fixing the issues. my compiler suddenly got old when the `#include <format>` arrived to the developed branch

azoitl:
> thank you for fixing the issues. my compiler suddenly got old when the `#include <format>` arrived to the developed branch

happy to support. Maybe you are the prove that our C++20 decision was still a bit to soon.

PR #459, opened 2025-05-19
Silence C++20 codecvt facets deprecation warnings on MSVC

mx990:
The codecvt facets have been deprecated in C++20 and there is currently no viable replacement in the standard library.
    
This silences the deprecation warnings on MSVC to prevent unnecessary repeated warnings during builds until we can find a viable replacement.


cochicde:
why silence it instead of  fixing it?

mx990:
If you know of a quick way to fix it, please let me know. As far as I am aware, there is currently no viable alternative in the standard C++ library.

After discussing the problem a while ago with @azoitl and @jwalt, we were considering implementing a replacement ourselves. In the meantime, I just wanted to keep windows builds from spamming the log with repeated warnings.

cochicde:
no no, sorrry I read my comment again now and it sounds bad. I meant mainly to have the reason as comment on why it was decided to supress instead of fixing it, so we don't ask ourselves in the future whenever we are cleaning up

mx990:
Ah, thank you for clarifying. I have expanded on the comment and added a description to the commit message and this PR.

cochicde (approved):

PR #464, opened 2025-05-21
Added reset method that allows to reset pins to their type default value

azoitl:
As part of this work also for all variant resets this is used.

Furthermore the include files for all data types have been cleand. This should improve compilation a bit.

Currently the FB tests are failing for FBs with generic in or outputs. The reason is that the FB test is creating a fresh instance and checks if the the reseted Test FBs has the same io values. With the new reset of variants this is not holding anymore. I'm not sure how to best solve this problem.

### DIFF
--- a/src/core/datatypes/forte_array.cpp
+++ b/src/core/datatypes/forte_array.cpp
@@ -25,10 +25,9 @@
  *******************************************************************************/
 
 #include "forte_array.h"
+#include "forte_ulint.h"
 
 USE_STRING_ID(ARRAY);
-
-#include "forte_ulint.h"
 
 CStringDictionary::TStringId CIEC_ARRAY::getTypeNameID() const {
   return STRID(ARRAY);
@@ -37,6 +36,12 @@ CStringDictionary::TStringId CIEC_ARRAY::getTypeNameID() const {
 void CIEC_ARRAY::setValue(const CIEC_ANY &paValue) {
   if (paValue.getDataTypeID() == CIEC_ANY::e_ARRAY) {
     operator=(static_cast<const CIEC_ARRAY &>(paValue));
+  }
+}
+
+void CIEC_ARRAY::reset() {
+  for (intmax_t i = getLowerBound(), end = getUpperBound(); i <= end; ++i) {
+    operator[](i).reset();
   }
 }
 
--- a/src/core/datatypes/forte_date_and_time.cpp
+++ b/src/core/datatypes/forte_date_and_time.cpp
@@ -19,6 +19,8 @@
 #include <string.h>
 #include <ctype.h>
 #include "forte_date_and_time.h"
+#include "forte_architecture_time.h"
+#include "string_utils.h"
 
 USE_STRING_ID(DATE_AND_TIME);
 
--- a/src/core/datatypes/forte_ldate_and_time.cpp
+++ b/src/core/datatypes/forte_ldate_and_time.cpp
@@ -19,6 +19,8 @@
 #include <string.h>
 #include <ctype.h>
 #include "forte_ldate_and_time.h"
+#include "forte_architecture_time.h"
+#include "string_utils.h"
 
 USE_STRING_ID(LDATE_AND_TIME);
 
--- a/src/core/datatypes/forte_struct.cpp
+++ b/src/core/datatypes/forte_struct.cpp
@@ -39,7 +39,7 @@ CIEC_ANY *CIEC_STRUCT::getMemberNamed(const char *paMemberName) {
 
 size_t CIEC_STRUCT::getMemberIndex(CStringDictionary::TStringId paMemberNameId) {
   const CStringDictionary::TStringId *punMemberNameIds = elementNames();
-  for (size_t i = 0; i < getStructSize(); ++i) {
+  for (size_t i = 0, structSize = getStructSize(); i < structSize; ++i) {
     if (punMemberNameIds[i] == paMemberNameId) {
       return i;
     }
@@ -51,11 +51,16 @@ void CIEC_STRUCT::setValue(const CIEC_ANY &paValue) {
   if (paValue.getDataTypeID() == e_STRUCT) {
     auto &otherStruct = static_cast<const CIEC_STRUCT &>(paValue);
     if (getStructTypeNameID() == otherStruct.getStructTypeNameID()) {
-      size_t structSize = getStructSize();
-      for (size_t i = 0; i < structSize; ++i) {
+      for (size_t i = 0, structSize = getStructSize(); i < structSize; ++i) {
         getMember(i)->setValue(*otherStruct.getMember(i));
       }
     }
+  }
+}
+
+void CIEC_STRUCT::reset() {
+  for (size_t i = 0, structSize = getStructSize(); i < structSize; ++i) {
+    getMember(i)->reset();
   }
 }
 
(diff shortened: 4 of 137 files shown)


src/core/datatypes/forte_array.cpp
```diff
@@ -40,6 +39,12 @@ void CIEC_ARRAY::setValue(const CIEC_ANY &paValue) {
   }
 }
 
+void CIEC_ARRAY::reset() {
+  for (intmax_t i = getLowerBound(), end = getUpperBound(); i <= end; ++i) {
+    operator[](i).reset();
+  }
+}
+
```
mx990, line 47: I would suggest also adding a specialization in `CIEC_ARRAY_FIXED`, where we can take advantage of the known bounds at compile time.

src/core/datatypes/forte_date_and_time.cpp
```diff
@@ -19,6 +19,8 @@
 #include <string.h>
 #include <ctype.h>
 #include "forte_date_and_time.h"
+#include <forte_architecture_time.h>
```
mx990:
```suggestion
#include "forte_architecture_time.h"
```

src/core/datatypes/forte_ldate_and_time.cpp
```diff
@@ -19,6 +19,8 @@
 #include <string.h>
 #include <ctype.h>
 #include "forte_ldate_and_time.h"
+#include <forte_architecture_time.h>
```
mx990:
```suggestion
#include "forte_architecture_time.h"
```

src/core/datatypes/forte_struct.cpp
```diff
@@ -59,6 +59,12 @@ void CIEC_STRUCT::setValue(const CIEC_ANY &paValue) {
   }
 }
 
+void CIEC_STRUCT::reset() {
+  for (size_t i = 0; i < getStructSize(); ++i) {
```
mx990: I would suggest assigning `getStructSize()` to a local variable to keep it from getting called on every iteration:
```suggestion
  for (size_t i = 0, size = getStructSize(); i < size; ++i) {
```

src/core/datatypes/forte_time_of_day.cpp
```diff
@@ -19,6 +19,8 @@
 #include <string.h>
 #include <ctype.h>
 #include "forte_time_of_day.h"
+#include <forte_architecture_time.h>
```
mx990:
```suggestion
#include "forte_architecture_time.h"
```

mx990 (changes requested):
I believe you are missing a delegating `reset()` in `CIEC_ANY_SIGNED_VARIANT` and `CIEC_ANY_UNSIGNED_VARIANT`.

mx990:
> Currently the FB tests are failing for FBs with generic in or outputs. The reason is that the FB test is creating a fresh instance and checks if the the reseted Test FBs has the same io values. With the new reset of variants this is not holding anymore. I'm not sure how to best solve this problem.

I would suggest the following to typify the inputs (same for outputs) of the fresh instance before comparing them:
```cpp
  for (size_t i = 0; i < interfaceSpec.mNumDIs; ++i) {
    CIEC_ANY *testVar = mFBUnderTest->getDI(i);
    CIEC_ANY *freshVar = freshInstance->getDI(i);
    if (freshVar->getDataTypeID() == CIEC_ANY::e_ANY) {
      if (auto tempVar = std::unique_ptr<CIEC_ANY>(testVar->unwrap().clone(nullptr))) {
        tempVar->reset(); // reset the CIEC_ANY not the unique_ptr
        freshVar->setValue(*tempVar);
      }
    }
  }
```

azoitl:
> > Currently the FB tests are failing for FBs with generic in or outputs. The reason is that the FB test is creating a fresh instance and checks if the the reseted Test FBs has the same io values. With the new reset of variants this is not holding anymore. I'm not sure how to best solve this problem.
> 
> I would suggest the following to typify the inputs (same for outputs) of the fresh instance before comparing them:
> 
> ```c++
>   for (size_t i = 0; i < interfaceSpec.mNumDIs; ++i) {
>     CIEC_ANY *testVar = mFBUnderTest->getDI(i);
>     CIEC_ANY *freshVar = freshInstance->getDI(i);
>     if (freshVar->getDataTypeID() == CIEC_ANY::e_ANY) {
>       if (auto tempVar = std::unique_ptr<CIEC_ANY>(testVar->unwrap().clone(nullptr))) {
>         tempVar->reset(); // reset the CIEC_ANY not the unique_ptr
>         freshVar->setValue(*tempVar);
>       }
>     }
>   }
> ```

@mx990 I also thought about something along this lines. However I think that this will not be correct in all cases. If we have FBs with the retain flag set these test would also fail. Therefore I'm strongly considering to remove the variable checks as a whole.  Also I'm not sure if your code above is testing anything or just repeating the code that we have in the FBs.

mx990:
> However I think that this will not be correct in all cases. If we have FBs with the retain flag set these test would also fail. Therefore I'm strongly considering to remove the variable checks as a whole.

Maybe we could also make `performFBResetTests()` virtual so individual tests could disable this step or account for the retain flags of their FBs in the future? That said, I am also open to removing them.

> Also I'm not sure if your code above is testing anything or just repeating the code that we have in the FBs.

At the very least, if I comment one of the resets in `setInitialValues()` of an FB under test, the test fails and otherwise succeeds.

azoitl:
> > However I think that this will not be correct in all cases. If we have FBs with the retain flag set these test would also fail. Therefore I'm strongly considering to remove the variable checks as a whole.
> 
> Maybe we could also make `performFBResetTests()` virtual so individual tests could disable this step or account for the retain flags of their FBs in the future? That said, I am also open to removing them.
> 
> > Also I'm not sure if your code above is testing anything or just repeating the code that we have in the FBs.
> 
> At the very least, if I comment one of the resets in `setInitialValues()` of an FB under test, the test fails and otherwise succeeds.

That are both good points. Will go with both.

azoitl:
Virtual will not work. This is handled in the destructor.

src/core/datatypes/forte_array.cpp
azoitl: done

src/core/datatypes/forte_date_and_time.cpp
azoitl: done

src/core/datatypes/forte_ldate_and_time.cpp
azoitl: done

src/core/datatypes/forte_struct.cpp
azoitl: done, also checked all other for loops in that file.

src/core/datatypes/forte_time_of_day.cpp
azoitl: done

azoitl:
> I believe you are missing a delegating `reset()` in `CIEC_ANY_SIGNED_VARIANT` and `CIEC_ANY_UNSIGNED_VARIANT`.

yes I did and accidentally added it to a wrong file.  Found with the proposed change to the FB tester so I think it is good to have it.

mx990:
> Virtual will not work. This is handled in the destructor.

Ah, did not notice that. There are some ways around that[^1], but I would say this can wait until we actually run into a problem.

[^1]: e.g., https://en.cppreference.com/w/cpp/language/crtp

mx990 (approved):

PR #467, opened 2025-05-22
Remove SinglyListLink usage from opcua layer

cochicde:
Some logic was also cleanup, mainly because there were memory problems which were easier to fix by cleaning up the logic.

Some things had to be changed to allow erasing elements from the vector, and to properly handle some pointers being passed to the library, some vector of objects had to be chaned to a vector of unique pointers

### DIFF
--- a/src/com/opc_ua/opcua_local_handler.cpp
+++ b/src/com/opc_ua/opcua_local_handler.cpp
@@ -223,12 +223,9 @@ void COPC_UA_Local_Handler::referencedNodesIncrement(const std::vector<UA_NodeId
       }
     }
     if (!found) {
-      nodesReferencedByActions newRef;
       UA_NodeId *newNode = UA_NodeId_new();
       UA_NodeId_copy(node, newNode);
-      newRef.mNodeId = newNode;
-      newRef.mActionsReferencingIt.push_back(&paActionInfo);
-      mNodesReferences.push_back(newRef);
+      mNodesReferences.emplace_back(newNode, std::vector<CActionInfo *>{&paActionInfo});
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/com/opc_ua/opcua_local_handler.cpp
```diff
@@ -550,7 +548,7 @@ UA_StatusCode COPC_UA_Local_Handler::handleNonExistingVariable(CActionInfo &paAc
     if (UA_STATUSCODE_GOOD == retVal) {
       UA_NodeId *tmp = UA_NodeId_new();
       UA_NodeId_copy(variableInformation.mReturnedNodeId, tmp);
-      paReferencedNodes.pushBack(tmp);
+      paReferencedNodes.push_back(tmp);
```
m-meingast, line 548: You used `emplace_back()` in a number of other instances, couldn't we do that in this case as well?

src/com/opc_ua/opcua_local_handler.h
```diff
@@ -720,8 +721,8 @@ class COPC_UA_Local_Handler : public COPC_UA_HandlerAbstract, public CThread {
           return this == &rhs;
         }
 
-        CLocalMethodInfo &mActionInfo;
-        COPC_UA_Helper::UA_SendVariable_handle &mSendHandle;
+        CLocalMethodInfo *mActionInfo;
+        COPC_UA_Helper::UA_SendVariable_handle *mSendHandle;
```
m-meingast, line 725: Just to get a better understanding: Why did you change these variables from references to pointers?

src/com/opc_ua/opcua_local_handler.cpp
cochicde: I checked again the best practices on when to use either. Apparently the rule of thumb is to use `emplace_back` when you pass the parameters of the constructor directly and the object is constructed in place. In all other cases, is the same. In that case, since both are the same, maybe using `push_back` is not that bad, since for me at least, feels more intuituive what's happening. 

Also, in most cases, we are using pointers, so moving or copying is the same.

I found however some other places where a temporary object was create and then pushed back. I changed them to be constructed directly in the vector. A change is comming (compilation and tests still going on)

src/com/opc_ua/opcua_local_handler.h
cochicde: I didn't want to, but apparently erasing elements of a vector requires [MoveAssignable](https://en.cppreference.com/w/cpp/container/vector/erase) for the object, which the reference was forbidding it. Maybe there's a nice work around this issue, but I couldn't find it :(

azoitl: there is only a not so nice solution: https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper/

cochicde: yes, I thougth about that, but it didn't look clean, mainly because we are still using raw pointers everywhere. Also, I'm wasn't sure how it would affect performance. Overall, I thought it would be an over complication just to avoid pointers

azoitl: It shouldn't have an overhead. The advantage to use references wherever possible is that we know that when we are not useing this ref anymore we don't need to delete it. This is not as clear with pointers.
But I agree that we should leave it as is for now.

src/com/opc_ua/opcua_local_handler.cpp
m-meingast: I understand, that's very interesting.

m-meingast (approved):

azoitl (approved):

PR #469, opened 2025-05-23
Export cmake compilation commands

cochicde:
This files helps some tools to provide extra features. The VS code cmake plugin uses it to create the whole autocomplete and similart (or at least it improves it). 

To run clang-tidy separately on files this file is also needed


mx990:
I would suggest to ship such global defaults in a `CMakePresets.json`[^1] rather than forcibly setting them. That gives users the opportunity to override the presets if they wish. I have created a basic `CMakePresets.json` in https://github.com/eclipse-4diac/4diac-forte/pull/471.

[^1]: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

azoitl:
To add to this: with the work that @jwalt has started with PR #447 I would like to avoid any parallel cmake changes to make our life of merging not more complicated.

cochicde:
I see... I love the presets idea. I'll close this then

PR #471, opened 2025-05-23
Add CMake presets

mx990:


cochicde:
I really like the idea of this feature. I would recommend to add now also `CMakeUserPresets.json` to the `.gitignore` file. 

Also,  does this mean that we can later get rid of all the `setup_XXXX.sh` files in the root directory and migrate them to this format?

mx990:
> I would recommend to add now also `CMakeUserPresets.json` to the `.gitignore` file.

Good point, done.

> Also, does this mean that we can later get rid of all the `setup_XXXX.sh` files in the root directory and migrate them to this format?

Yes, I believe we can.

cochicde (approved):

PR #472, opened 2025-05-24
Add CMake presets for CI builds

mx990:
Add CMake presets for CI builds with tests and system tests. These presets allow to always use the same configuration as the CI builds.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -21,6 +21,45 @@
       }
     },
     {
+      "name": "base-tests",
+      "hidden": true,
+      "inherits": "base",
+      "cacheVariables": {
+        "FORTE_COM_HTTP": "ON",
+        "FORTE_COM_RAW": "ON",
+        "FORTE_COM_SER": "ON",
+        "FORTE_COM_STRUCT_MEMBER": "ON",
+        "FORTE_IO": "ON",
+        "FORTE_MODULE_RECONFIGURATION": "ON",
+        "FORTE_MODULE_RT_Events": "ON",
+        "FORTE_MODULE_SIGNALPROCESSING": "ON",
+        "FORTE_TESTS": "ON",
+        "FORTE_TEST_SANITIZE": "ON"
+      }
+    },
+    {
+      "name": "tests",
+      "hidden": true,
+      "inherits": "base-tests",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "FORTE_BUILD_STATIC_LIBRARY": "ON",
+        "FORTE_BUILD_SHARED_LIBRARY": "ON",
+        "FORTE_C_INTERFACE": "ON",
+        "FORTE_TRACE_CTF": "ON"
+      }
+    },
+    {
+      "name": "systemtests",
+      "hidden": true,
+      "inherits": "base-tests",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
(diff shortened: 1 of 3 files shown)


azoitl (approved):
LGTM, but I guess @cochicde can better say

CMakePresets.json
```diff
@@ -20,6 +20,53 @@
...
+    },
+    {
+      "name": "systemtests",
+      "hidden": true,
+      "inherits": "base-tests",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "FORTE_MODULE_POWERLINK": "ON",
+        "FORTE_TESTS": "ON",
+        "FORTE_TEST_SANITIZE": "ON",
```
cochicde: I think:
-  `FORTE_TEST` should be in `base-test`
- `FORTE_MODULE_POWERLINK` should also go to `base-test` (although the CI currently has it also only in the system test)
- `FORTE_TEST_SANITIZE` also go to `base-test`? This I'm not sure, since we might at some point run release test maybe?

CMakePresets.json
```diff
@@ -20,6 +20,53 @@
...
+        "FORTE_COM_RAW": "ON",
+        "FORTE_COM_SER": "ON",
+        "FORTE_COM_STRUCT_MEMBER": "ON",
+        "FORTE_IO": "ON",
+        "FORTE_MODULE_CONVERT": "ON",
+        "FORTE_MODULE_IEC61131": "ON",
+        "FORTE_MODULE_RECONFIGURATION": "ON",
+        "FORTE_MODULE_RT_Events": "ON",
+        "FORTE_MODULE_SIGNALPROCESSING": "ON",
+        "FORTE_MODULE_UTILS": "ON"
```
cochicde: since you inherit from `base`, the following can be omitted here:
- `FORTE_COM_ETH`
- `FORTE_COM_FBDK`
- `FORTE_COM_LOCAL`
- `FORTE_MODULE_CONVERT`
- `FORTE_MODULE_IEC61131`
- `FORTE_MODULE_UTILS`

CMakePresets.json
mx990: >     * `FORTE_TEST` should be in `base-test`

Makes sense, moved.

>     * `FORTE_MODULE_POWERLINK` should also go to `base-test` (although the CI currently has it also only in the system test)

I had tried that originally, but it failed because it does not build as a shared library:
```
/usr/bin/ld: /home/runner/work/4diac-forte/4diac-forte/build/_deps/openpowerlink-src/libpowerlink.a(EplSdoComu.c.o): warning: relocation against `EplSdoComConCb' in read-only section `.text'
/usr/bin/ld: /home/runner/work/4diac-forte/4diac-forte/build/_deps/openpowerlink-src/libpowerlink.a(EplObd.c.o): relocation R_X86_64_PC32 against symbol `abEplObdTrashObject_g' can not be used when making a shared object; recompile with -fPIC
```

I would like to just get the status quo working for now, since all of this is anyhow about to change with https://github.com/eclipse-4diac/4diac-forte/pull/447.

>     * `FORTE_TEST_SANITIZE` also go to `base-test`? This I'm not sure, since we might at some point run release test maybe?

I have also moved it for now. We can still change it when adding release tests or maybe keep the sanitizers for them as well.

CMakePresets.json
mx990: Done

cochicde (approved):

PR #473, opened 2025-05-24
Type lib clean up

azoitl:


src/core/resource.cpp
```diff
@@ -336,27 +336,31 @@ EMGMResponse CResource::readValue(forte::core::TNameIdentifier &paNameList, std:
 
 #ifdef FORTE_SUPPORT_QUERY_CMD
 
+namespace {
+
+  template<typename T>
+  void appendTypeNameList(std::string &paValue, const std::vector<T *> &paTypeEntries) {
+    if (!paTypeEntries.empty()) {
+    }
```
mx990: I guess there should be something in here?

src/core/typelib.cpp
```diff
@@ -19,87 +19,85 @@
...
 USE_STRING_ID(ARRAY);
 
 using namespace std::string_literals;
 
-CTypeLib::CTypeEntry::CTypeEntry(CStringDictionary::TStringId paTypeNameId) :
-    mTypeNameId(paTypeNameId),
-    mNext(nullptr) {
-}
+namespace {
+  EMGMResponse mLastErrorMSG = EMGMResponse::Ready;
```
mx990: I would suggest declaring this variable as `thread_local`[^1] to avoid potential race conditions with other threads.

[^1]: https://en.cppreference.com/w/cpp/language/storage_duration#Thread_storage_duration

mx990 (changes requested):

src/core/typelib.cpp
azoitl: I'm not sure if all our operating systems are supporting thread_local. Also I never liked the design that I did 20 years ago. I now came up with a different approach. This should be thread save and cleaner.

src/core/resource.cpp
azoitl: Thx for noting.

src/core/typelib.cpp
```diff
@@ -19,124 +19,127 @@
...
 USE_STRING_ID(ARRAY);
 
 using namespace std::string_literals;
 
-CTypeLib::CTypeEntry::CTypeEntry(CStringDictionary::TStringId paTypeNameId) :
-    mTypeNameId(paTypeNameId),
-    mNext(nullptr) {
-}
+namespace {
+  // EMGMResponse mLastErrorMSG = EMGMResponse::Ready;
```
mx990: I believe we do not need to keep the comment?

src/core/typelib.h
```diff
@@ -249,7 +251,8 @@ class CTypeLib {
      */
     static CFunctionBlock *createFB(CStringDictionary::TStringId paInstanceNameId,
                                     CStringDictionary::TStringId paFBTypeId,
-                                    forte::core::CFBContainer &paContainer);
+                                    forte::core::CFBContainer &paContainer,
+                                    EMGMResponse &paErrorMSG);
```
mx990, line 255: I would suggest adding delegates without the `EMGMResponse &` parameter, so we do not need those extra variables all over the place.

mx990 (changes requested):

src/core/typelib.cpp
azoitl: yes, removed.

src/core/typelib.h
azoitl: good idea, this will also help for the next change upcoming. 

I only added it for data types and fbs. Adapters will anyhow be reworked so I kept it the way for now.

mx990 (approved):

mx990:
It seems the tests get stuck on Windows. I have created a PR https://github.com/eclipse-4diac/4diac-forte/pull/474 that adds a timeout to prevent stuck builds.

mx990:
I believe you need to rebase for the timeout change to be incorporated into the workflows. We will then hopefully see more output than when just canceling the workflow altogether.

azoitl:
I'm in the middle of something will do that afterwords.

src/core/typelib.cpp
```diff
@@ -19,130 +19,138 @@
...
-CTypeLib::CTypeEntry::CTypeEntry(CStringDictionary::TStringId paTypeNameId) :
-    mTypeNameId(paTypeNameId),
-    mNext(nullptr) {
-}
+namespace {
+  std::vector<CTypeLib::CFBTypeEntry *> fbTypeLib;
+
+  std::vector<CTypeLib::CAdapterTypeEntry *> adapterTypeLib;
+
+  std::vector<CTypeLib::CDataTypeEntry *> dataTypeLib;
```
mx990: We may be running into the infamous *static initialization order fiasco*[^1] here. The pointers in the original version were simply *zero-initialized*[^2], so these lists were likely safe to be used in other initializers. However, `std::vector` is likely not. I would suggest the *construct on first use idiom*[^3] as a possible solution.

[^1]: https://en.cppreference.com/w/cpp/language/siof
[^2]: https://en.cppreference.com/w/cpp/language/initialization#Static_initialization
[^3]: https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use

mx990 (changes requested):

src/core/typelib.cpp
azoitl: This brings me to the question if memory wise the old single linked list was better? The only advantage the new design has the sorted inserts and improved search. But the draw-back of the vectors is that it uses now more heap.

azoitl: But for testing I did the proposed change. Now all builds run through.

mx990: As an alternative, we could also look into using a skip list with fixed layers. That would allow us to store the actual list in the entries, while also enabling an efficient search.

A skip (or linked) list would also be easier to make thread safe, which may become relevant when we dynamically load types while FORTE is running. In that case, we should only need locking during inserts and there are even completely lock-free algorithms available.

azoitl: That sounds like an interesting thing for later. I would love to look into it when the urgent 3.0 things are fixed. Maybe for 3.1. But after some thinking I would leave it for now as is.

mx990 (approved):

PR #477, opened 2025-05-26
Added Type hash checking to FBs, Adapters, and Data Types

azoitl:


src/core/typelib.cpp
```diff
@@ -145,7 +158,12 @@ CFunctionBlock *CTypeLib::createFB(CStringDictionary::TStringId paInstanceNameId
     newFB = createGenericFB(paInstanceNameId, paFBTypeId, paContainer, paErrorMSG);
   }
 
-  if (newFB != nullptr && !newFB->initialize()) {
+  if (newFB == nullptr) {
+    paErrorMSG = EMGMResponse::UnsupportedType;
```
mx990: I believe this will overwrite any previous error codes set above in line 155 or in `createGenericFB`.

src/stdfblib/events/E_CTU_fbt.cpp
mx990, line 1: Should this be committed?

mx990 (changes requested):

src/core/typelib.cpp
azoitl: yes you are right, changed this with a wrong assumption.

src/stdfblib/events/E_CTU_fbt.cpp
azoitl: I wanted to show how I would export the type hash so that a better total picture is available. Especially as I was not sure if we should have to different FB defines or not.

src/stdfblib/events/E_CTU_fbt.cpp
```diff
@@ -25,15 +33,11 @@ USE_STRING_ID(R);
...
-#include "forte_uint.h"
-#include "forte_bool.h"
-#include "iec61131_functions.h"
-#include "forte_array_common.h"
-#include "forte_array.h"
-#include "forte_array_fixed.h"
-#include "forte_array_variable.h"
+namespace {
+  constexpr char TypeHash_FORTE_E_CTU[] = "b01bf801fa456ad839e2b3148fd69fcb105d6213599225e9b7e37a30d404929a780f1d16418c6b976ebbc3387b8ae9184b2ea960845ac72f8b65fb63680542bb";
+}
```
mx990: It seems a bit strange that this is seemingly unused but still referenced from within `DEFINE_FIRMWARE_FB_HASH`. I believe it would be better to simply pass the string literal as a parameter to `DEFINE_FIRMWARE_FB_HASH`.

src/stdfblib/events/E_CTU_fbt.cpp
mx990: > I wanted to show how I would export the type hash so that a better total picture is available.

I just wanted to make sure that we do not accidentally add a hash to a standard library type.

> Especially as I was not sure if we should have to different FB defines or not.

I have a slight preference towards only one define with a varargs parameter, but I am also not entirely sure.

azoitl: > I just wanted to make sure that we do not accidentally add a hash to a standard library type.
Why not? Shouldn't all of our types have at one point a type hash?

 > I have a slight preference towards only one define with a varargs parameter, but I am also not entirely sure.

When I originally tried varargs I was not happy with the result. I searched a bit further and found that C++ 20 now has `__VA_OPT__` which allows this and also your requirement from below.

src/stdfblib/events/E_CTU_fbt.cpp
azoitl: good point. With my change this is now done.

src/stdfblib/events/E_CTU_fbt.cpp
mx990: It seems MSVC needs an extra compiler flag `/Zc:preprocessor`[^1] to support this.

[^1]: https://devblogs.microsoft.com/cppblog/announcing-full-support-for-a-c-c-conformant-preprocessor-in-msvc/

src/core/resource.cpp
```diff
@@ -446,154 +454,59 @@ void CResource::createConnectionResponseMessage(const CConnection &paConn,
...
-    createDataInterfaceResponseMessage(paInterfaceSpec, paReqResult);
-    createAdapterInterfaceResponseMessage(paInterfaceSpec, paReqResult);
-    paReqResult.append("</InterfaceList>\n");
-    retVal = EMGMResponse::Ready;
-  }
-  return retVal;
-}
+    if (paTypeHash.empty()) {
+      // return type hash
+      paReqResult += '<';
```
ernstblechaPT: I think this line is superfluous as this will be overwritten in the next line

Also: this piece of code seems to have tabs and spaces mixed up?

src/core/resource.cpp
```diff
@@ -133,10 +135,16 @@ EMGMResponse CResource::executeMGMCommand(forte::core::SManagementCMD &paCommand
       case EMGMCommandType::QueryAdapterTypes: retVal = queryAllAdapterTypes(paCommand.mAdditionalParams); break;
       case EMGMCommandType::QueryFB: retVal = queryFBs(paCommand.mAdditionalParams, *this, ""); break;
       case EMGMCommandType::QueryFBType:
-        retVal = createFBTypeResponseMessage(paCommand.mFirstParam.front(), paCommand.mAdditionalParams);
+        retVal = createFBTypeResponseMessage(paCommand.mFirstParam.front(), paCommand.mAdditionalParams,
```
ernstblechaPT, line 138: It's probably just me being confused, but I do not understand why paCommand.mAdditionalParams is passed twice?

azoitl: I was not sure about this thing. The problem is that in `mAdditionalParams` the type hash is given from the command parser. Also any response generated shall be put into `mAdditionalParams`. I wanted to have a means inside of createFBTypeResponseMessage and its siblings to distinguish this. Which makes it also more future prove if the design of the command parsing will change.

src/core/resource.cpp
azoitl: thx for noting this was mixed up. The first line should be an = and the next an +=. fixed now.

src/stdfblib/events/E_CTU_fbt.cpp
azoitl: Oh didn't expect that.  Added to our cmake.

mx990 (approved):
Looks good to me. Please remember to remove the hash in `src/stdfblib/events/E_CTU_fbt.cpp` before merging.

azoitl:
> Looks good to me. Please remember to remove the hash in `src/stdfblib/events/E_CTU_fbt.cpp` before merging.

done

ernstblechaPT (approved):

PR #479, opened 2025-05-28
addeded type hash and query support for global const

azoitl:


src/core/typelib.h
```diff
@@ -153,6 +153,18 @@ private:
...
 
+#define DECLARE_FIRMWARE_GLOBAL_CONST()                                                                                \
+private:                                                                                                               \
+  const static CTypeLib::CGlobalConstEntry csmGlobalConstEntry;                                                        \
+                                                                                                                       \
+public:                                                                                                                \
+  FORTE_DUMMY_INIT_DEC                                                                                                 \
+private:
+
+#define DEFINE_FIRMWARE_GLOBAL_CONST(gcClass, gcTypeNameId)                                                            \
```
mx990: I believe we should also pass the hash as an optional argument here.

mx990 (changes requested):

mx990:
I would also suggest already adding the new defines to the existing global constant types in [src/modules/utils/const/quarterconst_gcf.h](https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/modules/utils/const/quarterconst_gcf.h) and [src/modules/signalprocessing/FIELDBUS_SIGNAL_gcf.h](https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/modules/signalprocessing/FIELDBUS_SIGNAL_gcf.h).

azoitl:
> I would also suggest already adding the new defines to the existing global constant types in [src/modules/utils/const/quarterconst_gcf.h](https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/modules/utils/const/quarterconst_gcf.h) and [src/modules/signalprocessing/FIELDBUS_SIGNAL_gcf.h](https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/modules/signalprocessing/FIELDBUS_SIGNAL_gcf.h).

Can do that as soon as this is merged: https://github.com/eclipse-4diac/4diac-ide/pull/1347

src/core/typelib.h
azoitl: thx for noting, resulted from a rebase error. fixed now.

azoitl:
> I would also suggest already adding the new defines to the existing global constant types in [src/modules/utils/const/quarterconst_gcf.h](https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/modules/utils/const/quarterconst_gcf.h) and [src/modules/signalprocessing/FIELDBUS_SIGNAL_gcf.h](https://github.com/eclipse-4diac/4diac-forte/blob/develop/src/modules/signalprocessing/FIELDBUS_SIGNAL_gcf.h).

done

ernstblechaPT (approved):

mx990 (approved):

PR #481, opened 2025-06-02
Adapter rework

azoitl:


src/core/adapter.cpp
```diff
@@ -12,40 +12,41 @@
...
-        paContainer, paIsPlug ? paInterfaceSpecPlug : paInterfaceSpecSocket, paInstanceNameId),
-    mOutputEventIds(getFBInterfaceSpec().mNumEOs, cgInvalidEventID),
-    mParentAdapterListEventID(0),
+                   bool paIsPlug,
+                   TForteUInt8 paParentAdapterlistID,
+                   bool paIsInCfb) :
+    CFunctionBlock(paContainer, paInterfaceSpec, paInstanceNameId),
     mIsPlug(paIsPlug),
+    mIsInCfb(paIsInCfb),
+    mParentAdapterListEventID(static_cast<TForteUInt16>((paParentAdapterlistID + 1) << 8)),
```
mx990: Could we avoid the separate `pa/mIsInCfb` by setting `paParentAdapterlistID` to `-1` in CFBs, resulting in a `mParentAdapterListEventID` of `0`?

src/core/adapter.h
```diff
@@ -113,28 +77,29 @@ class CAdapter : public CGenFunctionBlock<CFunctionBlock> {
...
     void executeEvent(TEventID paEIID, CEventChainExecutionThread *const paECET) override;
-    void setupEventEntryList();
 
     const bool mIsPlug;
+    const bool mIsInCfb;
+    TForteUInt16 mParentAdapterListEventID;
+
     CAdapter *mPeer;
-    CIEC_ANY **mLocalDIs;
     CAdapterConnection *mAdapterConn;
```
mx990: It seems there is some redundant information here between the adapter (the `mPeer`) and the adapter connection (`mPlug` and `mSocket`). Could we eliminate either side?

src/core/adapter.cpp
```diff
@@ -12,40 +12,41 @@
...
     mPeer(nullptr),
-    mLocalDIs(mDIs),
-    mAdapterConn(nullptr),
-    mParentFB(nullptr) {
+    mAdapterConn(nullptr) {
+
+  if (isPlug()) {
+    // the plug is in charge of managing the adapter connection
+    mAdapterConn = new CAdapterConnection(static_cast<CFunctionBlock &>(getParent()), paParentAdapterlistID, *this);
```
mx990: Would it make sense to also create separate plug and socket base classes, one with the `CAdapterConnection` by value and the other with a pointer?

Alternatively, maybe the adapter connection could also live in the parent FB, similar to data connections? We could then maybe even create a `CAnyAdapterConnection` and `virtual CFunctionBlock::connectAdapter` that handle the creation of the typed adapter instance on connect, again similar to generic data connections?

src/core/anyadapter.cpp
```diff
@@ -14,52 +14,58 @@
...
+bool CAnyAdapter::connect(CAdapter *paPeer, CAdapterConnection *paAdConn) {
+  if (!CAdapter::connect(paPeer, paAdConn)) {
+    return false;
+  }
+  EMGMResponse errorMSG;
+  mConfiguredAdapter = std::unique_ptr<CAdapter>(CTypeLib::createAdapter(
+      getInstanceNameId(), paPeer->getFBTypeId(), getParent(), isPlug(), mParentAdapterlistID, errorMSG));
+
+  if (mConfiguredAdapter) {
+    mConfiguredAdapter->connect(paPeer, paAdConn);
```
mx990: It seems we are connecting twice, once with this adapter above and now with the typed adapter here. Will this not overwrite the peer in `paPeer`?

src/stdfblib/events/E_TimeOut_fbt.h
```diff
@@ -43,7 +43,7 @@ class FORTE_E_TimeOut final : public CEventSourceFB {
     FORTE_E_TimeOut(CStringDictionary::TStringId paInstanceNameId, forte::core::CFBContainer &paContainer);
     bool initialize() override;
 
-    FORTE_ATimeOut var_TimeOutSocket;
+    FORTE_ATimeOut_Socket var_TimeOutSocket;
```
mx990: I would suggest changing this to:
```suggestion
    forte::core::CInternalFB<FORTE_ATimeOut_Socket> fb_TimeOutSocket;
```

That way, we can avoid explicitly delegating to adapters in `initailize()` and also do not need special handling in `CFunctionBlock::getFB`.

src/core/anyadapter.cpp
azoitl: no this will not be the case, it will only set the peer for the given adapter. most probably the name of the method will need to change.

src/core/adapter.cpp
azoitl: this will only work when we move the adapter connection out of the adapters.

src/core/adapter.h
azoitl: when we move the adapter connection out of the adapter we would only need the peer.

src/core/adapter.cpp
azoitl: I thought about a plug and socket base class. but then we would loose the advantage of the common base class of plug and socket holding all data variables. Therefore I decided against it. 

Reading all your comments and as I also thought about it moving the adapter connection out of the adapter class is I think the better solution. Will investigate that further.

src/core/adapter.cpp
azoitl: I rechecked all usages of adapter connections and also normal connections and the using -1 for paParentAdapterListID works also when the connection is not extracted. Changed accordingly.

src/core/adapter.cpp
azoitl: Moving the connection outside would just duplicate a lot of code and after working through all of our adapter connection handling code I still think keeping it inside is the better solution.

src/core/adapter.h
azoitl: yes it is partly redundant. But moving the connection out of the adapter will keep the redundancy. Removing the peer will make it where hard to correctly implement the any adapter wrapping. Therefore I think having both is the better solution.

src/core/adapter.cpp
azoitl: I'm still undecided if turning getAdapterConnection virtual is a good solution. It whould definitly mean much more code to be generated and that is also not what I find good.

src/stdfblib/events/E_TimeOut_fbt.h
azoitl: done. But I still kept the getAdapterUnchecked because it is used a lot in different places.

src/core/adapter.cpp
```diff
@@ -12,112 +12,38 @@
...
-}
 
-bool CAdapter::connect(CAdapter *paPeer, CAdapterConnection *paAdConn) {
-  bool bRetVal = false;
-  if (mPeer == nullptr) {
-    mPeer = paPeer;
-    mDIs = paPeer->mDOs; // TODO: change is adapters of subtypes may be connected
-    if (isSocket()) {
-      mAdapterConn = paAdConn;
+  bool CAdapter::initialize() {
```
mx990: I believe this can be removed now.

src/core/adapter.cpp
```diff
@@ -12,112 +12,38 @@
...
-
-void CAdapter::fillEventEntryList() {
-  for (TEventID i = 0; i < getFBInterfaceSpec().mNumEOs; ++i) {
-    mOutputEventIds[i] = static_cast<TPortId>(mParentAdapterListEventID + i);
-  }
-}
-
-void CAdapter::setParentFB(CFunctionBlock *paParentFB, TForteUInt8 paParentAdapterlistID) {
-  mParentAdapterListEventID = static_cast<TForteUInt16>((paParentAdapterlistID + 1) << 8);
+USE_STRING_ID(ANY_ADAPTER);
```
mx990: I believe this can be removed now.

src/core/adapter.h
```diff
@@ -16,125 +16,119 @@
...
+          core::CInternalFB<T>(paInstanceNameId, paParentFB, paParentAdapterlistID),
+          mAdapterCon(paParentFB, paParentAdapterlistID, *this) {
+      }
+
+      CAdapter *getAdapterBlock() override {
+        return static_cast<CAdapter *>(core::CInternalFB<T>::get());
+      }
+
+      CStringDictionary::TStringId getAdapterTypeId() const override {
+        return getAdapterBlock().getFBTypeId();
```
mx990: I believe this needs a `->` instead of a `.`, since `getAdapterBlock()` returns a pointer (same below)?

src/core/adapterconn.cpp
```diff
@@ -29,42 +40,39 @@ CAdapterConnection::~CAdapterConnection() {
...
 
-  if (mPlug.connect(socket, this) && socket->connect(&mPlug, this)) {
-    mSocket = socket;
-    paDstFB.incConnRefCount();
-    getSourceId().getFB().incConnRefCount();
-    return EMGMResponse::Ready;
+  if (socket->getAdapterTypeId() == STRID(ANY_ADAPTER)) {
+    // we need to configure the socket first to get the right peer
+    socket->connect(this);
+    mPlug.setPeer(socket->getAdapterBlock());
```
mx990: How about we move this line also into `ISocketPin::connect` to avoid the special handling for `ANY_ADAPTER` here?

src/core/adapter.h
```diff
@@ -16,125 +16,119 @@
...
+
+    public:
+      CPlugPin(CStringDictionary::TStringId paInstanceNameId,
+               CFunctionBlock &paParentFB,
+               TForteUInt8 paParentAdapterlistID) :
+          core::CInternalFB<T>(paInstanceNameId, paParentFB, paParentAdapterlistID),
+          mAdapterCon(paParentFB, paParentAdapterlistID, *this) {
+      }
+
+      CAdapter *getAdapterBlock() override {
```
mx990: I would suggest returning the narrower `T*` here instead.

src/core/adapterconn.cpp
```diff
@@ -12,14 +12,25 @@
...
-#include "funcbloc.h"
-#include "adapter.h"
-#include "anyadapter.h"
+namespace {
+
+  bool isCompatible(CStringDictionary::TStringId paPlugTypeId, CStringDictionary::TStringId paSocketTypeId) {
+    // Need to check any adapter here as we don't know which adapter side is used for checking compatibility
+    return ((paPlugTypeId == paSocketTypeId) ||
+            ((paPlugTypeId == STRID(ANY_ADAPTER)) && (STRID(ANY_ADAPTER) != paSocketTypeId)) ||
+            ((paPlugTypeId != STRID(ANY_ADAPTER)) && (STRID(ANY_ADAPTER) == paSocketTypeId)));
```
mx990: Could we move this into `CAnyAdapterPin` by making `isCompatible` a virtual method and using a reciprocating check, such as `mPlug.isCompatible(*socket) || socket->isCompatible(mPlug)`?

src/core/adapterconn.h
```diff
@@ -11,55 +11,68 @@
...
+  class ISocketPin : public IAdapterPin {
+    public:
+      virtual void connect(CAdapterConnection *paConn) = 0;
+      virtual void disconnect() = 0;
+      virtual CAdapterConnection *getAdapterCon() = 0;
+
+      const CAdapterConnection *getAdapterCon() const {
+        return const_cast<ISocketPin *>(this)->getAdapterCon();
+      }
+  };
```
mx990, line 85: I would suggest moving the pin interfaces into `adapter.h`, so they are together with the non-any implementations.

src/core/anyadapter.cpp
```diff
@@ -8,58 +8,96 @@
...
+void CAnyAdapterPin::createConfiguredAdapter(CAdapter *paPeer,
+                                             CFunctionBlock &paParent,
+                                             bool paIsPlug,
+                                             TForteUInt8 paParentAdapterlistID) {
+  removeConfiguredAdapter();
+  EMGMResponse errorMSG;
+  mConfiguredAdapter = std::unique_ptr<CAdapter>(forte::core::createAdapter(
+      mInstanceNameId, paPeer->getFBTypeId(), paParent, paIsPlug, paParentAdapterlistID, errorMSG));
+
+  if (mConfiguredAdapter) {
```
mx990, line 41: Should we propagate possible errors up through `connect`?

src/core/funcbloc.cpp
```diff
@@ -440,10 +416,6 @@ void CFunctionBlock::freeEventMonitoringData() {
 CFunctionBlock *CFunctionBlock::getFB(NameIterator &paNameListIt, NameIterator paNameListEnd) {
```
mx990: I believe this can be removed now.

mx990 (changes requested):

src/core/adapterconn.h
azoitl: I chose to have the interfaces in adapterconn because the pin interfaces are the classes for managing the connection. By having it here adapter conn is in that sense self-contained, which I found cleaner.

src/core/adapter.cpp
azoitl: yes, mixed it with setInitialValue, done

src/core/adapter.cpp
azoitl: yes, done

src/core/adapter.h
azoitl: strange that the compiler didn't tell me that. But you are totally right.

src/core/adapterconn.cpp
azoitl: good point. This allows future special connection handling.

src/core/adapter.h
azoitl: Didn't know that this is allowed in C++. Thought this is only possible in Java.

src/core/adapterconn.cpp
azoitl: thx for this idea. changed accordingly.

src/core/anyadapter.cpp
azoitl: yes I think this should be done. changed

src/core/funcbloc.cpp
azoitl: yes

src/core/adapterconn.cpp
```diff
@@ -12,14 +12,14 @@
  *    Martin Melik-Merkumians - adds typifyAnyAdapter
  *******************************************************************************/
 #include "adapterconn.h"
+#include "funcbloc.h"
+#include "adapter.h"
 
 USE_STRING_ID(ANY_ADAPTER);
```
mx990: I believe this can also be removed now?

src/core/funcbloc.cpp
```diff
@@ -352,9 +338,14 @@ EMGMResponse CFunctionBlock::changeExecutionState(EMGMCommandType paCommand) {
...
+      if (forte::IPlugPin *plug = getPlugPinUnchecked(i); plug->getAdapterBlock() != nullptr) {
+        plug->getAdapterBlock()->changeExecutionState(paCommand);
+      }
+    }
+    for (TPortId i = 0; i < getFBInterfaceSpec().mSocketNames.size(); i++) {
+      if (forte::ISocketPin *skt = getSocketPinUnchecked(i); skt->getAdapterBlock() != nullptr) {
+        skt->getAdapterBlock()->changeExecutionState(paCommand);
       }
     }
   }
```
mx990: I believe that should already be handled by `CFBContainer::changeExecutionState` now?

mx990 (changes requested):

mx990 (approved):

PR #487, opened 2025-06-04
Removed SingleLinkedList from different IO interfaces

azoitl:


src/modules/PLCnext/slaveHandler.cpp
```diff
@@ -63,17 +59,13 @@ void PLCnextSlaveHandler::initBufferImage(size_t size) {
 
 void PLCnextSlaveHandler::addHandle(PLCnextSlaveHandle *paHandle) {
   CCriticalRegion criticalRegion(handleMutex);
-  slaveHandles->pushBack(paHandle);
+  mDeviceHandles.pushBack(paHandle);
```
moniwe: should it be push_back(paHandle)?

azoitl: yes it should. Thanks for noting. Fixed it.

moniwe (approved):

PR #496, opened 2025-06-05
Remove type prefixes in coding rules

mx990:
This is a proposal to remove prefixes on type definitions, such as `C` for classes and `T` for typedefs, in the coding rules.

The actual benefit of type prefixes is somewhat questionable. It often matters little whether a type is a class, struct, or just a plain typedef. In modern IDEs, the definition of a type is also only a hover or click away.

On the other hand, it is often beneficial or even necessary to change the nature of a type as development progresses. For example, a class may be turned into a typedef to reuse some existing functionality. A strict adherence to type prefixes would then require changing all usages of a type to the new prefix. This not only requires additional work but also produces significant noise in diffs. It also threatens source compatibility or locks development into keeping a prefix despite reasons to change it.

I would therefore propose to remove type prefixes in the coding rules and for new types, while keeping prefixes on existing types.


azoitl:
I was the one who original introduced them, but by now I also don't think they are really not needed anymore. The only Prefix that could make sense in a C++ context is the `I` for indicating that the class is intended to be a pure interface with pure virtual methods only.

MartinMelikMerkumians:
I am fine if we remove them either with keeping `I` or not

cochicde (approved):
I'd remove all of them, even the `I` as it  could suffer the same issues as the other prefixes and the benefits it brings are too little.

azoitl:
> I'd remove all of them, even the `I` as it could suffer the same issues as the other prefixes and the benefits it brings are too little.

@cochicde in the last days I thought a bit more about this. Also because of the stuff I was doing for adapters. While I agree for all other prefixes I by know strongly disagree for classes that are used as interfaces. As C++ has no dedicated interface feature I would like to have something that tell other developers that the class I added is intended as interface. As such it should not have any fields and only have methods that use pure virtual methods (i.e. the equivalent to default methods in Java). If we don't have any indication it is way to easy that someone uses an class where the intention was to act as interface and violate these rules. 

If you know an other option on how to do this I'm very open to it.

gravemalte:
I’m in favor of keeping the `I` prefix for interfaces. Personally, I like it because it makes it super clear that a class is purely an interface. That kind of clarity really helps, especially when multiple inheritance starts to muddy things up.

So yeah, my vote: drop all prefixes except `I`.

cochicde:
I'd prefer to have this information in the documentation of the class. This gives other developers the same information as the `I` and more.

Also, if at some point the interface stops being one, or the other way around, a non interface becomes one, the name changes, bringing the problems which initially triggered this PR.

I think that providing the information that a class is an interface is good, but having it in the name itself seems too invasive for me. The mental effort it saves is too little and adds the extra effort for all developers to take this into account when creating an interface.

mx990:
I would also like to add that if we keep only the `I` prefix, it may become confusing whether a class is intended as an interface or if the name merely starts with an acronym that starts with an `I`. In fact, we already have several classes in the codebase that start with `IO`. More and less obvious cases may follow once we drop the other prefixes.

azoitl:
@mx990 you have a point here. However I still would like to have something that clearly indicates this is an interface. The proposal of doing this in documentation as @cochicde proposed is to weak for me. I went through lots of code and noticed that our documentation is very often not up to date. 

Also deciding for an interface is an architectural decision. Changing that shouldn't be that easy. First of all it changes the architecture and especially in C++ it may break something. 

What other options do we have?

mx990:
We could add a `static_assert` with a meaningful message that ensures the class is abstract and has no non-static data members. Unfortunately, there does not seem to be any type trait in the standard library to check for the latter, so we would have to add our own.

I have created an example to show what such a type trait could look like at https://godbolt.org/z/bhEqcjsof. It should work on most compilers and architectures, as long as the interface does not have more than one virtual base class itself.

azoitl:
In parallel I also searched a bit and discussed with ChatGPT. It came up with the proposal of an interface base class:

 `template<typename Derived>
  class Interface {  ...`

where this static assertions could be added. 

With these approaches I'm happy to drop all prefixes of classes and structs.

cochicde:
The original problem is to inform others developers that the class is intended as an interface. That's the only reason we had the `I` in the first place. Addding traits and templates is about enforcing a class to be an interface,  which is a feature that the original `I` didn't provide. Also, the concept of interface does not exist in C++, and the idea is probably coming from Java, where multiple inheritance is not possible.
If the documentation solution is not good enough because the documentation is not up to date, then we should update the documentation, not look for workarounds to patch this issue. I quickly check and could find only 5 "interface" classes (2 of them are not actually interfaces as default destructors are provided). The cost of replacing the information provided by the `I` prefix in the whole code is only adding documentation to these classes. 

If we really want to discuss how we could enforce a class to be an interface and get a compilation error, I'd say this PR has nothing to do with it.

azoitl:
@cochicde interfaces are definitely not coming originally from Java.  Interfaces are a concept for hiding implementation details and with that much more then just circumventing single inheritance. 

I just recently added two interface classes which have the intention of an interface allowing different implementations of these interfaces. For them I will add the proposed enforcer. While I'll agree with you that the documentation should be updated I see how easily it is to forget that and that the documentation is broken. Therefore I'm always a fan of documenting principles (like interfaces) with code. Then the compiler will check for me. 

Interface classes in C++ may need to provide a virtual default constructor. This is needed when you want to store pointers to instances of these interfaces and delete them later. So this is not a criteria.

cochicde:
You're right. I was too fast to write about interfaces. I'm not against enforcing certain aspects of the code, but I'm not sure if it's a good idea to have the enforcing also in the coding rules.

mx990:
It seems the discussion on this PR has gotten somewhat stale. I still believe we should drop all type prefixes in the coding rules for the reasons outlined above.

I would further consider any enforcement regarding interfaces as a separate issue. @azoitl Could I leave that to you, since you already looked into it?

If there are no more objections, I would consider merging this?

azoitl:
@mx990 yes lets merge this and proceed as discussed. I will write an issue for adding an interface enforcement class.

PR #498, opened 2025-06-09
CMake Overhaul (almost production ready)

jwalt:
This is the latest incarnation of #447, now with all of 4diac forte working, based off latest develop. However, this PR only includes the manual changes. In order to get a buildable forte, you need to run the included auto-refactoring scripts. As a result, this is Linux-only at the moment. However, the test suite builds and succeeds.

To build:

```sh
pip install cmake-format # or install cmake-format any other way you like
cd src
./cleanup-headers.sh && ./update-cmakelists.sh core com modules stdfblib
cd ..
```

then proceed normally (e.g. using 4diac-fbe). Please don't look at those scripts, they are as quick and dirty as they can get and need to be removed for the final merge.

Most of the comments in #447 have been incorporated, and I want to push this out now. Some small issues should probably be revisited in the near future to address the rest. But I propose merging this quickly as this is a huge change already.

Open issues:
* arch platforms (only posix is ported)
* test/systemtest cmakelists.txt (they work, but do not use the clean new style)
* stacktrace is missng (tracing compiles fine now)
* copyright notices have not been finalised yet

NOTE: You can find a branch that includes the auto-refactorign here: https://github.com/jwalt/4diac-forte/tree/feature/cmake-refactor-step2

PR #499, opened 2025-06-11
Feature/cmake refactor step2

jwalt:
This is #498 with the generated files/refactoring added. It only exists to take advantage of CI testing. Please do not comment on this, keep the discussion in one place, #498


moniwe:
I built Wago PFC200 on this branch and realized that `.../fordiac-fbe/modules/EclipseGeneratedFBs/CMakeLists.txt` needs to be updated, since some commands are not available anymore. I changed the last lines that it seemed to work for me:

```
# now work with the updated files
forte_register_module(CUSTOM_FBS ON "Custom FBs")

file(GLOB CUSTOM_SRCS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} edited/*.cpp)
add_library(custom_fbs OBJECT ${CUSTOM_SRCS})

foreach(CUSTOM_SRC IN LISTS CUSTOM_SRCS)
  string(REGEX REPLACE "\\.cpp" "" CUSTOM_SRC "${CUSTOM_SRC}")
  message(STATUS ">>> Adding custom FB: ${CUSTOM_SRC}")
endforeach()
```

I was also able to start forte on Wago PFC200. There are less files within the bundle folder now. Currently I was not able to deploy the modular IOs, since:

```
<!-- 192.168.2.44:61499 -->
<Request ID="0" Action="QUERY">
    <FBType Name="E_RESTART#v1:SHA3-512:Fp-h2OJvUi-xRKANuxa6BbOUW0toz0VVL4mHV8PtMBqt-muQvuJXIqH1VUSvybH_5Ma5amBzUCsrQxUW8etYwg=="/>
</Request>

<Response ID="0" Reason="UNSUPPORTED_TYPE"/>


<!-- 192.168.2.44:61499 -->
<Request ID="1" Action="QUERY">
    <FBType Name="WagoMaster#v1:SHA3-512:S-c0JUHHIRK9iRVwFaSXCnevExOtWWd9ftAcYmHIOqYLVt6SaUAnlKD4WX51F5DbD2Lylk-Ovh9WEXkqYUd3mg=="/>
</Request>

<Response ID="1" Reason="UNSUPPORTED_TYPE">
  
</Response>
```

cochicde:
I was trying to build the C_INTERFACES and started working on the `CMakeLists.txt` files, but I see that now each module adds itself to the compilation by linking directly to `forte`. This is a problem, since this is assuming we are building only one executable/library. This affects also the static/shared library separation we had before. Previously everything compilation unit was added to the so called `forte-lite` and then executables and libraries were defined out of it. 

By having everthing directly added to `forte`, libraries are harder to configure. For example, the arch libraries add also `main.cpp` to forte, which is not valid for libraries. 

I'm wondering also, is it the responsability of the module to add themselves to the overall compilation? Isn't clear that the container module/folder "pulls" the library out of the submodules/subfolders and adds them to an intermediate library or similar? Like, the `com` folder would create `com` library which links to the `HTTP`, `modbus`, `opc_ua` and so on, and then the overall `forte` would link only to `com`. I'd argue that the responsability of a folder/module is to create a library but not also link it.

As it is now, adding a new compilation artifact like a library, requires adding in every  folder

azoitl:
@cochicde To be honest I never needed the feature to build executable and libs from one configuration. So for me it would be fine if the target forte would be one of the three options. Would this solve the issue? 
Also can we add the features that you need in later stages without the need of changing module cmakes? Because this is what we need to achieve for API stability.

cochicde:
> @cochicde To be honest I never needed the feature to build executable and libs from one configuration. So for me it would be fine if the target forte would be one of the three options. Would this solve the issue? 

I would rather keep this functionality. It's easier for testing and releasing, since you can have all in one cmake setup, and i think the maintenance cost is even cheaper, as the logic depends directly on the selected cmake variables.

> Also can we add the features that you need in later stages without the need of changing module cmakes? Because this is what we need to achieve for API stability.
I'll be fine with that. I can do that later.  But which exactly API stability are we talking about that would be affected by this?

jwalt:
> I built Wago PFC200 on this branch and realized that `.../fordiac-fbe/modules/EclipseGeneratedFBs/CMakeLists.txt` needs to be updated, since some commands are not available anymore. I changed the last lines that it seemed to work for me:

FBE's custom exported modules code isn't updated yet to this new code, so it is to be expected that it doesn't work with it, just like all other external modues fail. This needs to be done after this has been merged.

jwalt:
> I was trying to build the C_INTERFACES and started working on the `CMakeLists.txt` files, but I see that now each module adds itself to the compilation by linking directly to `forte`. This is a problem, since this is assuming we are building only one executable/library. This affects also the static/shared library separation we had before. Previously everything compilation unit was added to the so called `forte-lite` and then executables and libraries were defined out of it.
> 
> By having everthing directly added to `forte`, libraries are harder to configure. For example, the arch libraries add also `main.cpp` to forte, which is not valid for libraries.
> 
> I'm wondering also, is it the responsability of the module to add themselves to the overall compilation? Isn't clear that the container module/folder "pulls" the library out of the submodules/subfolders and adds them to an intermediate library or similar? Like, the `com` folder would create `com` library which links to the `HTTP`, `modbus`, `opc_ua` and so on, and then the overall `forte` would link only to `com`. I'd argue that the responsability of a folder/module is to create a library but not also link it.
> 
> As it is now, adding a new compilation artifact like a library, requires adding in every folder

This is by design. One build configuration is for one type of output, be it share or static library, or executable. Everything else is not portable anyways: Depending on compiler and compiler options, shared libraries need different compilation options than static libraries or executables. This cannot be managed within a single build configuration, and the old build system was a hack that really only was guaranteed to do the right thing if you were building a single target anyways.

Additionally, the C library build has not yet been ported over, so it is not available.

Regarding the style of how objects are added to the main output: this is required if you are building CMake OBJECT libraries, they do not support hierarchical composition. Using intermediate (static or shared) libraries leads to all sorts of complications with shared vs. static compilation, so that would increase complexity a lot. Also, I think it is much more maintainable if each module/subdir contains its entire logic (including the decision if and what to include in forte) in its cmake file. Because of the high variability of our code base (and because it was much easier to work with) I settled on this push style instead of a pull style.

mx990:
There are no tests found/executed during the CI build[^1]:
```
No tests were found!!!
```

[^1]: https://github.com/eclipse-4diac/4diac-forte/actions/runs/15970302694/job/45039819047?pr=499#step:8:13

mx990 (approved):
I was now able to successfully compile and run the tests. The CI build also seems to be working properly. I am therefore fine with merging this for now and resolving any remaining issues afterwards.

azoitl (approved):

franz-hoepfinger-4diac:
@all, thanks for that great work. 

just one Remark: 
we have a lot of 

`SET_PROPERTY(GLOBAL APPEND PROPERTY FORTE_INCLUDE_DIRECTORIES ${TARGET_PROJECT_ROOT}/Core/Inc)`

e.g. in `/4diac-forte/buildsupport/toolchains/toolchain-eliteboard.cmake`

as the proprietary `FORTE_INCLUDE_DIRECTORIES` no more exist, 
we need some Directions how to re-write the Files in the toolchains Directory. 

this is mainly used for Embedded Targets, like ESP32, STM32 ... i dont think it is used for Linux Targets. 

Thanks a lot.

mx990:
Please use `target_include_directories`[^1] to add include directories to the `forte-global` target, for example: https://github.com/eclipse-4diac/4diac-forte/blob/9409e88a33304a973db4b8eabea627f0fa684884/src/CMakeLists.txt#L14

[^1]: https://cmake.org/cmake/help/v3.25/command/target_include_directories.html

PR #503, opened 2025-06-12
CI automatic format check

### LINKED ISSUE #379: [CI] Formatting check

gravemalte:
This PR introduces a GitHub workflow which check if every file is well formatted according to our `.clang-format`, closes #379

Only one problem remains. As of now many files are not formatted correctly, so if this PR gets merged many errors will appear.
First, we need to introduce a formatted state of the source code.

Feedback welcome.

~To test it on your system, run the following script:~

```sh
#!/bin/bash
set -e

SRC_DIR="src"

cd "$(git rev-parse --show-toplevel)"
find "$SRC_DIR" -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)' -exec clang-format --dry-run --Werror "{}" +
```

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,5 +1,11 @@
 src/core/trace/classic/generated/*
-src/core/trace/replay_debugging/*
-src/modules/*
+src/core/trace/replay_debugging/generated/*
 src/stdfblib/*
+src/stdfblib/*/*
 src/com/opc_ua/FBs/*
+src/com/opc_ua/FBs/*/*
+src/modules/*/*
+src/modules/*/*/*
+src/modules/*/*/*/*
+src/modules/*/*/*/*/*
+src/modules/*/*/*/*/*/*
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,4 @@
+rc/core/trace/classic/generated/*
+src/core/trace/replay_debugging/*
+src/modules/*
+src/stdfblib/*
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,4 +1,4 @@
-rc/core/trace/classic/generated/*
+src/core/trace/classic/generated/*
 src/core/trace/replay_debugging/*
 src/modules/*
 src/stdfblib/*
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -2,3 +2,4 @@ rc/core/trace/classic/generated/*
 src/core/trace/replay_debugging/*
 src/modules/*
 src/stdfblib/*
+src/com/opc_ua/FBs/*
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -22,7 +22,9 @@ jobs:
         id: find_sources
         run: |
           FILES=$(find src/ -type f \( -name "*.cpp" -o -name "*.cc" -o -name "*.c" -o -name "*.h" -o -name "*.hpp" \))
-
+          echo "files<<EOF" >> $GITHUB_OUTPUT
+          echo "$FILES" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
       - name: Run clang-format check
         run: |
           echo "${{ steps.find_sources.outputs.files }}" | xargs clang-format --dry-run --Werror
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -17,15 +17,11 @@ jobs:
 
       - name: Install clang-format
         run: sudo apt-get install -y clang-format
-
-      - name: Find all source
-        id: find_sources
+      
+      - name: Run clang-format in dry mode
         run: |
-          FILES=$(find src/ -type f \( -name "*.cpp" -o -name "*.cc" -o -name "*.c" -o -name "*.h" -o -name "*.hpp" \))
-          echo "files<<EOF" >> $GITHUB_OUTPUT
-          echo "$FILES" >> $GITHUB_OUTPUT
-          echo "EOF" >> $GITHUB_OUTPUT
-      - name: Run clang-format check
-        run: |
-          echo "${{ steps.find_sources.outputs.files }}" | xargs clang-format --dry-run --Werror
+          FILES=$(find src/ -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)$')
+          echo "Checking formatting on:"
+          echo "$FILES"
+          echo "$FILES" | xargs clang-format --dry-run --Werror
 
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,32 @@
+name: Clang Format Check
+
+on:
+  push:
+    branches: [ "release", "develop", "freeze", "feature/**" ]
+  pull_request:
+    branches: [ "release", "develop", "freeze", "feature/**" ]
+
+jobs:
+  clang-format:
+    name: Check Clang Format
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install clang-format
+        run: sudo apt-get install -y clang-format
+
+      - name: Find all source
+        id: find_sources
+        run: |
+          FILES=$(find src/ -type f \( -name "*.cpp" -o -name "*.cc" -o -name "*.c" -o -name "*.h" -o -name "*.hpp" \))
+          echo "files<<EOF" >> $GITHUB_OUTPUT
+          echo "$FILES" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+      - name: Run clang-format check
+        run: |
+          echo "${{ steps.find_sources.outputs.files }}" | xargs clang-format --dry-run --Werror
+
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -22,9 +22,6 @@ jobs:
         id: find_sources
         run: |
           FILES=$(find src/ -type f \( -name "*.cpp" -o -name "*.cc" -o -name "*.c" -o -name "*.h" -o -name "*.hpp" \))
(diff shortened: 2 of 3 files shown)


cochicde:
you could have the script as file and in the repo, and then the CI will just call this.
 This will allows anyone to easily check formatting, or even add to their pre-commit hooks

ernstblechaPT:
Special Case: a way for handling the code generated by/for barectf (introduced in https://github.com/eclipse-4diac/4diac-forte/pull/417/files#diff-8721332ceca3b73ff2cafccb38e2b11759b5b5304eb654270732a666648aab05) would be needed.

There are multiple options I could think of:
 - exclusion list for specific files (e.g. `src/core/trace/classic/generated/barectf-bitfield.h`, `src/core/trace/classic/generated/barectf.c`, `src/core/trace/classic/generated/barectf.h`, ...)
 - exclusion list for directories (e.g. `src/core/trace/classic/generated/`, ...)
 - reformatting the generated code according to the formatting guidelines directly after running the generator in the cmake (adds a dependency on clang-format for generating the barectf configuration)

ernstblechaPT:
Looked at the result of the check: right now only a single line of the generated barectf-code does not match the formatting rules; 

```
2025-06-12T18:28:59.5995752Z src/core/trace/classic/generated/barectf.c:43:41: error: code should be clang-formatted [-Wclang-format-violations]
2025-06-12T18:28:59.5996906Z     (_at_var) = ((_at_var) + ((_align) -1)) & -(_align);                                                               \
2025-06-12T18:28:59.5997831Z                                         ^
2025-06-12T18:28:59.5998958Z src/core/trace/classic/generated/barectf.c:43:57: error: code should be clang-formatted [-Wclang-format-violations]
2025-06-12T18:28:59.5999999Z     (_at_var) = ((_at_var) + ((_align) -1)) & -(_align);                                                               \
2025-06-12T18:28:59.6000591Z                                                         ^
2025-06-12T18:28:59.6468811Z src/core/trace/replay_debugging/generated/barectf.c:43:41: error: code should be clang-formatted [-Wclang-format-violations]
2025-06-12T18:28:59.6469970Z     (_at_var) = ((_at_var) + ((_align) -1)) & -(_align);                                                               \
2025-06-12T18:28:59.6470579Z                                         ^
2025-06-12T18:28:59.6471470Z src/core/trace/replay_debugging/generated/barectf.c:43:57: error: code should be clang-formatted [-Wclang-format-violations]
2025-06-12T18:28:59.6472583Z     (_at_var) = ((_at_var) + ((_align) -1)) & -(_align);                                                               \
```

sadly that part is a preprocessor macro - reformatting this might actually break things in unexpected ways

gravemalte:
This complicates the whole thing of course, but I could imagine the following.

We have a script (which also calls the CI see @cochicde suggestion) and this script reads a `.ignore_format`, in which
all files are specified, which should be ignored by the formatter. I think this combines the best of both approaches.

ernstblechaPT:
Maybe the easiest way would be to have a `.clang-format-ignore` file for the two affected directories

https://clang.llvm.org/docs/ClangFormat.html#clang-format-ignore

that would also avoid the case where automatic reformatting touches the generated code and turns the generated code into derived work

gravemalte:
@ernstblechaPT thanks for point that out. I wasn't aware that `clang-format` has an ignore file. That simplifies it

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Install clang-format
+        run: sudo apt-get install -y clang-format
+      
+      - name: Run clang-format in dry mode
+        run: |
+          FILES=$(find src/ -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)$')
+          echo "Checking formatting on:"
+          echo "$FILES"
+          echo "$FILES" | xargs clang-format --dry-run --Werror
```
ernstblechaPT:
```suggestion
          find src -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)$' -exec clang-format --dry-run -Werror "{}" +
```

This would avoid the pitfalls around filenames with spaces; If you want to still print all the files to be checked you would need to run a second `find` like
```suggestion
          echo "Checking formatting on:"
          find src -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)$'
          find src -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)$' -exec clang-format --dry-run -Werror "{}" +
```

ernstblechaPT (commented):
If you don't mind you could add the ignore files `src/core/trace/classic/generated/.clang-format-ignore` and `src/core/trace/replay_debugging/generated/.clang-format-ignore` containing something along 
```
# ignore all files in this directory
*
# ingore all files in subdirectories
**/*
```

EDIT: you could also just put a single ignore-file `src/core/trace/.clang-format-ignore` with 
```
classic/generated/*
classic/generated/**/*
replay_debugging/generated/*
replay_debugging/generated/**/*
```

mx990:
I would prefer a single `.clang-format-ignore` in the top-level folder, just as with `.gitignore`.

gravemalte:
I will add this over the weekend.

gravemalte:
So I outsource the job into a script, including a `.clang-format-ignore`.

Still, there are some parts of the code that are not properly formatted. You can try it locally on your machine by running the script.

checkstyle.sh
```diff
@@ -0,0 +1,43 @@
...
+    fi
+done
+
+if [[ ${#CHECK_FILES[@]} -eq 0 ]]; then
+    echo "✅ No files to check"
+    exit 0
+fi
+
+echo "🔍 Checking formatting on ${#CHECK_FILES[@]} files..."
+clang-format --dry-run --Werror "${CHECK_FILES[@]}"
```
ernstblechaPT:
```suggestion
set -e

SRC_DIR="src"

cd "$(git rev-parse --show-toplevel)"
find "$SRC_DIR" -type f -regex '.*\.\(c\|cpp\|cc\|h\|hpp\)' -exec clang-format --dry-run --Werror "{}" +
```

with the modified `.clang-format-ignore` this does the same except for not showing the filecount

.clang-format-ignore
```diff
@@ -0,0 +1,5 @@
+src/core/trace/classic/generated/*
+src/core/trace/replay_debugging/*
+src/modules/*
+src/stdfblib/*
+src/com/opc_ua/FBs/*
```
ernstblechaPT:
```suggestion
src/core/trace/classic/generated/*
src/core/trace/replay_debugging/generated/*
src/stdfblib/*
src/stdfblib/*/*
src/com/opc_ua/FBs/*
src/com/opc_ua/FBs/*/*
src/modules/*/*
src/modules/*/*/*
src/modules/*/*/*/*
src/modules/*/*/*/*/*
src/modules/*/*/*/*/*/*
```

I'm not really happy with having to explicitly state a number of subdirectories - but ubuntu:latest (currently ubuntu 24.04.2) only provides clang 18.x

with clang 20.x (available on ubuntu 25.04) globstar will be available to simplify this down to 

```
src/core/trace/classic/generated/*
src/core/trace/replay_debugging/generated/*
src/stdfblib/*
src/stdfblib/**/*
src/com/opc_ua/FBs/*
src/com/opc_ua/FBs/**/*
src/modules/**/*
```

checkstyle.sh
gravemalte: Much simpler yeah. I will work it in.

.clang-format-ignore
gravemalte: Didn't know that, can you send me a link where you read it? I searched the documentation of clang-format. I assumed it's the same behaviour like it's on my system (clang edge from the Arch repo)

ernstblechaPT: https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html#clang-format

Only figured that out when my changes would not work on Ubuntu and looked up the differences between the 18.x Documentation and the 20.x Documentation

 - https://releases.llvm.org/18.1.0/tools/clang/docs/ClangFormat.html#clang-format-ignore (missing the "globstar" information)
 - https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormat.html#clang-format-ignore (missing the "globstar" information)
 - https://releases.llvm.org/20.1.0/tools/clang/docs/ClangFormat.html#clang-format-ignore
 - https://clang.llvm.org/docs/ClangFormat.html#clang-format-ignore

gravemalte:
@ernstblechaPT Pushed your changes.

ernstblechaPT (approved):

ernstblechaPT:
I only now realized that @mx990 already propsed something similar in https://github.com/eclipse-4diac/4diac-forte/pull/419

mx990:
The workflow in https://github.com/eclipse-4diac/4diac-forte/pull/419 relies on a GitHub action[^1], which supports both `clang-format` and `clang-tidy`. There were still lots of problems reported at the time, so I had left it as a draft and did not have the time to look into it further.

I have now changed https://github.com/eclipse-4diac/4diac-forte/pull/419 to only perform the `clang-format` check and it seems to also honor `.clang-format-ignore`. It still requires a dependency on a GitHub action but has the advantage of providing file annotations and a summary report.

I am personally open towards either solution. What do you think?

[^1]: https://github.com/cpp-linter/cpp-linter-action

gravemalte:
@mx990 I think your work in #419 is better suited. You use clang-tidy which is more powerful than my naive approach with just using `clang-format` for checking the formatting. But @cochicde suggestion is useful I think, when you want to check your style in a pre-commit hook an external script would be useful.

EDIT: Maybe increase the level of warning to an `-Werror` for the formatting in #419

gravemalte:
Closed because of #419

PR #513, opened 2025-06-20
replace makros from external event handler by CRTP

moniwe:
I tested to replace the makros within the external event handler by the CRTP Curiously Recurring Template Pattern for the XQuery module. For that I introduced the _IExternalEventHandler_ that I do not have to change all classes. The _IExternalEventHandler_ could be removed when all uses of the makros haven been changed. Please check if you agree that markos are old fashioned and that CRTP is a proper replacement.


azoitl:
In order to simplify our CMake code generation we had to rework the external event handler registration. This is done in ##566. With this changes this PR is now obsolete.

PR #517, opened 2025-06-23
Fix compile error in the IEC61131-3 package using the debug log level

gravemalte:
This little PR fixes the compile issue in the IEC61131-3 package, in which the old interface API was used in two `DEVLOG_DEBUG(...)` statements.

The CI didn't catch the error because the binaries are not build using the debug level.

### DIFF
--- a/src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
+++ b/src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
@@ -83,7 +83,7 @@ bool GEN_ADD::createInterfaceSpec(const char *paConfigString, SFBInterfaceSpec &
   pcPos++;
   // we have an underscore and it is the first underscore after AND
   unsigned int numDIs = static_cast<unsigned int>(forte::core::util::strtoul(pcPos, nullptr, 10));
-  DEVLOG_DEBUG("DIs: %d;\n", paInterfaceSpec.getNumDIs());
+  DEVLOG_DEBUG("DIs: %d;\n", numDIs);
 
   if (numDIs < 2) {
     return false;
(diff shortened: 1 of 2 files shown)


src/modules/IEC61131-3/Arithmetic/GEN_ADD_fbt.cpp
```diff
@@ -83,7 +83,7 @@ bool GEN_ADD::createInterfaceSpec(const char *paConfigString, SFBInterfaceSpec &
   pcPos++;
   // we have an underscore and it is the first underscore after AND
   unsigned int numDIs = static_cast<unsigned int>(forte::core::util::strtoul(pcPos, nullptr, 10));
-  DEVLOG_DEBUG("DIs: %d;\n", mDInputs);
+  DEVLOG_DEBUG("DIs: %d;\n", paInterfaceSpec.getNumDIs());
```
azoitl: at that point this will always return 0, you need to use `numDIs` instead.

src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
```diff
@@ -52,7 +52,7 @@ bool CGenBitBase::createInterfaceSpec(const char *paConfigString, SFBInterfaceSp
     pcPos++;
     // we have an underscore and it is the first underscore after AND
     numDIs = static_cast<TPortId>(forte::core::util::strtoul(pcPos, nullptr, 10));
-    DEVLOG_DEBUG("DIs: %d;\n", paInterfaceSpec.mNumDIs);
+    DEVLOG_DEBUG("DIs: %d;\n", paInterfaceSpec.getNumDIs());
```
azoitl: at that point this will always return 0, you need to use `numDIs` instead.

azoitl (changes requested):

azoitl (approved):

PR #522, opened 2025-06-28
Add OPC UA property mapping to support IEC 61499 data in HMI tools

m-meingast:
This PR is on top of freeze, because the Message CFB is currently not exportable on develop. This can be changed to develop, if necessary.

### DIFF
--- a/src/com/opc_ua/opcua_ac_layer.h
+++ b/src/com/opc_ua/opcua_ac_layer.h
@@ -55,22 +55,6 @@ class COPC_UA_AC_Layer : public COPC_UA_Layer {
       PathToInstance
     };
 
-    static const std::string scmAlarmTypeBrowsePath;
-    static const std::string scmAlarmConditionName;
-
-    static char smEmptyString[];
-    static char smEnabledState[];
-    static char smEnableStateProperty[]; // This is needed to avoid potential delete-subscription error with HMI tools
-    static char smActiveState[];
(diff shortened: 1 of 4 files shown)


src/com/opc_ua/opcua_ac_layer.h
```diff
@@ -60,6 +60,7 @@ class COPC_UA_AC_Layer : public COPC_UA_Layer {
 
     static char smEmptyString[];
```
azoitl: I recently learned that private static is a bit an anti-pattern in C++. Better move these into an anonymous workspace in the CPP file.

azoitl (changes requested):

src/com/opc_ua/opcua_ac_layer.h
m-meingast: Done

m-meingast: This does also apply to the private static callback functions, doesn't it?

azoitl: yes for these as well. But I guess this is better for another pr.

azoitl (approved):

azoitl:
can you please also port it back to develop

m-meingast:
Yes, I will put it onto develop

azoitl:
> Yes, I will put it onto develop

thx

PR #456, opened 2025-05-18
replace CSinglyLinkedList by vector, fix send/receive

moniwe:

PR #528, opened 2025-07-04
replace CSinglyLinkedList by vector in FDSelectHandler

franz-hoepfinger-4diac:
class CSinglyLinkedList' is deprecated: Use STL containers instead. Preferably std::vector [-Wdeprecated-declarations]

### DIFF
--- a/core/arch/common/src/fdselecthand.cpp
+++ b/core/arch/common/src/fdselecthand.cpp
@@ -63,8 +63,7 @@ namespace forte::arch {
 
       if (retval > 0) {
         mSync.lock();
-        TConnectionContainer::Iterator itEnd(mConnectionsList.end());
-        for (TConnectionContainer::Iterator itRunner = mConnectionsList.begin(); itRunner != itEnd;) {
+        for (auto itRunner = mConnectionsList.begin(); itRunner != mConnectionsList.end();) {
           // need to retrieve the callee as the iterator may get invalid in the recvDat function below in case of
           // connection closing
           com_infra::CComCallback *callee = itRunner->mCallee;
@@ -96,7 +95,7 @@ namespace forte::arch {
     {
       util::CCriticalRegion criticalRegion(mSync);
       TConnContType stNewNode = {paFD, paComCallback};
-      mConnectionsList.pushBack(stNewNode);
+      mConnectionsList.push_front(stNewNode);
       mConnectionListChanged = true;
     }
     if (!isAlive()) {
@@ -107,23 +106,7 @@ namespace forte::arch {
   void CFDSelectHandler::removeComCallback(TFileDescriptor paFD) {
     util::CCriticalRegion criticalRegion(mSync);
 
-    TConnectionContainer::Iterator itRunner(mConnectionsList.begin());
-    TConnectionContainer::Iterator itRefNode(mConnectionsList.end());
-    TConnectionContainer::Iterator itEnd(mConnectionsList.end());
-
-    while (itRunner != itEnd) {
-      if (itRunner->mSockDes == paFD) {
-        if (itRefNode == itEnd) {
-          mConnectionsList.popFront();
-        } else {
-          mConnectionsList.eraseAfter(itRefNode);
-        }
-        break;
-      }
-
-      itRefNode = itRunner;
-      ++itRunner;
-    }
+    std::erase_if(mConnectionsList, [&paFD](TConnContType x) { return (x.mSockDes == paFD); });
 
     mConnectionListChanged = true;
   }
@@ -131,11 +114,10 @@ namespace forte::arch {
   CFDSelectHandler::TFileDescriptor CFDSelectHandler::createFDSet(fd_set *mFDSet) {
     TFileDescriptor nRetVal = scmInvalidFileDescriptor;
     FD_ZERO(mFDSet);
-    TConnectionContainer::Iterator itEnd(mConnectionsList.end());
-    for (TConnectionContainer::Iterator itRunner = mConnectionsList.begin(); itRunner != itEnd; ++itRunner) {
-      FD_SET(itRunner->mSockDes, mFDSet);
-      if (itRunner->mSockDes > nRetVal || scmInvalidFileDescriptor == nRetVal) {
-        nRetVal = itRunner->mSockDes;
+    for (auto itRunner : mConnectionsList) {
+      FD_SET(itRunner.mSockDes, mFDSet);
+      if (itRunner.mSockDes > nRetVal || scmInvalidFileDescriptor == nRetVal) {
+        nRetVal = itRunner.mSockDes;
       }
     }
     mConnectionListChanged = false;
--- a/core/arch/common/include/forte/arch/fdselecthand.h
+++ b/core/arch/common/include/forte/arch/fdselecthand.h
@@ -13,11 +13,12 @@
 
 #pragma once
 
-#include "../../core/src/fortelist.h"
 #include "forte/extevhan.h"
 #include "forte/arch/forte_thread.h"
 #include "forte/arch/forte_sync.h"
 #include "forte/arch/sockhand.h"
+
+#include <forward_list>
 
 namespace forte {
   namespace com_infra {
@@ -62,11 +63,9 @@ namespace forte {
             com_infra::CComCallback *mCallee;
         };
 
-        typedef CSinglyLinkedList<TConnContType> TConnectionContainer;
-
         TFileDescriptor createFDSet(fd_set *mFDSet);
 
-        TConnectionContainer mConnectionsList;
+        std::forward_list<TConnContType> mConnectionsList;
         CSyncObject mSync;
         bool mConnectionListChanged;
     };


franz-hoepfinger-4diac:
this was inspired by: 

https://github.com/eclipse-4diac/4diac-forte/pull/456  @moniwe 

https://github.com/eclipse-4diac/4diac-forte/pull/441 @cochicde

core/arch/common/src/fdselecthand.cpp
```diff
@@ -107,16 +107,16 @@ void CFDSelectHandler::addComCallback(TFileDescriptor paFD, forte::com_infra::CC
 void CFDSelectHandler::removeComCallback(TFileDescriptor paFD) {
   CCriticalRegion criticalRegion(mSync);
 
-  TConnectionContainer::Iterator itRunner(mConnectionsList.begin());
```
azoitl: please use std::vector::erase together with std::remove_if

src/arch/fdselecthand.cpp
```diff
@@ -63,8 +63,8 @@ void CFDSelectHandler::run() {
 
     if (retval > 0) {
       mSync.lock();
-      TConnectionContainer::Iterator itEnd(mConnectionsList.end());
-      for (TConnectionContainer::Iterator itRunner = mConnectionsList.begin(); itRunner != itEnd;) {
+      auto itEnd(mConnectionsList.end());
+      for (auto itRunner = mConnectionsList.begin(); itRunner != itEnd;) {
```
azoitl: Have you tested this loop with several different cases, especially with servers and clients where the other end is disconnecting, when several subscribers are listening? I don't think this loop will work in that case. That is the reason I have not yet replaced this single linked list with a std::vector. I'm not even sure if it is possible that we can use std::vector here.

core/arch/common/src/fdselecthand.cpp
```diff
@@ -131,8 +131,8 @@ void CFDSelectHandler::removeComCallback(TFileDescriptor paFD) {
 CFDSelectHandler::TFileDescriptor CFDSelectHandler::createFDSet(fd_set *mFDSet) {
   TFileDescriptor nRetVal = scmInvalidFileDescriptor;
   FD_ZERO(mFDSet);
-  TConnectionContainer::Iterator itEnd(mConnectionsList.end());
-  for (TConnectionContainer::Iterator itRunner = mConnectionsList.begin(); itRunner != itEnd; ++itRunner) {
```
azoitl: please use the more efficient for each loop:

for(auto it : mConnecitonList) {... }

src/arch/fdselecthand.cpp
franz-hoepfinger-4diac: as you already wrote: std::list is the right one.

franz-hoepfinger-4diac:
at least: this does not work any bit. 

it results in Console flooded with 

`ERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file numberERROR: T#50000000000: Select failed: Bad file number`

franz-hoepfinger-4diac:
i need to fix the linter Problems. 
at least this one working now without Error Messages.

franz-hoepfinger-4diac:
so, it is Green, and Basic Case working now.

src/arch/fdselecthand.h
```diff
@@ -69,7 +69,7 @@ class CFDSelectHandler : public CExternalEventHandler, private CThread {
         forte::com_infra::CComCallback *mCallee;
     };
 
-    typedef CSinglyLinkedList<TConnContType> TConnectionContainer;
+    using TConnectionContainer = std::list<TConnContType>;
```
mx990: I would suggest using the singly-linked `std::forward_list`[^1] instead of the doubly-linked `std::list`.

[^1]: https://en.cppreference.com/w/cpp/container/forward_list.html

src/arch/fdselecthand.cpp
```diff
@@ -107,35 +107,19 @@ void CFDSelectHandler::addComCallback(TFileDescriptor paFD, forte::com_infra::CC
...
-        mConnectionsList.eraseAfter(itRefNode);
-      }
-      break;
-    }
-
-    itRefNode = itRunner;
-    ++itRunner;
-  }
+  mConnectionsList.erase(std::remove_if(mConnectionsList.begin(), mConnectionsList.end(),
+                                        [&paFD](TConnContType x) { return (x.mSockDes == paFD); }));
```
mx990: I would suggest using the combined `std::erase_if`[^1] here.

[^1]: https://en.cppreference.com/w/cpp/container/forward_list/erase2.html

mx990 (changes requested):

src/arch/fdselecthand.h
franz-hoepfinger-4diac: OK, but a push_back is no more possible then. 
should we use push_front then ?????? 
otherwise we must iterate through all Elements, which is inefficient, and add it at the last Element ?

core/arch/common/src/fdselecthand.cpp
```diff
@@ -96,7 +96,7 @@ void CFDSelectHandler::addComCallback(TFileDescriptor paFD, forte::com_infra::CC
   {
     CCriticalRegion criticalRegion(mSync);
     TConnContType stNewNode = {paFD, paComCallback};
-    mConnectionsList.pushBack(stNewNode);
+    mConnectionsList.push_front(stNewNode);
```
franz-hoepfinger-4diac: @mx990 @azoitl 
can this be true ? 
std::forward_list has no push_back Method, only push_front, 
by Design, 
to replace here the old pushBack to the new push_front ? 

maybe in the old CSinglyLinkedList it was just a matter of naming, 
so it is functional equivalent ?

franz-hoepfinger-4diac: I am currently testing this, and i found no negative Effects up to now. (which means nothing, but just saying)

MartinMelikMerkumians: The CSinglyLinkedList was a single linked list, it just saved the first and last element, which allowed a push back

franz-ms-muc: so, we need a push_back here ? 
then how to do ?
go back to a list instead of forward_list ?

azoitl: No we don't. This list does not care about any ordering. It has to main task: check if one of the sockets has received data, or if a connection is closed to remove it. So if we only have a push_front it is totally fine.

src/arch/fdselecthand.cpp
```diff
@@ -63,8 +63,8 @@ void CFDSelectHandler::run() {
 
     if (retval > 0) {
       mSync.lock();
-      TConnectionContainer::Iterator itEnd(mConnectionsList.end());
-      for (TConnectionContainer::Iterator itRunner = mConnectionsList.begin(); itRunner != itEnd;) {
+      auto itEnd(mConnectionsList.end());
```
azoitl: please move the `mConnectionList.end()` into the for loop.

src/arch/fdselecthand.h
```diff
@@ -69,7 +69,7 @@ class CFDSelectHandler : public CExternalEventHandler, private CThread {
         forte::com_infra::CComCallback *mCallee;
     };
 
-    typedef CSinglyLinkedList<TConnContType> TConnectionContainer;
+    using TConnectionContainer = std::forward_list<TConnContType>;
```
azoitl: I don't think that in modern C++ the using makes much sense anymore. The type is only needed in one place and that is in line 76. Therefore I think it is better to remove this using statement and use `std::forward_list<TConntType>` in line 76.

azoitl (changes requested):

core/arch/common/src/fdselecthand.cpp
diplfranzhoepfinger: > No we don't. This list does not care about any ordering. It has to main task: check if one of the sockets has received data, or if a connection is closed to remove it. So if we only have a push_front it is totally fine.

yes, i have the same Feeling, i did not see any difference if there is a pushBack or a push_front.

MartinMelikMerkumians: I just tried to explain why the singlylinkedlist supported both. The order of connections is irrelevant

azoitl (approved):

franz-hoepfinger-4diac:
@azoitl what Tests needed to do next to make this a Request rather than a Draft ?

azoitl:
> @azoitl what Tests needed to do next to make this a Request rather than a Draft ?

@franz-hoepfinger-4diac I good test would be if you have an several subscriberes, at least one client and one server in the same device which are constantly recieving data. Then you connect and disconnect several times monitoring and see if 4diac FORTE is correctly receiving the data for the subscribers, client and server. If you want to improve your test further you can also deinitialize the opposite ends for the client and the server and see if the blocks handles this correctly and if the other blocks are disturbed. Also for the server block it should be possible to connect and disconnect several times. But this would be tested with the monitoring connection also.

franz-hoepfinger-4diac:
ah. in the "same" Device is important right?

azoitl:
> ah. in the "same" Device is important right?

yes because than the receiving handling is in the same list and that is what you want to test.

mx990:
I have rebased this PR onto current develop. @franz-hoepfinger-4diac Could you please test with the current version?

core/arch/common/src/fdselecthand.cpp
```diff
@@ -93,49 +92,27 @@ namespace forte::arch {
...
 
   void CFDSelectHandler::addComCallback(TFileDescriptor paFD, com_infra::CComCallback *paComCallback) {
-    {
-      util::CCriticalRegion criticalRegion(mSync);
-      TConnContType stNewNode = {paFD, paComCallback};
-      mConnectionsList.pushBack(stNewNode);
-      mConnectionListChanged = true;
-    }
-    if (!isAlive()) {
-      this->start();
```
azoitl, line 103: please bring back the starting of the handler

azoitl (changes requested):

core/arch/common/src/fdselecthand.cpp
franz-hoepfinger-4diac: should be back !

mx990:
I have found no issues during testing. I believe we can finally merge this now.

mx990 (approved):

azoitl (approved):

PR #533, opened 2025-07-11
Add new communication infrastructure

mx990:
Add new communication infrastructure with the following features:
- different event configurations,
- polled and asynchronous operation,
- minimum and maximum receive size,
- type-safe communication channels with self-registration.

There is currently a top-layer `raw` channel that sends and receives single `STRING` values. There is also a bottom-layer `null` channel that ignores any sent data and receives empty data, mainly for testing purposes. They can be used with the channel ID `raw[].null[]`. There is further a generic `SEND_RECV` communication block for individual send and receive on duplex communication channels.

This is currently in an early prototyping stage, so there may still be significant changes towards the final version.


src/core/com/common.h
```diff
@@ -0,0 +1,118 @@
...
+     * \brief Not enough resources
+     *
+     * There are not enough resources.
+     */
+    Overflow = 12,
+  };
+
+  struct ComChannelDescriptor {
+      CStringDictionary::TStringId mChannel{CStringDictionary::scmInvalidStringId};
+      std::string mAddress;
```
azoitl: could this also be a string_view?

src/core/com/channel.h
```diff
@@ -0,0 +1,123 @@
...
+      /**
+       * \brief Open a channel with the given address and descriptors
+       * \param paAddress The address to open the channel with
+       * \param paDescriptors Communication channel descriptors for configuration
+       * \return The result of opening the channel
+       * \retval Ok The channel was opened successfully
+       * \retval InvalidId The address or descriptors are invalid
+       * \retval Overflow There are not enough resources
+       */
+      virtual ComResult open(std::string_view paAddress, std::span<ComChannelDescriptor> paDescriptors) = 0;
```
azoitl: is the name address not a bit limiting. How I understand the setup it is similar to the existing communication infrastructure all configuration data for a ComChannel or? Would then something like `paChannelConfData` be more appropriate?

src/core/com/common.h
mx990: I was initially concerned that someone would keep a `ComChannelDescriptor` past the lifetime of the original string, but it should be fine with the current implementation.

src/core/com/channel.h
mx990: Makes sense. I have renamed it to `paConfigString`.

azoitl (approved):

PR #540, opened 2025-07-16
Enable system tests in cmake-multi-platform workflow

mx990:
Enable system tests in `cmake-multi-platform` workflow and remove separate `systemtests` workflow. This ensures that the system tests run on all platforms and avoids a duplicate build for the `systemtests` workflow.


systemtests/CMakeLists.txt
```diff
@@ -179,6 +170,8 @@ SET_TESTS_PROPERTIES(Test_valid_invalid_parameter PROPERTIES PASS_REGULAR_EXPRES
 ##############################################
 
 # Simple test with multi devices
-forte_add_multi_systemtests(SimpleMultiTest 10 1 multiDevice1.fboot "-c localhost:61500" multiDevice2.fboot " " multiDevice3.fboot "-c localhost:61501")
+if (NOT FORTE_ARCHITECTURE STREQUAL "Win32")
```
cochicde: we should clarify why this is not being executed in Windows, so whenever the obstacle is removed, we can enable this

mx990: That is a good point. I have created a new issue https://github.com/eclipse-4diac/4diac-forte/issues/544 to keep track of the problem and added a comment referring to it. I have also changed the condition to only disable the test when compiling with MSVC, since it does work locally when compiling with MinGW gcc on Windows.

cochicde (approved):

PR #546, opened 2025-07-21
Renames data input prefix for consistency

franz-hoepfinger-4diac:
Updates the prefix used for generating data input names from "IN_" to "IN"
to maintain consistency with other naming conventions.

this is because AND has Inputs IN1 and IN2 and so on.


azoitl (approved):

azoitl:
@franz-hoepfinger-4diac does this require any changes in the typelib?

franz-hoepfinger-4diac:
> @franz-hoepfinger-4diac does this require any changes in the typelib?

unless we merge this PR: yes
if we merge this PR: no

this was the Reason for this PR, i think changing the Typelib ist not preffered, right ?

franz-hoepfinger-4diac:
However, as we already said some time ago: there is some inconsistency if using _ or not, and if start at 0 or at 1 .
even in the IEC 61499 Standard.

PR #548, opened 2025-07-21
Adds access to DI via index

franz-hoepfinger-4diac:
Adds a method to access a specific digital input (DI)
by its index. This allows for more flexible access to
the input data of the function block.

### DIFF
--- a/src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
+++ b/src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
@@ -78,6 +78,10 @@ bool CGenBitBase::createInterfaceSpec(const char *paConfigString, SFBInterfaceSp
   return false;
 }
 
+CIEC_ANY *CGenBitBase::getDI(size_t paDINum) {
+  return &mGenDIs[paDINum];
+}
+
 CIEC_ANY *CGenBitBase::getDO(size_t paDONum) {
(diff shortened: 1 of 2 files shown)


src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
```diff
@@ -36,7 +36,7 @@ CGenBitBase::CGenBitBase(const CStringDictionary::TStringId paInstanceNameId, fo
 }
 
 void CGenBitBase::readInputData(TEventID) {
-  for (TPortId i = 0; i < getGenDINums(); ++i) {
+  for (TPortId i = 0; i < getFBInterfaceSpec().getNumDIs(); ++i) {
```
azoitl: I would prefer to keep the getGenDINums here this gives more flexibility for future changes.

azoitl (changes requested):

src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
diplfranzhoepfinger: Agree 👍

franz-hoepfinger-4diac: resolved !

azoitl (approved):

PR #551, opened 2025-07-23
Remove "-forte_[PORT]" suffix from custom OPC UA server hostname

m-meingast:
Discussed in #260

(https://github.com/eclipse-4diac/4diac-forte/pull/260#discussion_r1796460458)

### DIFF
--- a/src/com/opc_ua/opcua_local_handler.cpp
+++ b/src/com/opc_ua/opcua_local_handler.cpp
@@ -161,8 +161,6 @@ void COPC_UA_Local_Handler::generateServerStrings(TForteUInt16 paUAServerPort, U
   paServerStrings.mHostname = std::string("opc.tcp://");
 #ifdef FORTE_COM_OPC_UA_CUSTOM_HOSTNAME
   paServerStrings.mHostname.append(FORTE_COM_OPC_UA_CUSTOM_HOSTNAME);
-  paServerStrings.mHostname.append("-"s);
-  paServerStrings.mHostname.append(helperBuffer);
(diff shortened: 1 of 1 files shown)


src/com/opc_ua/opcua_local_handler.cpp
```diff
@@ -161,8 +161,6 @@ void COPC_UA_Local_Handler::generateServerStrings(TForteUInt16 paUAServerPort, U
   paServerStrings.mHostname = std::string("opc.tcp://");
 #ifdef FORTE_COM_OPC_UA_CUSTOM_HOSTNAME
   paServerStrings.mHostname.append(FORTE_COM_OPC_UA_CUSTOM_HOSTNAME);
-  paServerStrings.mHostname.append("-"s);
-  paServerStrings.mHostname.append(helperBuffer);
```
cochicde, line 165: as the helperBuffer is not being used under this #ifdef anymore, you could move it inside the #else

m-meingast: Do you mean the initialisation of the helperBuffer? Because it is used in line 157 as well if Multicast is enabled.

```c
#ifdef FORTE_COM_OPC_UA_MULTICAST
  paServerStrings.mMdnsServerName = std::string(helperBuffer);
#endif //FORTE_COM_OPC_UA_MULTICAST
```

cochicde: oh, I missed that. Great, then let's merge this!

cochicde (approved):

PR #553, opened 2025-07-24
Fixes parsing of configuration strings

franz-hoepfinger-4diac:
Corrects the parsing logic for configuration strings
to correctly identify the number of digital inputs.
The code now uses `strchr` to find the first underscore
instead of `strrchr` which finds the last underscore.
This ensures the correct number of digital inputs is parsed
when the config string contains multiple underscores.

### DIFF
--- a/src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
+++ b/src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
@@ -46,11 +46,11 @@ void CGenBitBase::writeOutputData(TEventID) {
 }
 
 bool CGenBitBase::createInterfaceSpec(const char *paConfigString, SFBInterfaceSpec &paInterfaceSpec) {
-  const char *pcPos = strrchr(paConfigString, '_');
+  const char *pcPos = strchr(paConfigString, '_');
   size_t numDIs = 0;
   if (nullptr != pcPos) {
-    pcPos++;
-    // we have an underscore and it is the first underscore after AND
+    pcPos++; // move after underscore
+    //we have an underscore and it is the first underscore after AND/OR/XOR
     numDIs = static_cast<TPortId>(forte::core::util::strtoul(pcPos, nullptr, 10));
     DEVLOG_DEBUG("DIs: %d;\n", numDIs);
   } else {


src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
```diff
@@ -46,11 +46,11 @@ void CGenBitBase::writeOutputData(TEventID) {
 }
 
 bool CGenBitBase::createInterfaceSpec(const char *paConfigString, SFBInterfaceSpec &paInterfaceSpec) {
-  const char *pcPos = strrchr(paConfigString, '_');
```
azoitl, line 49: As your are touching that code and improving it, it would be great if you could convert it to C++ style by using a string_view and string_view's search infrastructure.

azoitl (changes requested):

mx990:
Could you please elaborate more on why this change is necessary? I am only aware of the generic `AND`, `OR`, and `XOR` blocks that inherit from `CGenBitBase`. Why would there be multiple underscores? If there were multiple underscores, I would still expect the number to appear after the *last* underscore (e.g., `BIT_OP_3`), as it was implemented before?

azoitl:
@mx990 thx for asking this. I misread the commit message. The change is definitely wrong. For these kind of FBs it should always be the last underscore.

src/modules/IEC61131-3/BitwiseOperators/genbitbase_fbt.cpp
franz-hoepfinger-4diac: yes, Agree !

franz-hoepfinger-4diac:
> @mx990 thx for asking this. I misread the commit message. The change is definitely wrong. For these kind of FBs it should always be the last underscore.

if so, the Comment was wrong. 


i was thinking about the recommendation in the IEC 61499: 

where you can have a AND_3_BOO, a AND_3_BYTE and so on. 
then we need the first underscore. 

i will wait with the rework until this is clear.

azoitl:
In that case I would prefere to close this pr, we have so many have backed prs that makes it hard for me to get an overview. we can reopen it again when we start working on the core problem.

PR #557, opened 2025-07-30
fix cannot convert 'addrinfoexA* const' to 'const ADDRINFOEXA*'

moniwe:
I built with 4diac-fbe on Windows 11 and got

`cannot convert 'addrinfoexA* const' to 'const ADDRINFOEXA*'`

### DIFF
--- a/src/arch/win32/com/impl/net.cpp
+++ b/src/arch/win32/com/impl/net.cpp
@@ -33,7 +33,7 @@ SOCKET forte::com::impl::net::open(const std::string_view paConfigString, const 
     return -1;
   }
 
-  for (const ADDRINFOEXA *rp = result; rp != nullptr; rp = rp->ai_next) {
+  for (const ADDRINFOEXA *rp = result; rp != nullptr; rp = reinterpret_cast<ADDRINFOEXA *>(rp->ai_next)) {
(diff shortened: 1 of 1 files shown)


mx990:
There was a bug in the `addrinfoexA` struct from MinGW-w64[^1][^2]. I am not sure which version is used by 4diac-fbe, but I would rather upgrade to a newer version than working around a MinGW bug in 4diac if we can avoid it.

[^1]: https://github.com/mingw-w64/mingw-w64/issues/9
[^2]: https://github.com/mingw-w64/mingw-w64/commit/619e9cde0542fc2a20afbc7b9c844b20eacbe2e0

moniwe:
as far as I have seen the toolchain is downloaded from [SourceForge 4diac-fbe/release-2025-01](https://sourceforge.net/projects/fordiac/files/4diac-fbe/release-2025-01/) and this seems to contain the version before the commit you provided. There is a newer version release-2025-02 I will check this one, and then close this PR when it works. Thanx for finding this.

moniwe:
[SourceForge 4diac-fbe/releae-2025-02](https://sourceforge.net/projects/fordiac/files/4diac-fbe/release-2025-02/) also contains the problematic header file. Others replaced this header file by that from the link above to get it work. Therefore the provided fix of this pull request, will not be needed after updating the toolchain.

Issue #558, opened 2025-07-30
Rework factories to use self-registration


mx990:
Rework factories to use self-registration instead of generating fixed factories at build time. This allows to add additional implementations to a pre-built forte, simplifies build configuration, and also avoids complex code generation.

PR #559, opened 2025-07-30
Rework device factory to use self-registration

mx990:
Part of https://github.com/eclipse-4diac/4diac-forte/issues/558


src/arch/utils/mainparam_utils.cpp
```diff
@@ -20,7 +20,9 @@
 #endif
 
 #include "stdfblib/ita/ForteBootFileLoader.h"
-#include "generated/devicefactory.h"
+#include "core/devicefactory.h"
+
+#include <ranges>
```
cochicde: is this include really needed here?

src/core/devicefactory.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <unordered_map>
+
+class CDevice;
+
+namespace forte::core {
+
+  class DeviceEntry;
+
+  class DeviceFactory final {
```
cochicde: This is a class with just static members. Maybe it could be jsut a namespace, and you won't need the friends in it or in the DeviceEntry

cochicde (commented):
Looks good! We might need still some #if/#else because in most cases there's only once device, and this factory was created for research reasons and for playing around with different devices. Therefore you have extra code when you're using only one device (which is most of the time)

src/arch/utils/mainparam_utils.cpp
mx990: Yes, it is needed for `std::views::keys` in line 81 below[^1].

[^1]: https://github.com/eclipse-4diac/4diac-forte/pull/559/files#diff-ffb126b1b740dbd3f5d9635abdf609a1cd0155ddb46752a15222c30ddea013edR81

src/core/devicefactory.h
mx990: Good point. I initially had private members in `DeviceFactory` but later moved them out of the class and left the class as-is.

mx990 (commented):
> We might need still some #if/#else because in most cases there's only once device, and this factory was created for research reasons and for playing around with different devices. Therefore you have extra code when you're using only one device (which is most of the time)

Its origins notwithstanding, this factory could also be quite useful in practical applications. For example, users will be able to add additional devices to a pre-compiled forte by simply linking with a static or shared library. I also believe the overhead should be quite negligible overall, both in terms of binary size and performance.

azoitl:
@cochicde has a point here. While for Windows and Linux (our main targets) the library approach makes a lot of sense, we have users on smaller embedded targets for the messages I see also in matrix every byte of heap seems to matter. So it would be great to have an advanced option for them beeing able to disable these factories and work with just a single instance.

mx990:
I have now changed it to a generic factory, which also supports selecting a fixed implementation at compile time. In that case, it should have zero runtime overhead. You can give it a try with:
```cmake
-DFORTE_DEVICE_FIXED_IMPL=RMT_DEV -DFORTE_DEVICE_FIXED_INCLUDE=stdfblib/ita/RMT_DEV.h
```

src/arch/utils/mainparam_utils.cpp
mx990: Now it is not needed here anymore.

cochicde (approved):
Looks good!

azoitl (approved):

PR #560, opened 2025-07-30
Rework timer handler factory to use self-registration

mx990:
Also fix missing DI connections in `FakeTimeDev`.

Part of #558


cochicde (commented):
looks good! I have the same comments as in the other factory PR.

src/arch/utils/mainparam_utils.cpp
```diff
@@ -59,6 +62,7 @@ void listHelp() {
 #ifdef FORTE_TRACE_CTF
   printf("%-20s Set the output directory for TRACE_CTF\n", "  -t <directory>");
 #endif // FORTE_TRACE_CTF
+  printf("%-20s Set the timer to be used\n", "  -T TIMER");
```
ernstblechaPT: I think to be consistent this should be 

```suggestion
  printf("%-20s Set the timer to be used\n", "  -T <timer>");
```

but this was also missed for the FORTE_MULTIPLE_DEVICES option

mx990: changed both

azoitl (approved):

PR #562, opened 2025-08-04
Fix tests in release mode

mx990:
The `BOOST_ASSERT` macro is disabled when building in *Release* mode, which causes tests that rely on side effects to fail. This replaces `BOOST_ASSERT` with `BOOST_REQUIRE` to ensure that tests also work when building in *Release* mode.


cochicde (approved):
LGTM

azoitl (approved):

PR #563, opened 2025-08-08
Add separate typedef for instance names

mx990:
Add a separate typedef for instance names to allow for easier refactoring in the future.


azoitl (commented):
How do we handle that in a TNameIdentifier we have FB instance name and pin instance names mixed. Especially when addressing struct members it is even more complex.

mx990:
I believe we should then also change `TNameIdentifier` to use `InstanceNameId` and convert to a `StringId` when actually looking up the pin or member. After all, it should not matter if we do the lookup a little later?

PR #568, opened 2025-08-17
Adds pin validation to I/O handles

franz-hoepfinger-4diac:
Adds a mechanism to validate I/O pins, ensuring that only valid pins are used in process interface function blocks.

This prevents errors caused by using invalid pins and provides a way to indicate a pin is not usable.

### DIFF
--- a/src/core/io/mapper/io_handle.cpp
+++ b/src/core/io/mapper/io_handle.cpp
@@ -9,7 +9,7 @@
  * Contributors:
  *   Johannes Messmer - initial API and implementation and/or initial documentation
  *   Jose Cabral - Cleaning of namespaces
- *   Franz Höpfinger - adding "pin Valid" 
+ *   Franz Höpfinger - adding "pin Valid"
  *******************************************************************************/
 
 #include "core/io/mapper/io_handle.h"
@@ -19,7 +19,10 @@
 
 using namespace forte::core::io;
 
-IOHandle::IOHandle(IODeviceController *paController, IOMapper::Direction paDirection, CIEC_ANY::EDataTypeID paType, bool paPinValid) :
+IOHandle::IOHandle(IODeviceController *paController,
+                   IOMapper::Direction paDirection,
+                   CIEC_ANY::EDataTypeID paType,
+                   bool paPinValid) :
     mController(paController),
     mType(paType),
     mDirection(paDirection),
--- a/src/core/io/mapper/io_handle.cpp
+++ b/src/core/io/mapper/io_handle.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 - 2018 Johannes Messmer (admin@jomess.com), fortiss GmbH
+ * Copyright (c) 2016, 2025 Johannes Messmer (admin@jomess.com), fortiss GmbH, HR Agrartechnik GmbH
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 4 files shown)


azoitl (commented):
Now that I see the code I better understand what you want. I don't think this should be added in this form. The reason is that IOHandles should be created for available pins. So if a handle as an invalid pin it shouldn't even be created. As such then the FB would not find a handle which indicates that.

franz-ms-muc:
> So if a handle as an invalid pin it shouldn't even be created. 

ah, then we would need a static Method, which is called before the Constructor, and it returns if this Pin is valid, 
and the Controller would not create the Handle then, right ? 

> As such then the FB would not find a handle which indicates that.

the IX or QX Block then would still say "Waiting for Handle"!

azoitl:
> > So if a handle as an invalid pin it shouldn't even be created.
> 
> ah, then we would need a static Method, which is called before the Constructor, and it returns if this Pin is valid, and the Controller would not create the Handle then, right ?

How I understand it is that you anyhow know at creation time if the pin is valid. Otherwise you could not pass it to the constructor of the IOHandle. If you can do that you can also check that at the point where you are either creating the handles or the descriptors. And not create them at all. 

Just for the future if you are using fields that are only settable in the constructor it could make sense to declare them const.

> 
> > As such then the FB would not find a handle which indicates that.
> 
> the IX or QX Block then would still say "Waiting for Handle"!

Yes it would. But if your IO configuration resource is up and running this means that for the given name no pin is existing. The reason it is called waiting for handle is that we are not implying any order on the initialization of IOHandles or IO FBs. And if the IO FB is first it has to wait if a handle will still be created.

franz-ms-muc:
> > > So if a handle as an invalid pin it shouldn't even be created.
> > 
> > 
> > ah, then we would need a static Method, which is called before the Constructor, and it returns if this Pin is valid, and the Controller would not create the Handle then, right ?
> 
> How I understand it is that you anyhow know at creation time if the pin is valid. Otherwise you could not pass it to the constructor of the IOHandle. If you can do that you can also check that at the point where you are either creating the handles or the descriptors. And not create them at all.

no, i set it to false in the constructor, call the constructor with the Default Argument, and the Handler knowns: noop, and set it to false. it is working at least. 


> 
> Just for the future if you are using fields that are only settable in the constructor it could make sense to declare them const.
> 

yep, but i set it true first, and in the Constructor Body: false: 

see: 

```
IOHandleESP32DigitalO::IOHandleESP32DigitalO(
    logiBUSIODeviceController *paDeviceCtrl,
    forte::core::io::IOMapper::Direction paDirection,
    CIEC_ANY::EDataTypeID paDataType,
    logiBUSIOEnums::DigitalOutPinNumber paPin
    )
    : forte::core::io::IOHandle(static_cast<forte::core::io::IODeviceController*>(paDeviceCtrl), paDirection, paDataType)
    , gpio_num(GPIO_PIN(static_cast<EPinDigitalOut>(paPin)))
  {
    if(gpio_num > GPIO_NUM_NC)
    {
      gpio_reset_pin((gpio_num_t)gpio_num);
      /* Set the GPIO as a push/pull output */
      gpio_set_direction((gpio_num_t)gpio_num, GPIO_MODE_OUTPUT);
    } else {
    	mPinValid = false;
    }
  }
```

azoitl:
I still think that performing these checks upfront during handle creation is the better place. Therefore I'm closing this PR now.

PR #591, opened 2025-09-03
Rework OPC UA to use self registration

mx990:
Rework OPC UA to use self registration and also fix warnings and string id usage.


m-meingast (commented):
While not being part of this PR in particular, the OPC UA A&C Layer does not compile anymore since the last changes. This is due to a typecast from `CIEC_ANY*` to `CIEC_STRING*`, which the compiler marks as a cast to an incomplete type. I suppose it is just the `#include 'core/datatypes/forte_string.h'` missing.

mx990:
Thank you for noticing! I added the missing includes.

m-meingast (approved):

azoitl (approved):

PR #623, opened 2025-09-26
Updates FreeRTOS includes to use paths

franz-hoepfinger-4diac:
Updates FreeRTOS include statements to use path-based includes.
This change aligns the include style with the FreeRTOS directory structure,
improving project organization and portability.

### DIFF
--- a/core/arch/freeRTOS/forte_architecture_time.cpp
+++ b/core/arch/freeRTOS/forte_architecture_time.cpp
@@ -12,8 +12,8 @@
  *  Martin Jobst - add high-resolution realtime clock fallback
  *******************************************************************************/
 
-#include <FreeRTOS.h>
-#include <task.h>
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
 
 #include "forte/arch/forte_architecture_time.h"
 #include "forte/util/forte_constants.h"
--- a/core/arch/freeRTOS/forte_sem.h
+++ b/core/arch/freeRTOS/forte_sem.h
@@ -13,8 +13,8 @@
 #ifndef SRC_ARCH_FREERTOS_SEMAPHORE_H_
 #define SRC_ARCH_FREERTOS_SEMAPHORE_H_
 
-#include <FreeRTOS.h>
-#include <semphr.h>
+#include <freertos/FreeRTOS.h>
+#include <freertos/semphr.h>
 #include "forte/datatype.h"
 
 namespace forte {
--- a/core/arch/freeRTOS/forte_sync.h
+++ b/core/arch/freeRTOS/forte_sync.h
@@ -12,8 +12,8 @@
 #ifndef _SYNC_H_
 #define _SYNC_H_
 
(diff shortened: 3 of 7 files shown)


azoitl (changes requested):
I checked severa official freeRTOS examples. The way we had it before is the way how a normal freeRTOS application is including it. I think you need to fix your include path setup instead.

gravemalte:
I can confirm this. FreeRTOS always excepts an include path that is not prefixed. See also the official examples:

https://github.com/FreeRTOS/FreeRTOS-SMP-Demos/blob/main/FreeRTOS/Demo/CORTEX_A53_64-bit_UltraScale_MPSoC/RTOSDemo_A53/src/Blinky_Demo/main_blinky.c#L67

https://github.com/FreeRTOS/FreeRTOS-Community-Supported-Demos/blob/main/C2000_F2838x_C28x_CCS/freertos_ex1_c28x_port_val.c#L97

franz-hoepfinger-4diac:
> I checked severa official freeRTOS examples. The way we had it before is the way how a normal freeRTOS application is including it. I think you need to fix your include path setup instead.

OK, other speaking: 

in CMAKE based Projects the **usual** way is to make a Definition: 

```
target_include_directories(FreeRTOS-Kernel PUBLIC
    Source/include
)
```

and include it like this: 


```
#include "freertos/FreeRTOS.h" // Liegt in <Pfad>/Source/include/freertos/FreeRTOS.h
#include "freertos/task.h"      // Liegt in <Pfad>/Source/include/freertos/task.h
```

see directory structure here: 
https://github.com/espressif/esp-idf/blob/25c40d4563be7d4caf375b59c52283ff80c62d08/components/freertos/FreeRTOS-Kernel/include/freertos/FreeRTOS.h


this way it is done in Espressif ESP-IDF, 
and this way is also syncronous to the new 4diac include File Organisation.

gravemalte:
> and include it like this:
> 
> ```
> #include "freertos/FreeRTOS.h" // Liegt in <Pfad>/Source/include/freertos/FreeRTOS.h
> #include "freertos/task.h"      // Liegt in <Pfad>/Source/include/freertos/task.h
> ```
> 
> see directory structure here: https://github.com/espressif/esp-idf/blob/25c40d4563be7d4caf375b59c52283ff80c62d08/components/freertos/FreeRTOS-Kernel/include/freertos/FreeRTOS.h
> 
> this way it is done in Espressif ESP-IDF, and this way is also syncronous to the new 4diac include File Organisation.

On top of this is a CMake layer that adds, in fact both include structures as you can see here:

https://github.com/espressif/esp-idf/blob/25c40d4563be7d4caf375b59c52283ff80c62d08/components/freertos/CMakeLists.txt#L128


Looking at the FORTE `arch` layer, we also don't provide for example an `#include <posix/myheader.h>` include prefix. And FreeRTOS is specified as an arch.

franz-hoepfinger-4diac:
i will close it then. And live with the double include Path ...

PR #627, opened 2025-09-26
Use fixed bitset to store force flags and reset all forced outputs after event execution

mx990:
Use fixed bitset to store force flags with a configurable maximum supported absolute port index for forcing data pins. Reset all forced outputs after event execution.

### DIFF
--- a/core/include/forte/funcbloc.h
+++ b/core/include/forte/funcbloc.h
@@ -235,6 +235,9 @@ namespace forte {
             mEventMonitorCount[paEIID]++;
           }
           executeEvent(paEIID, paExecEnv);
+          if (mForces.any()) [[unlikely]] {
+            resetForcedOutputs();
+          }
         }
       }
 
@@ -447,9 +450,6 @@ namespace forte {
       void writeData(const TAbsDataPortNum paAbsDataPortNum, T &paValue, U &paConn) {
         if (!getForce(paAbsDataPortNum)) [[likely]] {
           paConn.writeData(paValue);
-        } else {
-          // when forcing we write back the value from the connection to keep the forced value on the output
-          paConn.readData(paValue);
         }
 #ifdef FORTE_TRACE_CTF
         traceWriteData(paAbsDataPortNum - mInterfaceSpec.getNumDIs(), paValue);
@@ -541,7 +541,8 @@ namespace forte {
        *
        * \param paECET the event chain execution thread this FB was invoked from
        * \param paEIID Event input ID where event occurred.
-       * \param paExecEnv Event chain execution environment the FB will be executed in (used for adding output events).
+       * \param paExecEnv Event chain execution environment the FB will be executed in (used for adding output
+       * events).
        *
        *
        */
@@ -560,6 +561,10 @@ namespace forte {
       virtual void writeOutputData(TEventID paEO) = 0;
 
       bool setForce(TAbsDataPortNum paAbsDataPortNum, bool paForceValue);
+
+      /*!\brief Function resetting the values of forced outputs to the value from the output connection.
+       */
+      void resetForcedOutputs();
 
     public:
       CFunctionBlock(const CFunctionBlock &) = delete;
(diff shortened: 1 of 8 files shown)


core/include/forte/funcbloc.h
```diff
@@ -561,6 +560,21 @@ namespace forte {
 
       bool setForce(TAbsDataPortNum paAbsDataPortNum, bool paForceValue);
 
+      /*!\brief Function resetting the values of forced outputs to the value from the output connection.
+       */
+      void resetForcedOutputs() {
+        if (mForces.any()) [[unlikely]] {
```
azoitl: I would suggest to move this if up to the receiveInputEvent. In my performance measurements it was a bit faster, and we could move the implementation of resetForceOutputs into the cpp. Reducing the size of funcbloc.h.

core/include/forte/funcbloc.h
```diff
@@ -442,8 +444,8 @@ namespace forte {
        * \param paConn Connection to write into.
        */
       template<typename T, typename U>
-      void writeData(TAbsDataPortNum paAbsDataPortNum, T &paValue, U &paConn) {
-        if (!mForces[paAbsDataPortNum]) {
+      void writeData(const TAbsDataPortNum paAbsDataPortNum, T &paValue, U &paConn) {
+        if (!getForce(paAbsDataPortNum)) {
```
azoitl: should we put an unlikly here as well?

core/include/forte/funcbloc.h
```diff
@@ -423,11 +425,11 @@ namespace forte {
        * \param paValue Variable to read into.
        * \param paConn Connection to read from.
        */
-      void readData(TPortId paAbsDataPortNum, CIEC_ANY &paValue, const CDataConnection *const paConn) {
+      void readData(const TPortId paAbsDataPortNum, CIEC_ANY &paValue, const CDataConnection *const paConn) {
         if (!paConn) {
           return;
         }
-        if (!mForces[paAbsDataPortNum]) {
+        if (!getForce(paAbsDataPortNum)) {
```
azoitl: should we put an unlikly here as well?

azoitl (changes requested):

core/include/forte/funcbloc.h
mx990: yes, although it needs to be `[[likely]]` here

core/include/forte/funcbloc.h
mx990: yes, although it needs to be `[[likely]]` here

core/include/forte/funcbloc.h
mx990: good point, done

core/include/forte/funcbloc.h
azoitl: yes, I should have written the equivalent of unlikely.

core/include/forte/funcbloc.h
```diff
@@ -343,8 +348,8 @@ namespace forte {
 
       bool setForce(std::span<const StringId> paNameList, bool paForce) override;
 
-      bool getForce(TAbsDataPortNum paAbsDataPortNum) const {
-        return mForces[paAbsDataPortNum];
+      [[nodiscard]] constexpr bool getForce(const TAbsDataPortNum paAbsDataPortNum) const {
```
azoitl: would isForced be better show the intent?

core/include/forte/funcbloc.h
```diff
@@ -343,8 +348,8 @@ namespace forte {
...
+        return paAbsDataPortNum < mForces.size() && mForces[paAbsDataPortNum];
```
azoitl, line 352: do we have any chance to get rid of the first part of the condition?
getForce is on the critical path and that would make as faster then before.

mx990: Both `paAbsDataPortNum` and `mForces.size()` are known on the critical path, so the optimizer is able to skip that part (and even skip the rest if it is larger).

core/include/forte/funcbloc.h
mx990: makes sense, done

core/include/forte/funcbloc.h
azoitl: Most probably you are right. In all my measurements the difference where smaller then the measurement errors. I think we are now so fast that I have to rework my measurement setups.

azoitl (approved):

PR #630, opened 2025-09-30
Add script to combine static forte libraries

mx990:
Add script to combine static forte libraries for easier consumption by non-CMake build systems. The script combines all `libforte*.a` files in the current working directory to a `libforte-all.a`, except any existing `libforte-all.a` itself. It is intended to be run inside the `lib` directory of the install prefix after installing forte and all external modules.

### DIFF
--- a/buildsupport/combine-static-libs.sh
+++ b/buildsupport/combine-static-libs.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+#############################################################################
+# Copyright (c) 2025 Martin Erich Jobst
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0.
+#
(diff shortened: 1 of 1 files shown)


cochicde:
can't this be done in CMake?

jwalt:
The original version of the cmake refactor did this in pure cmake, and it did not need -fwhole-archive either. We could fix this and at the same time lower our cmake minimum version if we returned to something like that approach for the static linking case.

mx990:
I realize that it is possible to achieve the same result in pure CMake[^1]. As stated initially, this was intended specifically for *non-CMake build systems*.

> The original version of the cmake refactor did this [...]

No, it did not. When building external modules separately from forte, you would have also ended up with separate static libraries. Aside from that, I believe this is hardly the place to start a fundamental discussion about our CMake build system, so I'm closing this for now.

[^1]: see also https://gitlab.kitware.com/cmake/cmake/-/issues/19224

PR #631, opened 2025-10-01
Changed hardware lib to system lib

azoitl:
updated all cmake files and namespaces


stdfblib/system/src/CommandParser.cpp
```diff
@@ -11,15 +11,15 @@
  *    - initial API and implementation and/or initial documentation
  *******************************************************************************/
 
-#include "CommandParser.h"
+#include "../../system/src/CommandParser.h"
```
mx990: I believe the `../../system/src/` is not necessary here (same in other files).

mx990 (changes requested):

stdfblib/system/src/CommandParser.cpp
azoitl: Interesting. I don't remember that I did that. Looks like some auto clean-up magic from Eclipse CDT or clangd. Will revert that.

azoitl: fixed

stdfblib/system/src/replay/ReplayMGR.h
```diff
@@ -13,20 +13,19 @@
...
 
-#include "../OPCUA_MGR.h"
-#include "deviceReplayer.h"
-
 #include <functional>
 #include <string>
 #include <vector>
 #include <optional>
+#include "../OPCUA_MGR.h"
+#include "../replay/deviceReplayer.h"
```
mx990:
```suggestion
#include "deviceReplayer.h"
```

mx990 (changes requested):

mx990 (approved):

Issue #639, opened 2025-10-04
Impossible to link target 'forte' because the link item 'forte-iec61131-3',


franz-hoepfinger-4diac:
``
CMake Error at CMakeLists.txt:66 (add_library):
  Impossible to link target 'forte' because the link item 'forte-iec61131-3',
  specified without any feature or 'DEFAULT' feature, has already occurred
  with the feature 'WHOLE_ARCHIVE', which is not allowed.
``

franz-hoepfinger-4diac:
Pls. if there are some CMAKE Experts, give guidance how to handle this in the most elegant way. 
the PR https://github.com/eclipse-4diac/4diac-forte/pull/640 works, but looks not really easy to me.

mx990:
This should have been fixed with https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9510.

Which version of CMake are you using? If you are using a CMake toolchain file, could you please provide it?

franz-hoepfinger-4diac:
CMAKE 

[toolchain-esp32s3-HutschienenmopedXL.cmake.txt](https://github.com/user-attachments/files/22709291/toolchain-esp32s3-HutschienenmopedXL.cmake.txt)

3.30.2

franz-ms-muc:
i did not see this anymore !

PR #640, opened 2025-10-04
Fixes static linking of utils and IEC61131-3

franz-hoepfinger-4diac:
Ensures correct linking of the utils and IEC61131-3 libraries when building statically.
This prevents linker errors when these libraries are used in other parts of the system.
Solves: https://github.com/eclipse-4diac/4diac-forte/issues/639

Solves #639


mx990:
I believe when using the correct CMake property to deduplicate link libraries with `WHOLE_ARCHIVE`, see also PR https://github.com/eclipse-4diac/4diac-forte/pull/641, this is no longer necessary.

PR #643, opened 2025-10-07
Add improved array bounds checks

mx990:
Add improved array bounds checks that properly report both the index and bounds. Also follow the convention of checking only in `at()` to speed up access from within forte core functions.

### DIFF
--- a/core/include/forte/datatypes/forte_array.h
+++ b/core/include/forte/datatypes/forte_array.h
@@ -19,12 +19,13 @@
  *******************************************************************************/
 #pragma once
 
-#include <stddef.h>
-#include <inttypes.h>
+#include <cstddef>
+#include <cinttypes>
 
 #include "forte/datatypes/forte_any_derived.h"
 #include "forte/datatypes/forte_any_int.h"
 #include "forte/stringid.h"
+#include "forte/util/devlog.h"
 
 namespace forte {
   /** \brief A common supertype for all CIEC_ARRAY variants, providing the minimal interface an array must provide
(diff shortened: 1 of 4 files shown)


core/include/forte/datatypes/forte_array.h
```diff
@@ -147,6 +147,18 @@ namespace forte {
...
       static const intmax_t cmCollapseMaxSize = 100;
 
+      constexpr static void
+      checkBounds(const intmax_t paIndex, const intmax_t paLowerBound, const intmax_t paUpperBound, bool paValid) {
+        if (paIndex < paLowerBound || paIndex > paUpperBound || !paValid) {
+#if __cpp_exceptions
+          throw std::out_of_range("Array access to index " + std::to_string(paIndex) + " is out of bounds from " +
+                                  std::to_string(paLowerBound) + " to " + std::to_string(paUpperBound));
+#else
+          std::abort();
```
azoitl, line 160: should we print a log message in this case?

MartinMelikMerkumiansElectrify: Hi,

my idea of always having range checks was, that this is mandatory in IEC 61131 (at least according to the standard there should always be a check). To make sure that people do not bypass this checks by accident (or even on purpose) I made both versions checking the range, altough I knew thats not the standard in C++

mx990: I made a separate PR https://github.com/eclipse-4diac/4diac-ide/pull/1683 to use `at()` in the forte_ng exporter, which ensures that checks are used for IEC 61131-3 code, while allowing to avoid them in C++ code.

mx990: > should we print a log message in this case?

good point, done

azoitl (approved):

MartinMelikMerkumians (approved):

PR #644, opened 2025-10-08
Add extern template declarations for factories and registries

mx990:
This avoids limitations of the dynamic linker on Windows and also helps to reduce shared library sizes on other platforms.

### DIFF
--- a/com/opc_ua/include/forte/com/opc_ua/opcua_types.h
+++ b/com/opc_ua/include/forte/com/opc_ua/opcua_types.h
@@ -17,6 +17,12 @@
 
 #include <open62541.h>
 
-namespace forte::com_infra::opc_ua {
-  using OPC_UA_External_Types = util::Registry<"", const UA_DataType *>;
-}
+namespace forte {
+  namespace util {
+    extern template class Registry<"", const UA_DataType *>;
+  }
+
+  namespace com_infra::opc_ua {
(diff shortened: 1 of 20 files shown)


com/opc_ua/include/forte/com/opc_ua/opcua_types.h
```diff
@@ -17,6 +17,12 @@
 
 #include <open62541.h>
 
-namespace forte::com_infra::opc_ua {
-  using OPC_UA_External_Types = util::Registry<"", const UA_DataType *>;
-}
+namespace forte {
+  namespace util {
```
azoitl, line 21: for consistency shouldn't this Registry also be in the hook namespace?

mx990: No, this registry is actually in: https://github.com/eclipse-4diac/4diac-forte/blob/7a58e43ad943bfa3550550c4575ec0804dcd69e9/core/include/forte/util/registry.h#L20-L23
and the hook registry is in: https://github.com/eclipse-4diac/4diac-forte/blob/7a58e43ad943bfa3550550c4575ec0804dcd69e9/core/include/forte/util/hook.h#L18-L21

azoitl: thx for the clarification

PR #652, opened 2025-10-13
Set install RPATH in CMake presets

mx990:


gravemalte:
Dynamic linking works now, thanks.

gravemalte (approved):

azoitl (approved):

PR #657, opened 2025-10-14
Extended Logging of full Event Queue

azoitl:
To help identifying potential issues in networks the fb instance name and input event causing the overflow are logged.

### DIFF
--- a/core/src/ecet.cpp
+++ b/core/src/ecet.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2021 ACIN, Profactor GmbH, fortiss GmbH, Hit robot group
+ * Copyright (c) 2005, 2025 ACIN, Profactor GmbH, fortiss GmbH, Hit robot group
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -14,11 +14,8 @@
  *    zhaoxin
  *      -  fix that external event queue becomes event locker after it is full
  *******************************************************************************/
-#include "forte/config/forte_config.h"
 #include "forte/ecet.h"
-
 #include "forte/ecetfactory.h"
-#include "forte/esfb.h"
 #include "forte/util/criticalregion.h"
 #include "forte/util/devlog.h"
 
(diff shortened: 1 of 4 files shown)


core/src/ecet.cpp
```diff
@@ -105,4 +102,18 @@ namespace forte {
       default: break;
     }
   }
+
+  void CEventChainExecutionThread::logFullEventQueue(TEventEntry paEventToAdd) {
+    const SFBInterfaceSpec &interfaceSpec = paEventToAdd.getFB().getFBInterfaceSpec();
+    if (paEventToAdd.getPortId() > interfaceSpec.getNumEIs()) {
+      DEVLOG_ERROR("Event queue is full, event dropped! Function Block: %s, event num: %d (maxnum: %d)\n",
+                   paEventToAdd.getFB().getFullQualifiedApplicationInstanceName(',').c_str(), paEventToAdd.getPortId(),
```
mx990: Why are you using `,` instead of `.` here?

core/src/ecet.cpp
```diff
@@ -105,4 +102,18 @@ namespace forte {
...
+                   interfaceSpec.getNumDIs() - 1);
```
mx990: Should that not be `getNumEIs()`?

core/src/interfacespec.cpp
```diff
@@ -25,6 +25,16 @@
 using namespace forte::literals;
 
 namespace forte {
+
+  namespace {
+    StringId getPortNameId(TPortId paPortId, std::span<const StringId> paPortNames) {
+      if (paPortId > paPortNames.size()) {
+        return ""_STRID;
```
mx990: Since an empty string is a valid port name (i.e., function return variable), I would suggest using an invalid string id here:

```suggestion
        return {};
```

mx990 (changes requested):

core/src/interfacespec.cpp
```diff
@@ -25,6 +25,16 @@
 using namespace forte::literals;
 
 namespace forte {
+
+  namespace {
+    StringId getPortNameId(TPortId paPortId, std::span<const StringId> paPortNames) {
+      if (paPortId > paPortNames.size()) {
```
mx990:
```suggestion
      if (paPortId >= paPortNames.size()) {
```

core/src/ecet.cpp
```diff
@@ -105,4 +102,18 @@ namespace forte {
       default: break;
     }
   }
+
+  void CEventChainExecutionThread::logFullEventQueue(TEventEntry paEventToAdd) {
+    const SFBInterfaceSpec &interfaceSpec = paEventToAdd.getFB().getFBInterfaceSpec();
+    if (paEventToAdd.getPortId() > interfaceSpec.getNumEIs()) {
```
mx990:
```suggestion
    if (paEventToAdd.getPortId() >= interfaceSpec.getNumEIs()) {
```

mx990 (changes requested):

core/src/ecet.cpp
azoitl: I have no idea. Fixed

core/src/ecet.cpp
azoitl: yes, fixed

core/src/interfacespec.cpp
azoitl: fixed

core/src/ecet.cpp
azoitl: fixed

mx990 (approved):

PR #665, opened 2025-10-18
Fixes MUX/DEMUX type lookup

franz-hoepfinger-4diac:
Corrects the type lookup for MUX and DEMUX function blocks.

The original type lookup logic didn't properly resolve the correct type when using the base type names. This change adds a mapping to ensure that the lookup resolves to the correct derived type (STRUCT_MUX_1, STRUCT_DEMUX_1) when using the base type names (STRUCT_MUX, STRUCT_DEMUX).

This hotfix is necessary to ensure proper functionality of these function blocks.

### DIFF
--- a/core/src/typelib.cpp
+++ b/core/src/typelib.cpp
@@ -28,6 +28,12 @@ using namespace std::string_literals;
 
 namespace forte {
   namespace {
+
+    StringId demux = "eclipse4diac::convert::STRUCT_DEMUX"_STRID;
+    StringId demux_1 = "eclipse4diac::convert::STRUCT_DEMUX_1"_STRID;
+    StringId mux = "eclipse4diac::convert::STRUCT_MUX"_STRID;
+    StringId mux_1 = "eclipse4diac::convert::STRUCT_MUX_1"_STRID;
+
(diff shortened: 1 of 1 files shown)


azoitl (changes requested):
Thx for bringing this to our attention. However, I'm sorry this is not a fix for the problem. Struct muxers are registered as generic types in the typelib. If the type entry coming into 4diac FORTE is not correct it needs to be fixed on 4diac IDE side.

franz-hoepfinger-4diac:
> Thx for bringing this to our attention. However, I'm sorry this is not a fix for the problem. Struct muxers are registered as generic types in the typelib. If the type entry coming into 4diac FORTE is not correct it needs to be fixed on 4diac IDE side.

@azoitl i do not need to do any Change rigth ? 
this is a HACK and will not be merged. 

Fix need to go in IDE ?

azoitl:
I just debugged the code. Your problem must be something else. STruct mux demux typelookup works correctly.

franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-ide/pull/1725

https://github.com/eclipse-4diac/4diac-forte/pull/673

just to post the final Solution here.

PR #670, opened 2025-10-20
add export compiler info to cmake presets

azoitl:
updated eclipse project to cmake preset values.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -12,6 +12,10 @@
       "binaryDir": "${sourceDir}/build/${presetName}",
       "installDir": "${sourceDir}/install/${presetName}",
       "cacheVariables": {
+        "CMAKE_EXPORT_COMPILE_COMMANDS": {
+          "type": "BOOL",
+          "value": "ON"
+        },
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -12,6 +12,10 @@
       "binaryDir": "${sourceDir}/build/${presetName}",
       "installDir": "${sourceDir}/install/${presetName}",
       "cacheVariables": {
+        "CMAKE_EXPORT_COMPILE_COMMANDS": {
```
mx990, line 15: Should we add the `CMAKE_EXPORT_COMPILE_COMMANDS` to the CDT project configuration instead?

azoitl: I was not sure. I feared that I'm duplicating some stuff in the CDT project config if I also add release builds or builds for other platforms there. Also any clangd based editor (e.g., VSCode) would require this setting.

mx990: I guess it should not matter much.

mx990 (approved):

PR #678, opened 2025-10-24
Enables ESP32 module and integrates sdkconfig

franz-hoepfinger-4diac:
Configures the ESP32 module, allowing interaction with ESP32
devices through GPIO.

It includes the sdkconfig.h file from the specified
ESP32_SDK_CONFIG_DIR to configure the build, and establishes
the necessary library dependencies and include directories.

### DIFF
--- a/modules/esp32/CMakeLists.txt
+++ b/modules/esp32/CMakeLists.txt
@@ -14,22 +14,41 @@
 # FORTE_MODULE_ESP32  integration.
 # ############################################################################
 
-# option to enable OPC UA with cmake
+# option to enable ESP32 with cmake
 option(FORTE_MODULE_ESP32 "Interacting with GPIOs of esp32 devices via gpio" OFF)
 
 if (NOT FORTE_MODULE_ESP32)
     return()
 endif ()
 
-set(ESP32_SDK_CONFIG_DIR
-    ""
-    CACHE PATH "Path to the sdkconfig file to be used for this FORTE build"
-)
-
-if (NOT IS_DIRECTORY ${ESP32_SDK_CONFIG_DIR})
-    message(FATAL_ERROR "The ESP32_SDK_CONFIG_DIR (" ${ESP32_SDK_CONFIG_DIR}
+if (NOT IS_DIRECTORY ${CMAKE_SOURCE_DIR}/${ESP32_SDK_CONFIG_DIR})
+    message(FATAL_ERROR "The ESP32_SDK_CONFIG_DIR (" ${CMAKE_SOURCE_DIR}/${ESP32_SDK_CONFIG_DIR}
                         ") does not exist or is not set"
     )
 endif ()
 
-target_include_directories(forte-esp32 PUBLIC ${ESP32_SDK_CONFIG_DIR})
+
+
+add_library(forte-esp32
+					${CMAKE_SOURCE_DIR}/${ESP32_SDK_CONFIG_DIR}/sdkconfig.h)
+target_link_libraries(forte-esp32 PUBLIC forte-core)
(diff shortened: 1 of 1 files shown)


azoitl:
Is this really needed in that form? All it is doing is pulling in a header file that is not used by anything in the module. This sounds very fishy. Shouldn't this be with some io module. Especially why does that have to be something to be installed?

franz-ms-muc:
sdkconfig is needed for the whole Compile Process, especially also FreeRTOS. 
let's guide me how to do differently. 
i do not insist on this being a Module and lib.

azoitl:
`sdkconfig.h` is currently not used anywhere in 4diac FORTE. I fear to provide support you need to provide more information.

franz-ms-muc:
sdkconfig.h is pulled whenever you compile anything in ESP-IDF. Especially in FreeRTOS. if the Compiler does not find it, you get an Error. 
e.g. the Ticks per Second is defines in sdkcofig.h, but also Things like Network communication (like how many WIFI Buffers) etc..

azoitl:
So it is similar to freertosconfig.h, or? If yes we should treat it the same.

franz-ms-muc:
Yes, definitely. freertosconfig.h not existing in ESP-IDF, they mirror all settings to sdkconfig. 
how do you handle this file there ?

azoitl:
I guess similar as with the sysroot of freertos as additional include path for your build. So this is somthing for your preset or for you setupscript.

franz-ms-muc:
> I guess similar as with the sysroot of freertos as additional include path for your build. So this is somthing for your preset or for you setupscript.

how to do this ?

azoitl:
@franz-ms-muc I'm happy to google that for you.

The cleanest solution would be that the directory for your config file is already provided as part of the toolchain file. From what I found online you should have this as part of the espressif infrastructure. 

If that is not working for you then you can always add an additional include file path when you setup your project through this additional parameter:

 `-DCMAKE_CXX_FLAGS=-I ESP32_SDK_CONFIG_DIR`


I'm closing this PR and will submit a new PR to remove the ESP32 module.

franz-hoepfinger-4diac:
> @franz-ms-muc I'm happy to google that for you.
> 
> The cleanest solution would be that the directory for your config file is already provided as part of the toolchain file. From what I found online you should have this as part of the espressif infrastructure.
> 

nope, sdkconfig is a per-project config, not a Toolchain-wide config. 

> If that is not working for you then you can always add an additional include file path when you setup your project through this additional parameter:
> 
> `-DCMAKE_CXX_FLAGS=-I ESP32_SDK_CONFIG_DIR`

yes, in this case this could be done that way, as it is needed in every Module, you are right. 
a module-wise include is not neccessary here. 


> 
> I'm closing this PR and will submit a new PR to remove the ESP32 module.

franz-ms-muc:
`    -DCMAKE_CXX_FLAGS="-I ../Application/build_ESP32S3_no_lua/config/" \`
did do it, 
while 
`-DCMAKE_CXX_FLAGS=-I ESP32_SDK_CONFIG_DIR` did not work, as after remove the Module Cmake say 

`CMake Warning:
  Manually-specified variables were not used by the project:
ESP32_SDK_CONFIG_DIR
`


so , Thanks 
working solution found, not more effort than before. 
and less Modules.

franz-ms-muc:
my personal final Solution: 

replace
`    -DESP32_SDK_CONFIG_DIR="../Application/build_ESP32S3_no_lua/config/" \`

with 
`    -DCMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES="${PWD}/../../../Application/build_ESP32S3_no_lua/config/" \`

PR #680, opened 2025-10-25
Auto-detect FORTE_ARCHITECTURE for common cases

jwalt:
Currently, an empty cmake invocation fails due to no sane default for FORTE_ARCHITECTURE. This patch fixes this for the common desktop-os cases.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -251,5 +251,12 @@
         "windows"
       ]
     }
-  ]
+  ],
+  "vendor": {
+    "eclipse.dev/4diac/FBE/3.0": {
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


mx990:
I had done something similar not long ago while looking into combining the Linux and Windows presets. It adds alternate architecture names that correspond to those used by CMake and then uses the `CMAKE_SYSTEM_NAME` as the default for the `FORTE_ARCHITECTURE`. This has the advantage of being easily extendable for less common cases.

In the end, I did not pursue combining the presets further at the time, so I did not create a PR for it. I have now rebased my original branch and created a PR https://github.com/eclipse-4diac/4diac-forte/pull/681. Maybe you could have a look at it?

jwalt:
abandoning this is favour of #681

PR #681, opened 2025-10-25
Add alternate architecture names and use CMake system name as default architecture

mx990:


core/arch/CMakeLists.txt
```diff
@@ -56,11 +59,11 @@ add_subdirectory(win32)
 ###############################################################################
 # Configuration Options
 
-set(FORTE_ARCHITECTURE "None" CACHE STRING "Architecture for which to build 4diac FORTE")
+set(FORTE_ARCHITECTURE ${CMAKE_SYSTEM_NAME} CACHE STRING "Architecture for which to build 4diac FORTE")
```
jwalt: This is too late if no cached value exists, i.e. on the first run of cmake.  macro `forte_register_architecture` uses this value, so it must be defined before the various `add_subdirectory` calls.

core/arch/CMakeLists.txt
```diff
@@ -29,9 +29,12 @@ target_compile_options(forte-core PUBLIC
 ###############################################################################
 # Architectures
 
-macro(forte_register_architecture arch)
-    set_property(TARGET forte-core APPEND PROPERTY forte_architectures ${arch})
-    if (NOT FORTE_ARCHITECTURE STREQUAL "${arch}")
+macro(forte_register_architecture)
+    set(archs "${ARGN}")
```
jwalt: We need a solution for forte modules that test the architecture. Right now, they do something like https://github.com/eclipse-4diac/4diac-forte/blob/b28e6b5acc31cfe632673540b8e5c1f68b46a738/modules/gpiochip/CMakeLists.txt#L13

In fact, that example would work better with `Linux` instead of `Posix`, as it is not portable to other POSIX-compatible systems. With the recent release of OpenBSD for the Raspberry Pi, we may get users from other POSIX systems so I think we should take this distinction into account.

My suggestion therefore is: The code fragment in `gpiochip` should use CMAKE_SYSTEM_NAME instead, while this macro should normalize FORTE_ARCHITECTURE to a canonical name, which is either the traditional name (for less refactoring) or the name of the subdirectory (for better name/identifier match).

`list(GET archs 0 FORTE_ARCHITECTURE)` at the end of the macro would suffice for either version, assuming that the canonical arch name is always specified first.

The second variant would need something like this:
```
get_filename_component(archs ${CMAKE_CURRENT_SOURCE_DIR} NAME) 
list(APPEND archs ${ARGN})
```

jwalt (changes requested):

core/arch/CMakeLists.txt
mx990: Done. Thanks for noticing!

core/arch/CMakeLists.txt
mx990: That is a good point. I went with the first variant but used an explicit macro parameter for the canonical name.

Would you like to make the change to `gpiochip` in a separate PR, since you are the one who noticed it?

jwalt: Actually, I wonder if we should switch to standard CMake variables as far as possible. Something like `if (LINUX)`, since the toolchain clearly defines what system it supports. It won't build FreeBSD or Windows executables anyways. We then would only use FORTE_ARCHITECTURE for deeply embedded targets/RTOSes where `CMAKE_SYSTEM_NAME` is `Generic`, e.g. `arm-none-eabi` toolchains and the like. The case where a Linux toolchain is actually able to build something foreign is rare enough that we can require users to provide a modified CMake toolchain file.

Note that this suggestion would not affect this location, this would remain the same. But modules like GPIOCHIP would use cmake-standard variables.

mx990: I agree that we should prefer standard CMake variables where possible. I would suggest sticking to `CMAKE_SYSTEM_NAME`, though, since `UNIX`, `WIN32`, and the like may be easily confused with user-defined variables. Please create a separate issue or PR, where we can discuss further if necessary.

jwalt (approved):

PR #690, opened 2025-10-29
com/modbus: do not unconditionally add global include directories

jwalt:
this breaks cross-compilation


mx990 (approved):

azoitl (approved):
thx for finding this this should definitly not be there.

PR #696, opened 2025-10-29
modules: update target system guards to use cmake-standard conditions instead of FORTE_ARCHITECTURE

jwalt:
This addresses the discussion in #681. This uses the shorthand notation instead of CMAKE_SYSTEM_NAME for two reasons: It is much, much more readable, and (in the case of UNIX vs. LINUX/APPLE) it can express things that are difficult to express via CMAKE_SYSTEM_NAME.

Since the shorthands are officially documented, well known, and frequently used, there is no danger of confusion what they mean.


azoitl (approved):
Looks really much better. Can we do stuff like this then also for ather target oses like threadx, or zephyr?

jwalt:
> Looks really much better. Can we do stuff like this then also for ather target oses like threadx, or zephyr?

Standard cmake only has `CMAKE_SYSTEM_NAME STREQUAL "Generic"` for real-time OSes, but of course we could define our own shorthands, like `FORTE_THREADX`

azoitl:
> > Looks really much better. Can we do stuff like this then also for ather target oses like threadx, or zephyr?
> 
> Standard cmake only has `CMAKE_SYSTEM_NAME STREQUAL "Generic"` for real-time OSes, but of course we could define our own shorthands, like `FORTE_THREADX`

Good to know. This will then be something to look into when we rework the one or the other os.

mx990 (approved):

PR #698, opened 2025-10-30
Allow to use installed modbus lib

azoitl:
By adding the include prefix to the C++ files one can also easily use pre-installed modbus libraries.


mx990 (approved):

jwalt (approved):
Note that this will break FBE in the sense that the tiny update I will need to make to FBE-3.0 results in that it cannot build modbus for older FORTE releases. I have no idea if that works anyway, so it is probably not a practical issue.

azoitl:
@jwalt as we did so many major changes in 4diac FORTE 3.0 I'm fine that FBE is also not backwards compatible. With all the great changes we have I would also not recommend to anyone using pre 3.0 version from now on.

PR #699, opened 2025-10-30
core/arch/posix: make compatible to MacOS

jwalt:
This uses the POSIX architecture for MacOS, since it is very close. No need for a different arch anymore. Additionally, MacOS now has a serial comm layer, which it didn't have before.


core/arch/posix/CMakeLists.txt
```diff
@@ -23,7 +23,6 @@ target_link_libraries(forte-core PUBLIC
 set(FORTE_POSIX_GENERATE_MAP_FILE FALSE CACHE BOOL "Enable the generation of map files")
 mark_as_advanced(FORTE_POSIX_GENERATE_MAP_FILE)
 target_link_options(forte-core PUBLIC
-        LINKER:--no-as-needed # restore default behavior of including "unused" shared library dependencies on Ubuntu/Debian
```
mx990, line 26: Why did you remove this line? This will break shared builds on Ubuntu/Debian.

I believe the correct replacement for Apple is `-no_dead_strip_dylibs`.

core/arch/posix/src/posixsercommlayer.cpp
```diff
@@ -93,7 +102,9 @@ namespace forte::arch {
       struct termios stNewTIO;
       memset(&stNewTIO, 0, sizeof(stNewTIO));
 
+#ifdef HAVE_TERMIOS_C_LINE
       stNewTIO.c_line = mOldTIO.c_line;
+#endif
```
mx990: Instead of copying over members of the current termios struct into an empty one, I would suggest to only modify known members instead.

mx990 (changes requested):

core/arch/posix/src/posixsercommlayer.cpp
jwalt: I did not write this code and don't know about the motivation of doing it this way. This simply addresses the fact that the `c_line` field is a non-POSIX extension.

core/arch/posix/CMakeLists.txt
jwalt: GNU ld and LLVM lld linkers both have the equivalent of `--no-as-needed` as the standard behaviour. This line is only required on Debian-derivative distributions, since they elect to change that default for their builds of GCC. So this flag should be a special case for those particular linkers, no need to set up a fragile vocabulary of linker flags. I am open for suggestions how to detect those deviating linkers to make that flag conditional on their behaviour.

gravemalte: With `gcc -dumpspecs | grep "*no-as-needed"` we can find it.

core/arch/posix/src/posixsercommlayer.cpp
azoitl: As I was the one he did that, or at least one of the last ones to change into the current state I had a look at it again. I think the way I did it was a bit clumsy. I create PR #703 which not only removes the `c_line` setting but all in all is much cleaner now.

core/arch/posix/CMakeLists.txt
gravemalte: Or, the  `execute_process`[^1] in conjunction with `string`[^2] class should do the trick, instead on relying on grep.

[^1]: https://cmake.org/cmake/help/latest/command/execute_process.html
[^2]: https://cmake.org/cmake/help/latest/command/string.html

core/arch/posix/src/CMakeLists.txt
```diff
@@ -37,4 +36,21 @@ target_sources(forte-core PRIVATE
...
 )
 
+# missing on MacOS
+check_symbol_exists(pthread_condattr_setclock pthread.h have_pthread_condattr_setclock)
+if (have_pthread_condattr_setclock)
+    target_compile_definitions(forte-core PRIVATE HAVE_PTHREAD_CONDATTR_SETCLOCK)
+endif ()
+
+# missing on MacOS
+check_source_compiles(CXX [=[
```
azoitl: as `c_line` is not used anymore this should not be needed or?

core/arch/posix/CMakeLists.txt
jwalt: Well, I didn't think of the obvious way, a built-in cmake module for linker flag testing...

mx990 (approved):

azoitl (approved):

PR #703, opened 2025-10-30
Better use the existing termios settings

azoitl:
By copying the full existing termios settings several lines of asigments can be avoided. Furhtermore implementation specific settings are better hidden.

### DIFF
--- a/core/arch/posix/src/posixsercommlayer.cpp
+++ b/core/arch/posix/src/posixsercommlayer.cpp
@@ -1,5 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2017 ACIN, fortiss GmbH
+ * Copyright (c) 2012, 2025 ACIN, fortiss GmbH, Johannes Kepler University Linz
+ *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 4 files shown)


core/arch/posix/src/posixsercommlayer.cpp
```diff
@@ -117,9 +114,6 @@ namespace forte::arch {
         default: return com_infra::e_InitInvalidId; break;
       }
```
mx990: I would suggest to use `cfsetispeed()` and `cfsetospeed()` instead of directly manipulating the bits in `c_cflag`.

core/arch/posix/src/posixsercommlayer.cpp
mx990: I believe we now have to reset the bits beforehand:
```
stNewTIO.c_cflag &= ~CSIZE;
```

mx990 (changes requested):

jwalt (approved):
I didn't test this, but the changes should be POSIX-compliant, i.e. work on MacOS

core/arch/posix/src/posixsercommlayer.cpp
azoitl: Good point. Don't remember where I got the original code from. But now much cleaner.

core/arch/posix/src/posixsercommlayer.cpp
azoitl: To be nearer to the previous version I set the `c_cflag` to zero.

mx990 (approved):

PR #704, opened 2025-10-31
Upgraded XQuery to new namespaces

azoitl:


com/xquery/CMakeLists.txt
```diff
@@ -25,11 +25,9 @@ set(FORTE_BASEX_SRC_DIR
         CACHE PATH "Path to BaseX library directory"
 )
 
-message(FATAL_ERROR "This module has not yet been ported to 4diac FORTE 3.0")
-
 add_library(forte-com-xquery
-        $<$<BOOL:${FORTE_COM_XqueryClient}>:xqueryClientLayer.cpp>
-        $<$<BOOL:${FORTE_COM_XqueryClient}>:xqueryClientLayer.h>
+        $<$<BOOL:${FORTE_COM_XQUERYCLIENT}>:xqueryClientLayer.cpp>
+        $<$<BOOL:${FORTE_COM_XQUERYCLIENT}>:xqueryClientLayer.h>
```
mx990: I believe the condition is always true here?

com/xquery/CMakeLists.txt
mx990: I would suggest to check if the respective variables are non-empty first.

mx990: I believe we are missing an `if (WIN32)` here?

mx990 (changes requested):

com/xquery/CMakeLists.txt
azoitl: removed as it was anyhow duplicated

com/xquery/CMakeLists.txt
azoitl: good point

azoitl: ah yes. I missread the text. Fixed as well

com/xquery/CMakeLists.txt
```diff
@@ -44,14 +41,21 @@ add_library(forte-com-xquery
...
+if (OPENSSL_LIBRARY_DIR)
+  target_link_directories(forte-com-xquery PUBLIC ${OPENSSL_LIBRARY_DIR})
+endif ()
+
+if (WIN32)
+  target_link_libraries(forte-com-xquery PRIVATE libssl.a)
+  target_link_libraries(forte-com-xquery PRIVATE libcrypto.a)
+else ()
+  target_link_libraries(forte-com-xquery PRIVATE libssl.a -lpthread -ldl)
+  target_link_libraries(forte-com-xquery PRIVATE libcrypto.a -lpthread -ldl)
```
jwalt: These should not specify full library filenames. It is very probable that these libraries are present as dynamic libraries. Linking them statically when other parts may link them dynamically can lead to very bad things happening.

I also suggest to drop the transitive dependencies, since in the dynamic linking case (people using the system installed toolchain/libs) it is not needed. For static linking cases (like FBE), FORTE has established practice to leave this up to externally provided compiler/linker flags. An example of that is com/opcua, which also links to openssl.

Leaving all this out also removes the need for a system-specific case distinction, which is fragile anyways (e.g. `-ldl` would break FBE). A plain `target_link_libraries(forte-com-xquery PRIVATE ssl crypto)` should be sufficient.

Additionally, we could drop OPENSSL_INCLUDE_DIR and  OPENSSL_LIBRARY_DIR, since openssl is a widely-available library and in opc_ua we also do not use such an approach. Again, this can be specified through standard CMake compiler flags parameters if they are not on the default search path.

com/xquery/CMakeLists.txt
```diff
@@ -44,14 +41,21 @@ add_library(forte-com-xquery
...
-target_link_libraries(forte-com-xquery PRIVATE libssl.a)
-target_link_libraries(forte-com-xquery PRIVATE libcrypto.a)
-# else
-target_link_libraries(forte-com-xquery PRIVATE libssl.a -lpthread -ldl)
-target_link_libraries(forte-com-xquery PRIVATE libcrypto.a -lpthread -ldl)
-# forte_add_subdirectory(testFBs)
+if (OPENSSL_INCLUDE_DIR)
+  target_include_directories(forte-com-xquery PRIVATE ${OPENSSL_INCLUDE_DIR})
+endif ()
+if (FORTE_BASEX_SRC_DIR)
```
jwalt: Since `FORTE_BASEX_SRC_DIR` is unconditionally required, we need not guard against an empty value.

jwalt (changes requested):

com/xquery/CMakeLists.txt
azoitl: Thx for pointing this out. updated accordingly. This also has the advantage that no windows check is needed.

com/xquery/CMakeLists.txt
azoitl: Also changed

mx990 (approved):

jwalt (approved):

PR #709, opened 2025-11-02
Refactors toolchain configurations

franz-hoepfinger-4diac:
Consolidates ESP32 and ESP32-S3 toolchain configurations into reusable files.

This change removes duplicated toolchain files and introduces
common configurations for FreeRTOS, Logibus IO, Isobus,
Datapanel, Blinkmarine, Settings, NVS, CAN2 IO, Counter, and
Funk components, promoting maintainability and reducing redundancy.


franz-ms-muc:
Main Reason is to remove the old 
non 3.0.0 compatible Files. 
to not mis-guide new Users.

azoitl (changes requested):
I think all files that are either added or changed should be deleted. They contain configurations that are not part of 4diac IDE. Also the expect a certain file structure that is not documented (e.g., ../../../Application/lib/esp32s3).

azoitl:
Closed in favor of #714

PR #710, opened 2025-11-03
Upgraded GPIOCHIP module to namespaces and new FB layout

azoitl:
As part of this fix also the FB class name was aligned with the default generated one. This will help simplifying future upgrades.

### DIFF
--- a/modules/gpiochip/gpiochip_config_fb.cpp
+++ b/modules/gpiochip/gpiochip_config_fb.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 OFFIS e.V.
+ * Copyright (c) 2023, 2025 OFFIS e.V.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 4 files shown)


modules/gpiochip/gpiochip_config_fb.cpp
```diff
@@ -11,73 +11,172 @@
...
+    const auto cEventOutputNames = std::array{"INITO"_STRID, "IND"_STRID};
+    const auto cDataInputNames =
+        std::array{"QI"_STRID,       "VALUE"_STRID,    "ChipNumber"_STRID, "LineNumber"_STRID, "ReadWriteMode"_STRID,
+                   "BiasMode"_STRID, "ActiveLow"_STRID};
+    const auto cDataOutputNames = std::array{"QO"_STRID, "STATUS"_STRID};
+    const SFBInterfaceSpec cFBInterfaceSpec = {
+        .mEINames = cEventInputNames,
+        .mEITypeNames = {},
+        .mEONames = cEventOutputNames,
+        .mEOTypeNames = {},
```
mx990, line 31: Should `INIT` and `INITO` not have event type `EInit`?

mx990 (changes requested):

modules/gpiochip/gpiochip_config_fb.cpp
azoitl: I freshly exported it. I would keep that for now.

jwalt (approved):
builds fine in FBE-3.0-beta1

mx990 (approved):

PR #712, opened 2025-11-03
Improves error logging for external events

franz-hoepfinger-4diac:
Enhances error logging when external events are dropped due to a full queue. Now logs the Function Block name and the event name, aiding in debugging and identifying the source of dropped events.

### DIFF
--- a/core/src/ecet.cpp
+++ b/core/src/ecet.cpp
@@ -78,15 +78,18 @@ namespace forte {
 
   void CEventChainExecutionThread::startEventChain(TEventEntry paEventToAdd) {
     FORTE_TRACE("CEventChainExecutionThread::startEventChain\n");
+    bool inserted;
     {
       util::CCriticalRegion criticalRegion(mExternalEventListSync);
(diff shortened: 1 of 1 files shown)


core/src/ecet.cpp
```diff
@@ -84,7 +84,8 @@ namespace forte {
         mProcessingEvents = true;
         resumeSelfSuspend();
       } else {
-        DEVLOG_ERROR("External event queue is full, external event dropped!\n");
```
azoitl, line 87: as the new error logging is taking now more time we need to reconsider the locking strategy here. Although I admit it was already not good before. But I think it is better to change the lines 81 - 83 to:

```
{
  util::CCriticalRegion criticalRegion(mExternalEventListSync);
  bool inserted = mExternalEventList.push(paEventToAdd);
}
if (inserted) {
 ...
```

Or did I miss anything @mx990?

mx990: No, I believe that should cover everything.

mx990 (approved):

azoitl (approved):

PR #715, opened 2025-11-05
re-generate Timers

franz-hoepfinger-4diac:
re-generate Timers

they were not working anymore.

### DIFF
--- a/stdfblib/events/src/timers/E_TON_fbt.cpp
+++ b/stdfblib/events/src/timers/E_TON_fbt.cpp
@@ -6,13 +6,13 @@
  ***
  *** SPDX-License-Identifier: EPL-2.0
  ***
- *** This file was generated using the 4DIAC FORTE Export Filter V1.0.x NG!
+ *** This file was generated using the 4DIAC FORTE Export Filter 3.0.0.202511041205!
  ***
  *** Name: E_TON
  *** Description: standard timer function block (on-delay timing)
  *** Version:
- ***     1.0: 2024-03-04/Franz Hoepfinger - HR Agrartechnik GmbH -
  ***     3.0: 2025-04-14/Patrick Aigner -  - changed package
+ ***     1.0: 2024-03-04/Franz Hoepfinger - HR Agrartechnik GmbH -
  *************************************************************************/
 
 #include "forte/iec61499/events/timers/E_TON_fbt.h"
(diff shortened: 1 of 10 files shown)


azoitl (changes requested):
I don't see any change that would justify this PR. The only thing the new generation is doing is reordering elements.

stdfblib/events/src/timers/E_TON_fbt.cpp
```diff
@@ -104,7 +102,7 @@ namespace forte::iec61499::events::timers {
   void FORTE_E_TON::writeOutputData(const TEventID paEIID) {
     switch (paEIID) {
       case scmEventCNFID: {
-        writeData(cFBInterfaceSpec.getNumDIs() + 0, fb_E_RS->conn_Q.getValue(), conn_Q);
+        writeData(2, fb_E_RS->conn_Q.getValue(), conn_Q);
```
azoitl, line 105: based on the description that you gave in our matrix channel I would expect that this line is the problematic line. could you please revert the PR to just have the writeData lines changed. And test it again. This would make the PR less cluttered and better readable.

azoitl (changes requested):

stdfblib/events/src/timers/E_TON_fbt.cpp
franz-ms-muc: i will check later today. I still do not understand why ..

franz-hoepfinger-4diac: i re-replaced the "green" by the "red" line above, to check if this makes a difference. and the timer still works. 
i will now revert the whole commit, and check if this makes a difference !

stdfblib/events/src/timers/E_TON_fbt.cpp
```diff
@@ -51,8 +49,8 @@ namespace forte::iec61499::events::timers {
         {{}, "REQ"_STRID, "E_SWITCH"_STRID, "EI"_STRID},
         {"E_SWITCH"_STRID, "EO1"_STRID, "E_DELAY"_STRID, "START"_STRID},
         {"E_SWITCH"_STRID, "EO0"_STRID, "E_DELAY"_STRID, "STOP"_STRID},
-        {"E_SWITCH"_STRID, "EO1"_STRID, "E_RS"_STRID, "R"_STRID},
```
franz-hoepfinger-4diac, line 54: i guess it is this line. EO1vs EO0

PR #717, opened 2025-11-08
Adds handle removal functionality

franz-hoepfinger-4diac:
Adds the ability to remove IO handles from the controller.
This functionality ensures proper deregistration from the
IOMapper, preventing memory leaks and maintaining data consistency.

The `updateHandleList` function is also refactored to use
`std::find_if` and `erase` for improved readability and efficiency.

### DIFF
--- a/core/include/forte/io/device/io_controller.h
+++ b/core/include/forte/io/device/io_controller.h
@@ -104,6 +104,14 @@ namespace forte::io {
        */
       virtual void updateHandleList(std::string const &paId, IOHandle *paHandle);
 
+      /*! @brief Removes an IO handle from the controller and the IOMapper by ID
+       *
+       * The handle is removed from all internal lists.
+       *
+       * @param paId ID of the handle to remove
+       */
+      virtual void removeHandle(std::string const &paId);
+
     protected:
       explicit IODeviceController(CDeviceExecution &paDeviceExecution);
 
@@ -241,6 +249,8 @@ namespace forte::io {
       int mInitDelay;
 
       void addHandle(THandleList &paList, std::string const &paId, std::unique_ptr<IOHandle> paHandle);
+
+      void removeHandle(THandleList &paList, IOHandle *paRawHandle);
 
       // Functions needed for the external event handler interface
       void enableHandler() override {
--- a/core/src/io/device/io_controller.cpp
+++ b/core/src/io/device/io_controller.cpp
@@ -78,32 +78,61 @@ namespace forte::io {
     }
   }
 
-  void IODeviceController::updateHandleList(std::string const &paId, IOHandle *paHandle) {
-    for (auto it = mDiverseHandles.begin(); it != mDiverseHandles.end();) {
-      if (it->get() == paHandle) {
-        auto handleDirection = (*it)->getDirection();
-        switch (handleDirection) {
-          case IOMapper::In: {
-            util::CCriticalRegion criticalRegion(mHandleMutex);
-            mInputHandles.push_back(std::move(*it));
-            it = mDiverseHandles.erase(it);
-            break;
-          }
-          case IOMapper::Out: {
-            util::CCriticalRegion criticalRegion(mHandleMutex);
-            mOutputHandles.push_back(std::move(*it));
-            it = mDiverseHandles.erase(it);
-            break;
-          }
-          default:
-            DEVLOG_ERROR("[updateHandleList] %s's direction is neither `In` nor `Out`.\r\n", paId.c_str());
-            ++it;
-            break;
-        }
-        return;
-      } else {
-        ++it;
+  void IODeviceController::removeHandle(std::string const &paId) {
+    IOMapper &mapper = IOMapper::getInstance();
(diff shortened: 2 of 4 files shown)


franz-hoepfinger-4diac:
Tested, working.

core/include/forte/io/device/io_controller.h
```diff
@@ -104,6 +104,22 @@ namespace forte::io {
        */
       virtual void updateHandleList(std::string const &paId, IOHandle *paHandle);
 
+      /*! @brief Removes an IO handle from the controller and the IOMapper by pointer
+       *
+       * The handle is removed from all internal lists.
+       *
+       * @param paHandle IOHandle to remove
+       */
+      virtual void removeHandle(std::unique_ptr<IOHandle> paHandle);
```
azoitl: the version with the unique_ptr can not really work. Because the idea of a unique ptr is that only one entity holds it. So if you pass a unique_ptr to this method to compare if it is in a list this is contradicting.

azoitl (changes requested):

core/include/forte/io/device/io_controller.h
franz-hoepfinger-4diac: of course you are right. also this Method is nowhere called.

core/src/io/device/io_controller.cpp
```diff
@@ -78,32 +78,60 @@ namespace forte::io {
     }
   }
 
+  void IODeviceController::removeHandle(std::string const &paId) {
+    IOMapper &mapper = IOMapper::getInstance();
+    IOHandle *rawHandle = mapper.getHandle(paId);
+    switch (rawHandle->getDirection()) {
```
azoitl: you should check if the handle could be found

core/include/forte/io/device/io_controller.h
```diff
@@ -242,6 +250,10 @@ namespace forte::io {
 
       void addHandle(THandleList &paList, std::string const &paId, std::unique_ptr<IOHandle> paHandle);
 
+      void removeHandle(THandleList &paList, IOHandle *paRawHandle);
+
+      void removeHandle(THandleList &paList, std::string const &paId);
```
azoitl: This method is not implemented and AFAIK also not needed.

azoitl (changes requested):

core/src/io/device/io_controller.cpp
```diff
@@ -78,32 +78,62 @@ namespace forte::io {
...
+          mOutputHandles.push_back(std::move(*it));
+          mDiverseHandles.erase(it);
+          break;
         }
-        return;
-      } else {
-        ++it;
+        default: DEVLOG_ERROR("[updateHandleList] %s's direction is neither `In` nor `Out`.\r\n", paId.c_str()); break;
       }
+      return;
```
azoitl, line 119: you can save this return by adding either an else in line 137, or by chaning your if in line 119 to:

```
if (it == mDiverseHandles.end()) {
  DEVLOG_INFO("[updateHandleList] %s's is no member of mDiverseHandles.\r\n", paId.c_str());
  return;
}
```

which I like a bit more.

core/src/io/device/io_controller.cpp
```diff
@@ -78,32 +78,62 @@ namespace forte::io {
...
+    if (rawHandle) {
+      switch (rawHandle->getDirection()) {
+        case IOMapper::In: removeHandle(mInputHandles, rawHandle); return;
+        case IOMapper::Out: removeHandle(mOutputHandles, rawHandle); return;
+        case IOMapper::InOut: removeHandle(mDiverseHandles, rawHandle); return;
+        case IOMapper::UnknownDirection: removeHandle(mDiverseHandles, rawHandle); return;
+        default: break;
+      }
+    }
+    // else case
```
azoitl: This comment is a bit strange.

core/src/io/device/io_controller.cpp
```diff
@@ -78,32 +78,62 @@ namespace forte::io {
     }
   }
 
+  void IODeviceController::removeHandle(std::string const &paId) {
+    IOMapper &mapper = IOMapper::getInstance();
+    IOHandle *rawHandle = mapper.getHandle(paId);
+    if (rawHandle) {
```
azoitl: I like for these kind of code structure the C++17 init-statement. This allows you do define and check the variable in the if statement. Then the rawHandle would only be valid inside of the if:

`if(IOHandle * rawHandle = mapper.getHandle(paId); rawHandle) { `

core/src/io/mapper/io_mapper.cpp
```diff
@@ -86,6 +86,17 @@ namespace forte::io {
     }
   }
 
+  IOHandle *IOMapper::getHandle(std::string const &paId) {
+    util::CCriticalRegion criticalRegion(mSyncMutex);
+
+    auto handleIt = mHandles.find(paId);
```
azoitl: also here the C++17 init-statment would be nicer

azoitl (approved):

PR #732, opened 2025-12-03
TypeNames should not be inserted into the StringDict

azoitl:
TypeNames need to be present in 4diac FORTE's string dict therefore they shouldn't be inserted but only looked up.

### DIFF
--- a/stdfblib/system/src/OPCUA_MGR.cpp
+++ b/stdfblib/system/src/OPCUA_MGR.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022, 2024 Markus Meingast, Johannes Kepler University Linz
+ * Copyright (c) 2022, 2025 Markus Meingast, Johannes Kepler University Linz
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


stdfblib/system/src/OPCUA_MGR.cpp
```diff
@@ -1401,10 +1410,7 @@ namespace forte::iec61499::system {
     if (status != UA_STATUSCODE_GOOD) {
       return status;
     }
-    UA_String uaResp = UA_String_fromChars(uaMGR->mCommand.mMonitorResponse.c_str());
-    status = UA_Variant_setScalarCopy(output, &uaResp, &UA_TYPES[UA_TYPES_STRING]);
-    UA_String_clear(&uaResp);
-    return status;
+    return uaMGR->getCommandResult(output);
```
m-meingast: `getCommandResult()` writes `mAdditionalParams` to the output. For readWatches `mMonitorResponse` is necessary.

azoitl: thx for finding this. Missed that here a different response is generated. Changed it back to the original version and leave that part for a future clean-up.

m-meingast (approved):

PR #746, opened 2026-01-10
Adds Megalinter and config

MartinMelikMerkumians:
Megalinter is a curated suite of various linters to check every aspect of a git hosted repository. It is configurable in various aspects and helps to check on current best practices

### DIFF
--- a/.github/workflows/megalinter.yml
+++ b/.github/workflows/megalinter.yml
@@ -1,0 +1,69 @@
+
+---
+name: MegaLinter
+on:
+  pull_request:
+    branches: [ "release", "develop", "freeze", "feature/**", "3.0.x" ]
+  push:
+    branches: [ "release", "develop", "freeze", "feature/**", "3.0.x" ]
+
+permissions:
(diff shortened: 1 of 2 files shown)


MartinMelikMerkumians:
I have seen clang-tidy is not supported directly by megalinter. I need to write a descriptor for that what I can see from the manual

mx990:
We could also run `clang-tidy` via CMake using [`CMAKE_<LANG>_CLANG_TIDY`](https://cmake.org/cmake/help/v3.25/variable/CMAKE_LANG_CLANG_TIDY.html), which would also take care of providing the proper build commands for each file.

MartinMelikMerkumians:
yes, we could also do it that way. I would just like to have all in the same report (format). What I have read in the docs, it can be added via a plugin. If thats not working we can also still use the already existing clang-tidy workflow

MartinMelikMerkumians:
I will close this for now and do the development on my personal fork and resubmit as soon as I am ready. Thanks for the input

PR #751, opened 2026-01-16
Enable OPC UA on Linux in GitHub workflow

mx990:


mx990:
The official Ubuntu package of *open62541* comes with individual header files[^1], but FORTE expects a single `open62541.h`.

[^1]: https://packages.ubuntu.com/questing/amd64/libopen62541-1.4-dev/filelist

azoitl:
Moving to the individual files version is on my longerm whish/todo list. But this is a bit out in the future I fear.

PR #755, opened 2026-01-20
Fix Doxyfile paths for documentation generation

gravemalte:


Doxyfile
```diff
@@ -99,7 +99,7 @@ FULL_PATH_NAMES = YES
 # the path. The tag can be used to show relative paths in the file list. 
 # If left blank the directory from which doxygen is run is used as the 
 # path to strip.
-STRIP_FROM_PATH = ./src
+STRIP_FROM_PATH = ./core ./modules ./com
```
mx990: Does it actually make sense to strip these now?

Doxyfile
```diff
@@ -384,7 +384,7 @@ WARN_LOGFILE =
 # documented source files. You may enter file names like "myfile.cpp" or 
 # directories like "/usr/src/myproject". Separate the files or directories 
 # with spaces.
-INPUT = ./src
+INPUT = ./core ./modules ./com
```
mx990: I believe we also need `stdfblib`?

mx990 (changes requested):

Doxyfile
gravemalte: True

Doxyfile
gravemalte: My path looks stripped. But please try it.

gravemalte: Also the question is, if we want to strip them. When I think about it stripping maybe seem unnecessary

mx990: Yes, that is what I meant. I would suggest to remove it.

azoitl: Maybe as a side comment. I think I added the original strip because all code was in the src folder and this just added an indirection for all files and by stripping the output was cleaner. With the new structure this makes no sense anymore.

gravemalte: I will remove it.

mx990 (approved):

azoitl (approved):

PR #756, opened 2026-01-20
fix: FQN of FBs is now correct for the io layer

gravemalte:
Before that change the IO layer FBs were not found and this resulted in a 
`<Response ID="1" Reason="UNSUPPORTED_TYPE"/>` response.


azoitl (approved):
thx for finding and fixing this @gravemalte 

@franz-hoepfinger-4diac this PR may be important for you as well.

mx990 (approved):

PR #769, opened 2026-01-25
Fix missing includes for embrick module

mayrthom:
Adding` #include <cstring>` to the `modules/embrick/handler/pin.cpp`, so that the embrick module compiles without an error.


azoitl (approved):
Thx for resubmission on the right branch. 

For future references: You can also update your existing PR by locally rebasing and then force pushing. This feels a bit wrong but works for such cases very nicely.

PR #773, opened 2026-01-29
Initializes thread stack as nullptr

franz-hoepfinger-4diac:
in FreeRTOS, xTaskCreate needs just to know the SIZE of the Stack. making a "new" here is wasting Memory, as it is never used.


core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,7 +21,7 @@ namespace forte::arch {
   const int CFreeRTOSThread::scmForteTaskPriority = tskIDLE_PRIORITY + 4;
 
   CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize) {
-    mStack = new char[paStackSize];
+    mStack = nullptr;
```
azoitl: Thx for noticing this. This is really a waste of memory. Can you please remove the full line. `mStack` is already correctly initialized in the base class.

azoitl (changes requested):

azoitl (approved):

PR #774, opened 2026-01-30
 Adds data output type setup at GEN_ARRAY2ARRAY

franz-hoepfinger-4diac:


franz-hoepfinger-4diac:
@mx990 i want to learn: 

*********************
**learned!!!**
the trick is that "var_IN" here is re-used as a "var_OUT" not declaring it at all. 
Cool move ! 

*********************


1. is this `var_OUT.setup(arrayLength, valueTypeID);` needed here or not ? 

as in VALUES2ARRAY it is present. 

Thanks, 
Franz

PR #775, opened 2026-01-30
Sync COutDataConnection and COutInOutDataConnection

franz-hoepfinger-4diac:
Templated Type to return T instead of ANY at "getValue"

### DIFF
--- a/core/include/forte/inoutdataconn.h
+++ b/core/include/forte/inoutdataconn.h
@@ -68,9 +68,14 @@ namespace forte {
   template<typename T>
   class COutInOutDataConnection final : public CInOutDataConnection {
     public:
-      COutInOutDataConnection(CFunctionBlock &paSrcFB, TPortId paSrcPortId, const T &paValue) :
+      COutInOutDataConnection(CFunctionBlock &paSrcFB, const TPortId paSrcPortId, const T &paValue) :
           CInOutDataConnection(paSrcFB, paSrcPortId, &mValue),
           mValue(paValue) {
+      }
+
+
+      T &getValue2() {
+        return mValue;
       }
 
     private:


core/include/forte/inoutdataconn.h
```diff
@@ -68,11 +68,27 @@ namespace forte {
   template<typename T>
   class COutInOutDataConnection final : public CInOutDataConnection {
     public:
-      COutInOutDataConnection(CFunctionBlock &paSrcFB, TPortId paSrcPortId, const T &paValue) :
+      COutInOutDataConnection(CFunctionBlock &paSrcFB, const TPortId paSrcPortId, const T &paValue) :
```
MartinMelikMerkumians, line 71: I would suggest to not make paSrcPortId const, as it is passed by-value and const adds no value here.
To my knowledge most linters will even complain about this use of const because of the above mentioned reason

for reference https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

I saw that the const is also present in CoutDataConnection, and I would suggest to remove the const there too

core/include/forte/inoutdataconn.h
```diff
@@ -68,11 +68,27 @@ namespace forte {
...
           CInOutDataConnection(paSrcFB, paSrcPortId, &mValue),
           mValue(paValue) {
       }
 
+      void writeData(const T &paValue) {
+        mValue = paValue;
```
mx990: This does not work, unfortunately, because `COutInOutDataConnection::mValue` is only in use, when this is the "defining" InOut, that is, without a connection on the left.

Otherwise, `CInOutDataConnection::mValue` (note the different class) will point to the value of a different `COutInOutDataConnection` and the value needs to be written to that one. The type may not fully match either in such a case, for example, it may have an array with different bounds.

mx990 (changes requested):

MartinMelikMerkumians (changes requested):

core/include/forte/inoutdataconn.h
mx990: I agree regarding unnecessary `const` in declarations, but this is a definition, where is has some merit, same as in `COutDataConnection`.

mx990 (changes requested):
All of this is already handled correctly in `CInOutDataConnection`. Please leave it as is.

franz-hoepfinger-4diac:
> All of this is already handled correctly in `CInOutDataConnection`. Please leave it as is.

then this Line: 

`conn_inout_out_D1.getValue().setup(arrayLength, valueTypeID);`

does not Compile.

mx990:
Ah, now I understand what you are attempting to do. I have created a PR https://github.com/eclipse-4diac/4diac-forte/pull/778 that makes the *defining* value available, so you can setup the array with:
```cpp
conn_inout_out_D1.getDefiningValue().setup(arrayLength, valueTypeID);
```

franz-hoepfinger-4diac:
> ```c++
> getDefiningValue
> ```

coooooooooooool

PR #776, opened 2026-01-30
Enhances CIEC_ARRAY assignment with bounds check

franz-hoepfinger-4diac:
Adds a condition to set new bounds during assignment when
the array has variable bounds, ensuring boundary integrity
is maintained during dynamic array assignments.

### DIFF
--- a/core/include/forte/datatypes/forte_array.h
+++ b/core/include/forte/datatypes/forte_array.h
@@ -86,6 +86,9 @@ namespace forte {
       }
 
       CIEC_ARRAY &operator=(const CIEC_ARRAY &paSource) {
+        if (hasVariableBounds()) {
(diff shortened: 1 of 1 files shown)


core/include/forte/datatypes/forte_array.h
```diff
@@ -86,6 +86,9 @@ namespace forte {
       }
 
       CIEC_ARRAY &operator=(const CIEC_ARRAY &paSource) {
+        if (hasVariableBounds()) {
+          setBounds(paSource.getLowerBound(), paSource.getUpperBound());
+        }
```
mx990, line 91: The bounds of arrays are intentionally not modified on assignment, even if the destination has variable bounds. This is because variable-length arrays determine their bounds based on what is connected or passed via a parameter.

PR #778, opened 2026-01-30
Add getter for the defining value of an InOut connection

mx990:
Add getter for the defining value of an InOut connection and also rename `mValue` to `mDefiningValue` to avoid name clash.


franz-hoepfinger-4diac:
Support this. !

azoitl (approved):

PR #786, opened 2026-02-05
Adds constexpr to data type constructors and methods

franz-hoepfinger-4diac:
Enable constants "_gcf.cpp" to reside in Flash

### DIFF
--- a/core/include/forte/datatypes/forte_any.h
+++ b/core/include/forte/datatypes/forte_any.h
@@ -185,9 +185,7 @@ namespace forte {
        *  \return Returns TForteByte*
        */
 
-      virtual constexpr TLargestUIntValueType getLargestUInt() const {
-        return 0;
-      }
+      virtual constexpr TLargestUIntValueType getLargestUInt() const = 0;
       virtual constexpr void setLargestUInt(TLargestUIntValueType paVal) {
         (void) paVal;
       }
--- a/core/include/forte/datatypes/forte_any.h
+++ b/core/include/forte/datatypes/forte_any.h
@@ -1,8 +1,9 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2025 Profactor GmbH, ACIN, nxtcontrol GmbH, fortiss GmbH
+ * Copyright (c) 2005, 2026 Profactor GmbH, ACIN, nxtcontrol GmbH, fortiss GmbH
  *                          TU Vienna/ACIN, Martin Erich Jobst,
  *                          Martin Melik Merkumians,
  *                          Primetals Technologies Austria
+ *                          HR Agrartechnik GmbH
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -19,11 +20,13 @@
  *                - add support for data types with different size
  *   Martin Melik Merkumians - Add specialized numeric_limits for IEC types
  *   Alois Zoitl  - migrated data type toString to std::string
+ *   Franz Höpfinger - add constexpr
  *******************************************************************************/
 
 #pragma once
 
 #include <bit>
+#include <cstring>
 #include <span>
 #include <limits>
 #include <type_traits>
@@ -96,35 +99,31 @@ namespace forte {
         U oToCast;
         // If interacting with integers, add or remove sign extension
         if constexpr (std::is_base_of_v<CIEC_ANY_BIT, T> && std::is_base_of_v<CIEC_ANY_INT, U>) {
-          oToCast.setValueSimple(
-              U(static_cast<typename U::TValueType>(static_cast<typename T::TValueType>(paFromCast))));
+          oToCast = U(static_cast<typename U::TValueType>(static_cast<typename T::TValueType>(paFromCast)));
         } else if constexpr (std::is_base_of_v<CIEC_ANY_INT, T> && std::is_base_of_v<CIEC_ANY_BIT, U>) {
           typename T::TValueType fromValue = static_cast<typename T::TValueType>(paFromCast);
           typename std::make_unsigned_t<typename T::TValueType> fromValueUnsigned =
               static_cast<std::make_unsigned_t<typename T::TValueType>>(fromValue);
           typename U::TValueType toValue = static_cast<typename U::TValueType>(fromValueUnsigned);
-          oToCast.setValueSimple(U(toValue));
+          oToCast = U(toValue);
         } else if constexpr (std::is_base_of_v<CIEC_ANY_BIT, T> && // special cast binary to bool
                              std::is_base_of_v<CIEC_ANY_BIT, U>) {
           if constexpr (std::is_base_of_v<CIEC_BOOL, U>) { // reinterpret C/C++ bool to binary transfer
-            oToCast.setValueSimple(U(static_cast<typename T::TValueType>(paFromCast) % 2 == 1 ? true : false));
+            oToCast = U(static_cast<typename T::TValueType>(paFromCast) % 2 == 1 ? true : false);
           } else {
-            oToCast.setValueSimple(
-                U(static_cast<typename U::TValueType>(static_cast<typename T::TValueType>(paFromCast))));
+            oToCast = U(static_cast<typename U::TValueType>(static_cast<typename T::TValueType>(paFromCast)));
           }
         } else if constexpr (std::is_base_of_v<CIEC_ANY_REAL, T>) {
           specialCast(paFromCast, oToCast);
         } else if constexpr (std::is_base_of_v<CIEC_ANY_REAL, U>) {
           oToCast.setValue(paFromCast);
         } else {
-          oToCast.setValueSimple(paFromCast);
+          oToCast.setValue(paFromCast);
         }
         return oToCast;
       }
 
-      CIEC_ANY() {
-        setLargestUInt(0);
-      }
+      constexpr CIEC_ANY() = default;
 
       virtual ~CIEC_ANY() = default;
 
@@ -134,7 +133,7 @@ namespace forte {
        *
        */
       virtual void setValue(const CIEC_ANY &paValue) {
-        setValueSimple(paValue);
+        (void) paValue;
       }
 
       /*! \brief Reset the value for to the default initial value
@@ -186,17 +185,29 @@ namespace forte {
        *  \return Returns TForteByte*
        */
 
-      TForteByte *getDataPtr() {
-        return mAnyData.mData;
+      virtual constexpr TLargestUIntValueType getLargestUInt() const {
+        return 0;
+      }
+      virtual constexpr void setLargestUInt(TLargestUIntValueType paVal) {
+        (void) paVal;
+      }
+      virtual constexpr TLargestIntValueType getLargestInt() const {
+        return 0;
+      }
+      virtual constexpr void setLargestInt(TLargestIntValueType paVal) {
+        (void) paVal;
+      }
+      virtual TForteByte *getDataPtr() {
+        return nullptr;
       }
 
-      /*! \brief Get a const pointer to the union char array
+      /*! \brief Get a const pointer to the data array
        *
        *  \return Returns const TForteByte*
        */
 
-      const TForteByte *getConstDataPtr() const {
-        return mAnyData.mData;
+      virtual const TForteByte *getConstDataPtr() const {
+        return nullptr;
       }
 
       /*! \brief Converts string value to data type value
@@ -230,7 +241,10 @@ namespace forte {
        */
       [[nodiscard]] virtual bool equals(const CIEC_ANY &paOther) const {
         if (getDataTypeID() == paOther.getDataTypeID()) {
-          return mAnyData.mLargestUInt == paOther.mAnyData.mLargestUInt;
+          const size_t size = getIECMemorySize();
+          if (0 != size) {
+            return 0 == memcmp(getConstDataPtr(), paOther.getConstDataPtr(), size);
+          }
         }
         return false;
       }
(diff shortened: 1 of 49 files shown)


azoitl:
@franz-hoepfinger-4diac I just merged @mx990 CI fix into develop can you please rebase and force push

mx990:
I believe this PR should be split into two parts: adding `constexpr` to (most of) the data types and defining global constants in the header file. The latter will also require an exporter change in 4diac IDE.

As for the first part, I would really like to have `constexpr` for most of the data types at some point. However, we have to be careful, since the compiler does not check if something is a valid constant expression, unless it is actually required to be one. The following, for instance, does not compile[^1]:
```cpp
constinit CIEC_INT::TValueType seventeen = static_cast<CIEC_INT::TValueType>(17_INT);
```
We will have to rework parts of the data types before we can *safely* add `constexpr` to their constructors and methods. We should then also add tests to ensure they are valid constant expressions. This is something I would like to tackle at some point, but it will definitely take more time and effort.

As for the second part, there are two main points to consider when moving the definitions of constants into the header file:
1. A global constant may be initialized with any constant expression depending on both functions and other constants. This means all transitive dependencies would then also need to be included in the header file. If the constant should also be `constexpr`, then all of those dependencies would need to be `constexpr`, as well.
    
    In order to overcome this problem, we could only define global constants with *simple values*, that is, literals of an elementary type, in the header file. Any complex expressions would still be defined in the asscoiated `.cpp` file.
2. The value of a global constant may change in different versions of a library. If the constant is inlined into dependent code during compilation, its value may differ depending on which version the code was compiled against.

[^1]: with the error: `accessing ‘forte::CIEC_ANY::UAnyData::mInt16’ member instead of initialized ‘forte::CIEC_ANY::UAnyData::mLargestInt’ member in constant expression`

diplfranzhoepfinger:
I agree to all. 
Split is done. 

About complex constants. Agree to keep them in Cpp file. 
We should place them into bss-ram and set the Value with a init Method. This would then also save the very valuable IRAM.

diplfranzhoepfinger:
> 
> ```c++
> constinit CIEC_INT::TValueType seventeen = static_cast<CIEC_INT::TValueType>(17_INT);
> ```

to solve this Problem, there are 2 ways to do: 

1. the `setTINT16` should use `.mInt16` 
2. the `getTINT16` should always use the cast, removing the Access for `.mInt16`

both ways have pros-and-cons. 
in my first opionion: if somebody write this Line: 

> ```c++
> constinit CIEC_INT::TValueType seventeen = static_cast<CIEC_INT::TValueType>(17_INT);
> ```

he thinks all should be "INT" and writing a Constexpr with Type-Casts is something i would not consider ...

so that would opt for changing the "Setter" ... 

it would look this then: 

```c++
      constexpr void setTINT16(TForteInt16 src) {
        mAnyData.mInt16 = src;
      }
```

the above line then Compiles. 


@mx990 agree ? then i will incorporate this to all setter.

mx990:
Setting only the lower bits may leave unset or uninitialized bits when later accessing the union with a larger type. There are also other cases when different members of the union may be accessed, which would still pose a problem.

I believe the proper solution going forward is to remove the `UAnyData` union entirely, together with the getters and setters in `CIEC_ANY`. We would then have a data member `TValueType mData` in each non-generic data type instead. 

That should both simplify the code and prevent any problems due to type punning. It is something I wanted to do for quite some time but have not yet managed to find the time for.

diplfranzhoepfinger:
ah, yes. 
No Union, no Problems with type punning.

agree ...

diplfranzhoepfinger:
WOW ... it is green. 
now @mx990 what else Tests do we need to add ???

core/include/forte/datatypes/forte_any.h
```diff
@@ -186,17 +185,29 @@ namespace forte {
...
+        (void) paVal;
+      }
+      virtual constexpr TLargestIntValueType getLargestInt() const {
+        return 0;
+      }
+      virtual constexpr void setLargestInt(TLargestIntValueType paVal) {
+        (void) paVal;
+      }
+      virtual TForteByte *getDataPtr() {
+        return nullptr;
```
MartinMelikMerkumians: As CIEC_ANY wont hold any data anymore, does it still need to be instantiable?

If no, then all there virtual functions could be pure virtual, right? Or is there a need to provide a default behavior?

core/include/forte/datatypes/forte_any.h
```diff
@@ -186,17 +185,29 @@ namespace forte {
...
-      /*! \brief Get a const pointer to the union char array
+      /*! \brief Get a const pointer to the data array
        *
        *  \return Returns const TForteByte*
        */
 
-      const TForteByte *getConstDataPtr() const {
-        return mAnyData.mData;
+      virtual const TForteByte *getConstDataPtr() const {
+        return nullptr;
```
MartinMelikMerkumians, line 208: Same question as above

core/include/forte/datatypes/forte_any_elementary.h
```diff
@@ -37,11 +41,245 @@ namespace forte {
...
+          case e_TIME_OF_DAY:
+          case e_DATE_AND_TIME:
+          case e_LDATE:
+          case e_LTIME_OF_DAY:
+          case e_LDATE_AND_TIME: *reinterpret_cast<TForteUInt64 *>(data) = static_cast<TForteUInt64>(paVal); break;
+          case e_CHAR: *reinterpret_cast<TForteChar *>(data) = static_cast<TForteChar>(paVal); break;
+          case e_WCHAR: *reinterpret_cast<TForteWChar *>(data) = static_cast<TForteWChar>(paVal); break;
+          default: break;
+        }
+      }
```
MartinMelikMerkumians, line 197: similar question as in CIEC_ANY. Is this type planned to be instantiable? I thought that's what the _variant types are for and this would be only used as a type in references/pointers

core/include/forte/datatypes/forte_bool.h
```diff
@@ -39,32 +48,30 @@ namespace forte {
...
 
-      CIEC_BOOL() {
+      constexpr CIEC_BOOL() {
         setTBOOL8(false);
       }
 
-      CIEC_BOOL(const CIEC_BOOL &paValue) : CIEC_ANY_BIT() {
+      constexpr CIEC_BOOL(const CIEC_BOOL &paValue) : CIEC_ANY_BIT(), mData(paValue.mData) {
         *this = paValue;
       }
```
MartinMelikMerkumians, line 57: Why do you need to call this initializer and set *this = paValue?

If I understand this correctly you are setting mData now two times

core/include/forte/datatypes/forte_any.h
```diff
@@ -186,17 +185,29 @@ namespace forte {
...
 
-      /*! \brief Get a const pointer to the union char array
+      /*! \brief Get a const pointer to the data array
        *
        *  \return Returns const TForteByte*
        */
 
-      const TForteByte *getConstDataPtr() const {
-        return mAnyData.mData;
+      virtual const TForteByte *getConstDataPtr() const {
```
MartinMelikMerkumians, line 207: additional question, as I worked through half of the PR now, who is using the getDataPtr() and getConstDataPtr() now after your removal of the union data storage?

core/include/forte/datatypes/forte_byte.h
```diff
@@ -48,40 +56,38 @@ namespace forte {
...
       static constexpr TValueType scmMaxVal = std::numeric_limits<TValueType>::max();
 
-      CIEC_BYTE() = default;
+      constexpr CIEC_BYTE() = default;
 
-      CIEC_BYTE(const CIEC_BYTE &paValue) : CIEC_ANY_BIT() {
-        setValueSimple(paValue);
+      constexpr CIEC_BYTE(const CIEC_BYTE &paValue) : CIEC_ANY_BIT(), mData(paValue.mData) {
+        mData = static_cast<TValueType>(static_cast<CIEC_BYTE::TValueType>(paValue));
       }
```
MartinMelikMerkumians, line 63: similar question, you are settting mData via initializer, then again by line 62 by unboxing the value, getting a CIEC_BYTE::TValueType, then casting it again to CIEC_BYTE::TValueType and finally assigning it to mData a second time.

additional question, why do you not use the assignment operator?

core/include/forte/datatypes/forte_byte.h
```diff
@@ -48,40 +56,38 @@ namespace forte {
...
 
-      CIEC_BYTE(const CIEC_BOOL &paValue) : CIEC_ANY_BIT() {
-        setValueSimple(paValue);
+      constexpr CIEC_BYTE(const CIEC_BOOL &paValue) : CIEC_ANY_BIT() {
+        mData = static_cast<TValueType>(static_cast<CIEC_BOOL::TValueType>(paValue));
       }
```
MartinMelikMerkumians, line 67: why not use the assignment operator? also I think you do not need the second cast

core/include/forte/datatypes/forte_byte.h
```diff
@@ -48,40 +56,38 @@ namespace forte {
...
-        // Simple value assignment - no self assignment check needed
-        setValueSimple(paValue);
+        mData = paValue.mData;
         return *this;
       }
 
       CIEC_BYTE &operator=(const CIEC_BOOL &paValue) {
         // Simple value assignment - no self assignment check needed
-        setValueSimple(paValue);
+        mData = static_cast<TValueType>(static_cast<TValueType>(paValue));
```
MartinMelikMerkumians, line 81: I am surprised this is actually working, as you cast the BOOL to the elementary type of BYTE, which is not defined in CIEC_BOOL.

I would recommend to simply use
`mData = static_cast<CIEC_BOOL::TValueType>(paValue));`

someone correct me if I am wrong, this should suffice, as casting true to a number should be 1 and false should be 0

core/include/forte/datatypes/forte_char.h
```diff
@@ -33,30 +42,28 @@ namespace forte {
...
       static constexpr TValueType scmMaxVal = std::numeric_limits<TValueType>::max();
 
-      CIEC_CHAR() = default;
+      constexpr CIEC_CHAR() = default;
 
-      CIEC_CHAR(const CIEC_CHAR &paValue) : CIEC_ANY_CHAR() {
-        setValueSimple(paValue);
+      constexpr CIEC_CHAR(const CIEC_CHAR &paValue) : CIEC_ANY_CHAR(), mData(paValue.mData) {
+        mData = static_cast<TValueType>(static_cast<CIEC_CHAR::TValueType>(paValue));
       }
```
MartinMelikMerkumians, line 49: same question as at the previous data types

core/include/forte/datatypes/forte_date.h
```diff
@@ -29,32 +30,37 @@ namespace forte {
...
       constexpr static size_t scmBitLength = 64U;
 
-      CIEC_DATE() = default;
+      constexpr CIEC_DATE() = default;
 
-      CIEC_DATE(const CIEC_DATE &paValue) : CIEC_ANY_DATE() {
-        setValueSimple(paValue);
+      constexpr CIEC_DATE(const CIEC_DATE &paValue) : CIEC_ANY_DATE(), mData(paValue.mData) {
+        mData = static_cast<TValueType>(static_cast<CIEC_DATE::TValueType>(paValue));
       }
```
MartinMelikMerkumians, line 47: same question as at the previous data types

core/include/forte/datatypes/forte_date.h
```diff
@@ -29,32 +30,37 @@ namespace forte {
...
 
-      explicit CIEC_DATE(const TValueType paValue) {
-        setTUINT64(paValue);
+      constexpr explicit CIEC_DATE(const TValueType paValue) : mData(paValue) {
       }
```
MartinMelikMerkumians, line 50: while technically correct, I would still reuse the assignment operator

core/include/forte/datatypes/forte_date_and_time.h
```diff
@@ -29,32 +30,37 @@ namespace forte {
...
       constexpr static size_t scmBitLength = 64U;
 
-      CIEC_DATE_AND_TIME() = default;
+      constexpr CIEC_DATE_AND_TIME() = default;
 
-      CIEC_DATE_AND_TIME(const CIEC_DATE_AND_TIME &paValue) : CIEC_ANY_DATE() {
-        setValueSimple(paValue);
+      constexpr CIEC_DATE_AND_TIME(const CIEC_DATE_AND_TIME &paValue) : CIEC_ANY_DATE(), mData(paValue.mData) {
+        mData = static_cast<TValueType>(static_cast<CIEC_DATE_AND_TIME::TValueType>(paValue));
       }
```
MartinMelikMerkumians, line 47: same question as at the previous data types

core/include/forte/datatypes/forte_date_and_time.h
```diff
@@ -29,32 +30,37 @@ namespace forte {
...
         return *this;
       }
 
       /*! \brief Converts CIEC_TIME to unsigned 64 bit integer
        *
        *   Conversion operator for converting CIEC_TIME to unsigned 64 bit integer
        */
-      operator TForteUInt64() const {
+      constexpr operator TForteUInt64() const {
         return getTUINT64();
```
MartinMelikMerkumians, line 64: why not simply return mData as in the other datatypes?

core/include/forte/datatypes/forte_dint.h
```diff
@@ -45,48 +53,50 @@ namespace forte {
       [[deprecated("Please use the corresponding numeric_limits template")]]
       static constexpr TValueType scmMaxVal = std::numeric_limits<TValueType>::max();
 
-      CIEC_DINT() = default;
+      constexpr CIEC_DINT() = default;
 
-      CIEC_DINT(const CIEC_DINT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      constexpr CIEC_DINT(const CIEC_DINT &paValue) : CIEC_ANY_SIGNED(), mData(paValue.mData) {
```
MartinMelikMerkumians, line 58: I would suggest using the assignment operator

core/include/forte/datatypes/forte_dword.h
```diff
@@ -50,60 +58,60 @@ namespace forte {
       [[deprecated("Please use the corresponding numeric_limits template")]]
       static constexpr TValueType scmMaxVal = std::numeric_limits<TValueType>::max();
 
-      CIEC_DWORD() = default;
+      constexpr CIEC_DWORD() = default;
 
-      CIEC_DWORD(const CIEC_DWORD &paValue) : CIEC_ANY_BIT() {
-        setValueSimple(paValue);
+      constexpr CIEC_DWORD(const CIEC_DWORD &paValue) : CIEC_ANY_BIT(), mData(paValue.mData) {
```
MartinMelikMerkumians, line 63: as above, you are setting the value, then double casting and resetting mData.

I would recommend reusing the assignment op

core/include/forte/datatypes/forte_dword.h
```diff
@@ -50,60 +58,60 @@ namespace forte {
...
+        mData = static_cast<TValueType>(static_cast<CIEC_DWORD::TValueType>(paValue));
       }
 
-      CIEC_DWORD(const CIEC_WORD &paValue) : CIEC_ANY_BIT() {
-        setValueSimple(paValue);
+      constexpr CIEC_DWORD(const CIEC_WORD &paValue) : CIEC_ANY_BIT() {
+        mData = static_cast<TValueType>(static_cast<CIEC_WORD::TValueType>(paValue));
```
MartinMelikMerkumians, line 68: the second cast is not be necessary

core/include/forte/datatypes/forte_dword.h
```diff
@@ -50,60 +58,60 @@ namespace forte {
...
+      constexpr CIEC_DWORD(const CIEC_BYTE &paValue) :
+          CIEC_ANY_BIT(),
+          mData(static_cast<TValueType>(static_cast<TValueType>(paValue))) {
       }
 
-      CIEC_DWORD(const CIEC_BOOL &paValue) : CIEC_ANY_BIT() {
-        setValueSimple(paValue);
+      constexpr CIEC_DWORD(const CIEC_BOOL &paValue) :
+          CIEC_ANY_BIT(),
+          mData(static_cast<TValueType>(static_cast<TValueType>(paValue))) {
```
MartinMelikMerkumians, line 78: bth times you are double casting to CIEC_DWORD::TValueType, which is strange that this even works

core/include/forte/datatypes/forte_dword.h
```diff
@@ -50,60 +58,60 @@ namespace forte {
...
+        mData = static_cast<TValueType>(static_cast<TValueType>(paValue));
         return *this;
       }
 
       CIEC_DWORD &operator=(const CIEC_BOOL &paValue) {
         // Simple value assignment - no self assignment check needed
-        setValueSimple(paValue);
+        mData = static_cast<TValueType>(static_cast<TValueType>(paValue));
         return *this;
       }
```
MartinMelikMerkumians, line 107: same comment about the casts as above

core/include/forte/datatypes/forte_int.h
```diff
@@ -44,40 +52,40 @@ namespace forte {
...
-        setValueSimple(paValue);
+      constexpr CIEC_INT(const CIEC_INT &paValue) : CIEC_ANY_SIGNED(), mData(paValue.mData) {
       }
 
-      CIEC_INT(const CIEC_SINT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      constexpr CIEC_INT(const CIEC_SINT &paValue) :
+          CIEC_ANY_SIGNED(),
+          mData(static_cast<TValueType>(paValue.getSignedValue())) {
       }
```
MartinMelikMerkumians, line 63: why are you calling getSignedValue? the cast is not necessary

core/include/forte/datatypes/forte_int.h
```diff
@@ -44,40 +52,40 @@ namespace forte {
...
 
-      CIEC_INT(const CIEC_USINT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      constexpr CIEC_INT(const CIEC_USINT &paValue) :
+          CIEC_ANY_SIGNED(),
+          mData(static_cast<TValueType>(paValue.getSignedValue())) {
```
MartinMelikMerkumians, line 67: why are you needing getSignedValue? The cast should not be necessray, as a unsigned short should fit into an int

core/include/forte/datatypes/forte_ldate.h
```diff
@@ -30,28 +31,33 @@ namespace forte {
...
       constexpr static size_t scmBitLength = 64U;
 
-      CIEC_LDATE() = default;
+      constexpr CIEC_LDATE() = default;
 
-      CIEC_LDATE(const CIEC_LDATE &paValue) : CIEC_ANY_DATE() {
-        setValueSimple(paValue);
+      constexpr CIEC_LDATE(const CIEC_LDATE &paValue) : CIEC_ANY_DATE(), mData(paValue.mData) {
+        mData = static_cast<TValueType>(static_cast<CIEC_LDATE::TValueType>(paValue));
       }
```
MartinMelikMerkumians, line 48: I will now stop commenting on this, please check ALL datatype classes

core/include/forte/datatypes/forte_ldate.h
```diff
@@ -64,7 +70,7 @@ namespace forte {
        *
        *   Conversion operator for converting CIEC_TIME to unsigned 64 bit integer
        */
-      operator TForteUInt64() const {
+      constexpr operator TForteUInt64() const {
         return getTUINT64();
```
MartinMelikMerkumians, line 74: why not return mData?

core/include/forte/datatypes/forte_ldate_and_time.h
```diff
@@ -64,7 +70,7 @@ namespace forte {
        *
        *   Conversion operator for converting CIEC_TIME to unsigned 64 bit integer
        */
-      operator TForteUInt64() const {
+      constexpr operator TForteUInt64() const {
         return getTUINT64();
```
MartinMelikMerkumians, line 74: why not return mData?

core/include/forte/datatypes/forte_lint.h
```diff
@@ -46,56 +54,60 @@ namespace forte {
...
+      constexpr CIEC_LINT(const CIEC_USINT &paValue) :
+          CIEC_ANY_SIGNED(),
+          mData(static_cast<TValueType>(paValue.getSignedValue())) {
       }
 
-      explicit CIEC_LINT(const CIEC_ANY_INT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      constexpr explicit CIEC_LINT(const CIEC_ANY_INT &paValue) :
+          CIEC_ANY_SIGNED(),
+          mData(static_cast<TValueType>(paValue.getSignedValue())) {
```
MartinMelikMerkumians, line 94: please explain the need for using getSignedValue

core/include/forte/datatypes/forte_ltime_of_day.h
```diff
@@ -67,7 +77,7 @@ namespace forte {
        *
        *   Conversion operator for converting CIEC_SINT to elementary 32 bit integer
        */
-      operator TForteUInt64() const {
+      constexpr operator TForteUInt64() const {
         return getTUINT64();
```
MartinMelikMerkumians, line 81: why not return mData?

core/include/forte/datatypes/forte_string.h
```diff
@@ -219,30 +227,36 @@ namespace forte {
...
+              mData = '\0';
             }
           }
 
           virtual ~PARTIAL_ACCESS_TYPE() {
             if (mIndex > 0 && mIndex < mOriginalString.getMaximumLength()) {
-              const value_type value = operator value_type();
-              if (mOriginalString.length() < mIndex && value != '\0') {
+              const value_type value = *this;
+              if (mOriginalString.length() < mIndex && static_cast<TForteChar>(value) != '\0') {
```
MartinMelikMerkumians, line 254: please replace `static_cast<TForteChar>` with` static_cast<CIEC_CHAR::TValueType>`

core/include/forte/datatypes/forte_string.h
```diff
@@ -219,30 +227,36 @@ namespace forte {
       }
 
       class PARTIAL_ACCESS_TYPE : public CIEC_CHAR {
-          using value_type = TValueType;
-          using pointer_type = TValueType *;
```
MartinMelikMerkumians, line 223: Why did you remove this?

core/include/forte/datatypes/forte_string.h
```diff
@@ -252,10 +266,6 @@ namespace forte {
             return *this;
           }
 
-          operator value_type() const {
-            return getChar8();
-          }
-
```
MartinMelikMerkumians, line 258: Why did you remove this?

core/include/forte/datatypes/forte_time_of_day.h
```diff
@@ -31,36 +32,41 @@ namespace forte {
...
         return *this;
       }
 
       /*! \brief Converts CIEC_TIME_OF_DAY to elementary 64 bit unsigned integer
        *
        *   Conversion operator for converting CIEC_TIME_OF_DAY to elementary 64 bit unsigned integer
        */
-      explicit operator TForteUInt64() const {
+      constexpr explicit operator TForteUInt64() const {
         return getTUINT64();
```
MartinMelikMerkumians, line 70: why not return mData?

I also wont comment on this anymore, please check all datatypes yourself

tests/core/datatypes/constexpr_test.cpp
MartinMelikMerkumians, line 1: please add more tests, you are only testing a single point in a huge change of code

tests/core/datatypes/constexpr_test.cpp
```diff
@@ -0,0 +1,12 @@
+#include <boost/test/unit_test.hpp>
+#include "forte_boost_output_support.h"
+
+#include "forte/datatypes/forte_any_int.h"
+#include "forte/datatypes/forte_int.h"
+
+using namespace forte::literals;
+
+namespace forte::test {
+  // This should now compile because setTINT16 sets mInt16, which getTINT16 reads.
```
MartinMelikMerkumians, line 10: as far as I remember your changes, you are neither using setTINT16 nor getTINT16, so from my perspective this comments gives wrong information.

In general the comment is misleading as this is not the thing you want to test imho

MartinMelikMerkumians (changes requested):

core/include/forte/datatypes/forte_any.h
diplfranzhoepfinger: my thought YES ! i will try to make them PURE Virtual.

tests/core/datatypes/constexpr_test.cpp
diplfranzhoepfinger: you are right, i wrote this test BEFORE all the Changes. and in the Original Commit it was the case.

azoitl:
I only had a quick look on this PR but @MartinMelikMerkumians comments support the impression that I had. In its current form this PR is not acceptable.  It changes to much and to critical things at once. I would appreciate it if you could split this work on many smaller PRs. Each just addressing one small thing and each new one only submitted when the previous one is accepted. For example you could start with removing non needed API, or one data type after each other. This would give you the chance to learn the key points and we would have smaller junks to review.

diplfranzhoepfinger:
> I only had a quick look on this PR but @MartinMelikMerkumians comments support the impression that I had. In its current form this PR is not acceptable. It changes to much and to critical things at once. I would appreciate it if you could split this work on many smaller PRs. Each just addressing one small thing and each new one only submitted when the previous one is accepted. For example you could start with removing non needed API, or one data type after each other. This would give you the chance to learn the key points and we would have smaller junks to review.

I Agree. BUT: one thing will be a bigger PR: the Removal of the Union. I guess we should start with that one.

azoitl:
> I Agree. BUT: one thing will be a bigger PR: the Removal of the Union. I guess we should start with that one.

I disagree. Removing the union should be the last step. It is much saver and smaller to rework each class step by step.

PR #793, opened 2026-02-13
Restructure Datatypes PR 1 of N (deleted)

franz-hoepfinger-4diac:
This PR is the Start of a Series of small PRs, restructuring the Datatypes. 
https://github.com/eclipse-4diac/4diac-forte/pull/786 was a fail. 
so this is a 2nd try, refactoring small amounts of code one-by-one. 

related to https://github.com/eclipse-4diac/4diac-forte/issues/794

### DIFF
--- a/.github/workflows/cmake-multi-platform.yml
+++ b/.github/workflows/cmake-multi-platform.yml
@@ -59,7 +59,9 @@ jobs:
 
       - name: Install libboost (Linux)
         if: ${{ matrix.os != 'windows-latest' }}
-        run: sudo apt-get install -y libboost-all-dev
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libboost-all-dev
 
       - name: Install boost (Windows)
         if: ${{ matrix.os == 'windows-latest' }}


core/include/forte/datatypes/forte_int.h
```diff
@@ -46,38 +47,38 @@ namespace forte {
...
       CIEC_INT() = default;
 
-      CIEC_INT(const CIEC_INT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      CIEC_INT(const CIEC_INT &paValue) : CIEC_ANY_SIGNED(), mData(paValue.mData) {
       }
 
-      CIEC_INT(const CIEC_SINT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      CIEC_INT(const CIEC_SINT &paValue) : CIEC_ANY_SIGNED(), mData(static_cast<TValueType>(paValue.getSignedValue())) {
```
MartinMelikMerkumians: instead of calling getSignedValue and then downcasting it cia `static_cast `do

`static_cast<CIEC_SINT::TValueType>(paValue)` to extract the primitive type.

core/include/forte/datatypes/forte_int.h
```diff
@@ -46,38 +47,38 @@ namespace forte {
...
       }
 
-      CIEC_INT(const CIEC_USINT &paValue) : CIEC_ANY_SIGNED() {
-        setValueSimple(paValue);
+      CIEC_INT(const CIEC_USINT &paValue) :
+          CIEC_ANY_SIGNED(),
+          mData(static_cast<TValueType>(paValue.getSignedValue())) {
```
MartinMelikMerkumians: instead of calling getSignedValue and then downcasting it cia static_cast do

static_cast<CIEC_SINT::TValueType>(paValue) to extract the primitive type.

core/include/forte/datatypes/forte_int.h
```diff
@@ -46,38 +47,38 @@ namespace forte {
...
-        setTINT16(paValue);
+      explicit CIEC_INT(const TValueType paValue) : mData(paValue) {
       }
 
       ~CIEC_INT() override = default;
 
       CIEC_INT &operator=(const CIEC_INT &paValue) {
-        // Simple value assignment - no self assignment check needed
-        setValueSimple(paValue);
+        // this will be the new setter.
```
MartinMelikMerkumians: remove this comment, it has no added value

core/include/forte/datatypes/forte_int.h
```diff
@@ -46,38 +47,38 @@ namespace forte {
...
+        mData = paValue.mData;
         return *this;
       }
 
       template<typename T,
                std::enable_if_t<std::is_same_v<typename mpl::implicit_cast_t<T, CIEC_INT>, CIEC_INT>, int> = 0>
       CIEC_INT &operator=(const T &paValue) {
-        setValueSimple(paValue);
+        // this will be the new setter.
```
MartinMelikMerkumians: remove this comment, it has no added value

core/include/forte/datatypes/forte_int.h
```diff
@@ -46,38 +47,38 @@ namespace forte {
...
+        mData = static_cast<TValueType>(static_cast<typename T::TValueType>(paValue));
```
MartinMelikMerkumians: es the template ensures that the from value is implicitly castable to the target type simplify this to

`mData = static_cast<typename T::TValueType>(paValue));`
there should be no need for the second cast. if your compiler demands it, you are doing something wrong imho

core/include/forte/datatypes/forte_int.h
```diff
@@ -90,12 +91,23 @@ namespace forte {
...
         return e_INT;
       }
+
+      TForteByte *getDataPtr() override {
+        return reinterpret_cast<TForteByte *>(&mData);
+      }
+
+      const TForteByte *getConstDataPtr() const override {
+        return reinterpret_cast<const TForteByte *>(&mData);
+      }
```
MartinMelikMerkumians: I would suggest you remove the need for these functions first, before changing the data types at all.

https://github.com/search?q=repo%3Aeclipse-4diac%2F4diac-forte+%28getDataPtr+OR+getConstDataPtr%29&type=code

there are only 8 occurrences according to the search. This would eliminate the need for this entirely.
@azoitl what it your opinion?

with the union as goal implementation this was a performance gain, but with the target that each type has its own memory we are getting an expensive virtual function call now

MartinMelikMerkumians (changes requested):

core/include/forte/datatypes/forte_int.h
azoitl: I agree @MartinMelikMerkumians.  I checked and `getDataPtr()` is not used at all anymore. So this could simply be removed. 

`getConstDataPtr() ` is a bit a different story this is currently used by FBDASN1 layer and the CAN layer. Not sure how to replace that nicely. But yes this should definitely be removed first.

franz-hoepfinger-4diac:
Sorry, was closed accidentaly. i will start a new PR .

core/include/forte/datatypes/forte_int.h
franz-hoepfinger-4diac: > I would suggest you remove the need for these functions first, before changing the data types at all.
> 


any YES, i want to do things first before change anything. your are absolutely right !!!!!

Issue #794, opened 2026-02-13
Restructure Datatypes


franz-hoepfinger-4diac:
This Issue collects all PRs related with Restructure Datatypes

azoitl:
I guess this is the issue to track how we can remove the union from CIEC_ANY.  After reviewing the first PR and checking some of the current implementation aspects, triggered by @MartinMelikMerkumians question in this PR may main concern is how do we efficiently implement the `setValue()` method for all `ANY_ELEMENTARY` data types. As we currently have the union this can be done very efficiently and simple. Without we need some more generic way. I currently have not a good idea how to do this. Curios what others think.

MartinMelikMerkumians:
As there seems to be no good plan on proceeding with this, and due to the implications of critical path performance and no compelling reason to even do this, I close this issue. If there is a clear use case for this major rewrite, please reopen and provide a clear design and implementation path.

PR #797, opened 2026-02-14
Restructure Datatypes PR 1 of N

franz-hoepfinger-4diac:
This PR is the Start of a Series of small PRs, restructuring the Datatypes.
#786 was a fail.
so this is a 2nd try, refactoring small amounts of code one-by-one.

this PR now does introduce the "setValue" to all Types, and avoid the "setValueSimple". 

related to #794

### DIFF
--- a/core/include/forte/datatypes/forte_any.h
+++ b/core/include/forte/datatypes/forte_any.h
@@ -133,8 +133,7 @@ namespace forte {
        *  The data type value is set through the copy assignment
        *
        */
-      virtual void setValue(const CIEC_ANY &) {
-      }
+      virtual void setValue(const CIEC_ANY &) = 0;
 
       /*! \brief Reset the value for to the default initial value
        */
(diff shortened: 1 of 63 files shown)


core/include/forte/datatypes/forte_any.h
```diff
@@ -133,8 +133,7 @@ namespace forte {
        *  The data type value is set through the copy assignment
        *
        */
-      virtual void setValue(const CIEC_ANY &paValue) {
-        setValueSimple(paValue);
+      virtual void setValue(const CIEC_ANY &) {
```
diplfranzhoepfinger: Can be pure Virtual

franz-hoepfinger-4diac: done

azoitl:
@franz-hoepfinger-4diac what you are proposing here is a performance and maintenance nightmare. The `setValue()` method is the second most performance critical function in 4diac FORTE. I don't even want to measure how much slower your PR is compared to what we have now.  I think before you move forward you need to develop a solid concept and architecture how the new data type storage should look like and how we can address the very critical performance issue while at the same time also allow the implicit up-casting feature.

diplfranzhoepfinger:
@azoitl understand. 
Not soooo easy then. 
I will go deeper and think how this can work.

PR #798, opened 2026-02-14
use polymorphism. TRY.

franz-hoepfinger-4diac:
no real PR, just a Discussion base.


azoitl:
Unfortunately this is not the way to go.

PR #804, opened 2026-02-24
Refactor A&C Layer for Service Interface FB usage

m-meingast:


com/opc_ua/src/opcua_ac_layer.cpp
```diff
@@ -125,6 +116,7 @@ namespace forte::com_infra::opc_ua {
   }
 
   EComResponse COPC_UA_AC_Layer::recvData(const void *paData, unsigned int) {
+    DEVLOG_ERROR("RecvData\n");
```
mx990: It looks like this is for debugging purposes. Should this really be an error?

com/opc_ua/src/opcua_ac_layer.cpp
```diff
@@ -147,14 +142,16 @@ namespace forte::com_infra::opc_ua {
...
+    if (triggerAlarm(server, activate) != UA_STATUSCODE_GOOD) {
       DEVLOG_ERROR("[OPC UA A&C LAYER]: Sending Alarm Data failed for FB %s!\n", getCommFB()->getInstanceName());
       return e_ProcessDataSendFailed;
     }
+    readAlarmStateValues(server);
     return e_ProcessDataOk;
   }
 
   EComResponse COPC_UA_AC_Layer::processInterrupt() {
+    DEVLOG_ERROR("ProcessInterrupt\n");
```
mx990: It looks like this is for debugging purposes. Should this really be an error?

mx990 (changes requested):

com/opc_ua/src/opcua_ac_layer.cpp
m-meingast: Thank you for noticing that. I wanted to see the execution order and forgot to delete this.

com/opc_ua/src/opcua_ac_layer.cpp
m-meingast: Same thing here.

mx990 (approved):

azoitl (approved):

PR #807, opened 2026-02-27
Runtime now accepts inline vector literals for ANY_VARIANT

MehranKianiHub:
Runtime now accepts inline vector literals for ANY_VARIANT, so real-vector feeding is user-friendly for M_Inference/ML_Inference

### DIFF
--- a/core/src/datatypes/forte_any_variant.cpp
+++ b/core/src/datatypes/forte_any_variant.cpp
@@ -13,8 +13,109 @@
  *   Alois Zoitl  - migrated data type toString to std::string
  *******************************************************************************/
 #include "forte/datatypes/forte_any_variant.h"
+#include "forte/datatypes/forte_array_dynamic.h"
+
+#include <charconv>
+#include <cctype>
+#include <cstring>
+#include <memory>
+#include <string_view>
 
 using namespace forte::literals;
+
+namespace {
+  const char *skipWhitespace(const char *paPos, const char *paEnd) {
+    while (paPos < paEnd && std::isspace(static_cast<unsigned char>(*paPos))) {
+      ++paPos;
+    }
+    return paPos;
+  }
+
+  bool parseKeyword(const char *&paPos, const char *paEnd, std::string_view paKeyword) {
+    paPos = skipWhitespace(paPos, paEnd);
+    for (char keywordChar : paKeyword) {
+      if (paPos == paEnd || std::toupper(static_cast<unsigned char>(*paPos)) != keywordChar) {
+        return false;
+      }
+      ++paPos;
+    }
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


azoitl:
@MehranKianiHub thx for your contribution. In order that we can accept your PR we have to follow a few things that is defined in the Eclipse Development process. This process protects you and us.

For that you need to create an Eclipse account, link your Eclipse account with your github account and sign the Eclipse Contributor Agreement. With this you say that you developed the provided code, that you have the permission to give us the code and that we are allowed to use your code under the Eclipse Public License V2.

All this is documentation in our contributing guidelines: https://eclipse.dev/4diac/doc/development/contribute.html
I'm currently reworking that document to make it a bit more approachable. However it is not in state to share yet.

azoitl:
@MehranKianiHub I finally found time to do a new contributing guide: https://github.com/eclipse-4diac/4diac-documentation/pull/93

azoitl:
@MehranKianiHub we discussed your approach among the committers. We see the need for having the option of configuring array pins. However we see also the problem that your proposed syntax is not supported by IEC 61131-3. Therefore what would we propose is to download in such cases an array type using the device management interface and use the array type when configuring pins. The syntax for downloading array types would be defined by the [IEC 61499 compliance profile for feasibility demonstrations](https://holobloc.com/doc/ita/index.htm). 

The reason why we think that approach is better is that this would be the starting point for deploying data types to an existing run-time with out the need for recompiling 4diac FORTE. And secondly it would be better be aligned with IEC 61131-3. 

Would you be interesting in supporting us in that direction? We can also setup a meeting to discuss the details if you want.

MehranKianiHub:
> @MehranKianiHub we discussed your approach among the committers. We see the need for having the option of configuring array pins. However we see also the problem that your proposed syntax is not supported by IEC 61131-3. Therefore what would we propose is to download in such cases an array type using the device management interface and use the array type when configuring pins. The syntax for downloading array types would be defined by the [IEC 61499 compliance profile for feasibility demonstrations](https://holobloc.com/doc/ita/index.htm).
> 
> The reason why we think that approach is better is that this would be the starting point for deploying data types to an existing run-time with out the need for recompiling 4diac FORTE. And secondly it would be better be aligned with IEC 61131-3.
> 
> Would you be interesting in supporting us in that direction? We can also setup a meeting to discuss the details if you want.

@azoitl Sure, I'll start working on your suggestion shortly and will come back to you to set up a meeting if required

franz-hoepfinger-4diac:
just a humble question: 
the new "provide Array" Blocks do not help here ? 

you can define an Array, and forward it to ANY_VARIANT; however, it will always be `[0..x]` based index.

MartinMelikMerkumians (commented):
Agreeing with @azoitl as this is not a 61131/61499 conformant literal or expression, we should analyze the use case and come up with a conformant solution

PR #824, opened 2026-03-24
Reorganize includes in OPC UA Comlayer for non-amalgamation builds

m-meingast:


com/opc_ua/include/forte/com/opc_ua/opcua_nodesets.h
```diff
@@ -15,7 +15,10 @@
 
 #include "forte/util/hook.h"
 
-#include <open62541.h>
+#include <open62541/common.h>
+#include <open62541/types.h>
+
+struct UA_Server;
```
mx990, line 21: These forward declarations should not be needed, since `UA_Server`, `UA_ServerConfig`, and `UA_Client` are all declared in `open62541/common.h`.

com/opc_ua/src/detail/lds_me_handler.h
```diff
@@ -14,8 +14,6 @@
 
 #pragma once
 
-#include <open62541.h>
-
```
mx990, line 18: I believe this also needs to include both common and types:
```cpp
#include <open62541/common.h>
#include <open62541/types.h>
```

mx990 (changes requested):

com/opc_ua/include/forte/com/opc_ua/opcua_nodesets.h
m-meingast: Removing this forward declaration leads to the compile error `‘UA_Server’ was not declared in this scope`. I looked it up, `UA_Server` is defined in `ua_server_internal.h`. To avoid this declaration `open62541/server.h` could be included.

mx990: I re-checked and it looks like the forward declarations in `open62541/common.h` were removed for 1.5.0, so never mind.

com/opc_ua/src/detail/lds_me_handler.h
mx990: Since the forward declarations were removed in open62541 version 1.5.0, I believe we also need forward declarations here for `UA_Server` and so on.

mx990 (approved):

mx990 (approved):

azoitl (approved):

PR #825, opened 2026-03-27
Add ANY_DERIVED_VARIANT and ANY_STRUCT data types

mx990:


MartinMelikMerkumians (commented):
I think only ANY_DERIVED as STRUCTs are currently tests, right?
Wouldn't it be good to also include some basic ANY_DERIVED as ARRAY tests as well?

mx990:
Good point. I have now also added tests with arrays, except in `String_Conversion_test`, since we have no named array types, yet.

MartinMelikMerkumians (approved):

azoitl (approved):

PR #828, opened 2026-04-02
Change A&C layer to use full qualified name

m-meingast:
Log error outputs and OPC UA instance object creation use full qualified name

### DIFF
--- a/com/opc_ua/src/opcua_ac_layer.cpp
+++ b/com/opc_ua/src/opcua_ac_layer.cpp
@@ -32,6 +32,7 @@ namespace forte::com_infra::opc_ua {
     const std::string scmAlarmConditionName = "AlarmCondition";
 
     char smEmptyString[] = "";
+    char smSeparator = '/';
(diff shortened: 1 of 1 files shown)


com/opc_ua/src/opcua_ac_layer.cpp
```diff
@@ -278,7 +285,7 @@ namespace forte::com_infra::opc_ua {
       DEVLOG_ERROR("[OPC UA A&C LAYER]: Browsepath is invalid!");
       return UA_STATUSCODE_BAD;
     }
-    std::string instanceNameStr{getCommFB()->getInstanceName()};
+    std::string instanceNameStr{getCommFB()->getFullQualifiedApplicationInstanceName(smSeparator).c_str()};
```
azoitl: as you are creating a `std::string` you should remove the `.c_str()` at the end.

azoitl (changes requested):

azoitl (approved):

PR #829, opened 2026-04-04
Use `std::string_view` for MGM command response texts

franz-hoepfinger-4diac:
Optimize memory usage and avoid unnecessary string copies by employing `std::string_view` for the static response text array and its accessor function.

### DIFF
--- a/core/src/datatypes/forte_string.cpp
+++ b/core/src/datatypes/forte_string.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2025 Profactor GmbH, TU Wien/ACIN, nxtControl GmbH,
+ * Copyright (c) 2005       Profactor GmbH, TU Wien/ACIN, nxtControl GmbH,
  *                          fortiss GmbH, Primetals Technologies Austria GmbH
- *                          Martin Erich Jobst
+ *                          Martin Erich Jobst, HR Agrartechnik GmbH
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -18,6 +18,7 @@
  *                           - changes storage to std::string
  *   Alois Zoitl - migrated data type toString to std::string
  *   Martin Jobst - fix line feed and newline escape sequences
+ *    Franz Höpfinger - string_view for getResponseText
  *******************************************************************************/
 
 #include "forte/datatypes/forte_string.h"
@@ -139,7 +140,7 @@ namespace forte {
     this->append(paValue.getStorage());
   }
 
-  void CIEC_STRING::append(const std::string &paValue) {
+  void CIEC_STRING::append(const std::string_view paValue) {
     mValue.append(paValue);
   }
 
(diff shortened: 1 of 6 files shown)


core/src/mgmcmd.cpp
```diff
@@ -21,21 +22,21 @@ namespace forte::mgm_cmd {
...
-                                             "NOT_READY",
-                                             "UNSUPPORTED_CMD",
-                                             "UNSUPPORTED_TYPE",
-                                             "NO_SUCH_OBJECT",
-                                             "INVALID_OBJECT",
-                                             "INVALID_OPERATION",
-                                             "INVALID_STATE",
-                                             "OVERFLOW",
-                                             "INVALID_DST"};
+  constexpr std::string_view scmMGMResponseTexts[] = {"RDY",
```
diplfranzhoepfinger: with constexpr this Array is no longer in RAM, but in Flash, saving valuable RAM Space.

core/src/datatypes/forte_string.cpp
```diff
@@ -143,6 +144,10 @@ namespace forte {
     mValue.append(paValue);
   }
 
+  void CIEC_STRING::append(std::string_view paValue) {
```
azoitl: It is not necessary to duplicate this method. A `std::string` can automatically be converted to a `std:string_view` so please change the original one.

azoitl (changes requested):
Thx for improving our memory usage. 

Please Fix the commit message to clearly describe what this commit is doing. The code before also didn't do string copies as the getter returned a reference to a string and therefore no copies.

As described in our contributing guides make the commit message self contained: https://eclipse.dev/4diac/doc/development/contribute.html#_3_commit_messages

That also means not outlook or things that are not part of this commit. For that open either an issue or start a discussion. 

As you touched the response array the formatting should not be a separate commit. Again please refer to our contributing guide on commit messages and commit structure.

core/src/datatypes/forte_string.cpp
mx990: Please also change the parameter to a `const std::string_view`.

core/src/mgmcmd.cpp
mx990: Please also use the literal operator to directly create the string views[^1].

[^1]: see also https://en.cppreference.com/w/cpp/string/basic_string_view/operator%2522%2522sv.html

mx990 (changes requested):

azoitl (changes requested):
The commit message is still not reflecting the core points for this change.

Please note as per recent updates on the Eclipse Project handbook the second year is not needed anymore. Please remove everywhere (especially where you have not changed it. All copyright holders should be separated by commas.

mx990 (changes requested):
Please also remove those `.` lines in your commit message. Those unnecessarily bloat the commit log.

diplfranzhoepfinger:
> Please note as per recent updates on the Eclipse Project handbook the second year is not needed anymore. Please remove everywhere (especially where you have not changed it. All copyright holders should be separated by commas.

https://www.eclipse.org/projects/handbook/#ip-copyright-headers
i see. Only the creation year, nothing else.

azoitl (approved):

mx990 (approved):

PR #830, opened 2026-04-05
Use SO_EXCLUSIVEADDRUSE for Windows sockets

franz-hoepfinger-4diac:
Replacing SO_REUSEADDR with SO_EXCLUSIVEADDRUSE prevents multiple FORTE instances from binding to the same port simultaneously on Windows.

### DIFF
--- a/core/arch/win32/src/win32socketinterf.cpp
+++ b/core/arch/win32/src/win32socketinterf.cpp
@@ -1,6 +1,6 @@
 /*******************************************************************************
  * Copyright (c) 2010 ACIN, Profactor GmbH, AIT, fortiss GmbH, OFFIS e.V., HIT robot group
- *               Samator Indo Gas, Primetals Technologies Austria GmbH
+ *               Samator Indo Gas, Primetals Technologies Austria GmbH, HR Agrartechnik GmbH
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -14,6 +14,7 @@
  *  Zhao Xin - fix socket resource leakage
  *  Ketut Kumajaya - switch to the Unicode version of WSAStringToAddress
  *  Markus Meingast - add logging of actual tcp port being used
+ *  Franz Höpfinger - set SO_EXCLUSIVEADDRUSE to avoid port hijacking
  *******************************************************************************/
 
 #include "forte/arch/sockhand.h" //needs to be first pulls in the platform specific includes
@@ -71,9 +72,9 @@ namespace forte::arch {
     stSockAddr.sin_addr.s_addr = htonl(INADDR_ANY);
 
     int nOptVal = 1;
-    if (setsockopt(nSocket, SOL_SOCKET, SO_REUSEADDR, (char *) &nOptVal, sizeof(nOptVal)) == SOCKET_ERROR) {
+    if (setsockopt(nSocket, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, (char *) &nOptVal, sizeof(nOptVal)) == SOCKET_ERROR) {
       LPSTR pacErrorMessage = getErrorMessage(WSAGetLastError());
-      DEVLOG_ERROR("CWin32SocketInterface: could not set socket option SO_REUSEADDR:  %s\n", pacErrorMessage);
+      DEVLOG_ERROR("CWin32SocketInterface: could not set socket option SO_EXCLUSIVEADDRUSE:  %s\n", pacErrorMessage);
       LocalFree(pacErrorMessage);
       return nRetVal;
     }
@@ -241,7 +242,7 @@ namespace forte::arch {
 
     if (INVALID_SOCKET != nSocket) {
       int nReuseAddrVal = 1;
-      if (0 <= setsockopt(nSocket, SOL_SOCKET, SO_REUSEADDR, (char *) &nReuseAddrVal, sizeof(nReuseAddrVal))) {
+      if (0 <= setsockopt(nSocket, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, (char *) &nReuseAddrVal, sizeof(nReuseAddrVal))) {
         struct sockaddr_in stSockAddr = {};
         stSockAddr.sin_family = AF_INET;
         stSockAddr.sin_port = htons(paPort);
@@ -281,7 +282,8 @@ namespace forte::arch {
       } else {
         int nLastError = WSAGetLastError();
         LPSTR pacErrorMessage = getErrorMessage(nLastError);
-        DEVLOG_ERROR("CWin32SocketInterface: setsockopt(SO_REUSEADDR) failed: %d - %s\n", nLastError, pacErrorMessage);
+        DEVLOG_ERROR("CWin32SocketInterface: setsockopt(SO_EXCLUSIVEADDRUSE) failed: %d - %s\n", nLastError,
+                     pacErrorMessage);
         LocalFree(pacErrorMessage);
       }
     } else {


azoitl:
It would be interesting why we had `SO_REUSEADDR` in the beginning. For TCP listening sockets it makes little sense. I have something in the back of my mind that we introduced it so that when testing in windows and we killed 4diac FORTE from the debugger that we could quicker restart 4diac FORTE. But I'm not sure
@MartinMelikMerkumians do you remember? 

But maybe we should remove it as a whole from both the bsd and windows socket implementation.

diplfranzhoepfinger:
SO_REUSEADDR in LINUX is same as SO_EXCLUSIVEADDRUSE in Windows (beginning from a certain Version: Windows NT 4 SP4). So with this PR Windows and Linux will behave same then. this is my intention. SO_EXCLUSIVEADDRUSE does not exist at all in LINUX.

azoitl:
> SO_REUSEADDR in LINUX is same as SO_EXCLUSIVEADDRUSE in Windows (beginning from a certain Version: Windows NT 4 SP4). So with this PR Windows and Linux will behave same then. this is my intention. SO_EXCLUSIVEADDRUSE does not exist at all in LINUX.

I understood this. But if the Linux behavior is already wrong your proposed Windows behavior will be equally wrong. So your PR is a nice chance to discuss this.

MartinMelikMerkumians:
Hi. Reading the MSDN I think the explanation that SO_EXCLUSIVEADDRUSE on Windows behaves the same as SO_REUSEADDR on Linux is wrong.

On the contrary, it hardens port ownership where even a second bind with SO_REUSEADDR won't give you access to that interface/port combination, which equals Linux standard behavior with no option set.

Therefore, the commit is correct from my point of view, the explanation given isn't.

@azoitl yeah, I am quite sure that was a hack to get port access after a forte crash earlier, as on Linux it allows to bind on TIME_WAIT.

I would opt to remove the SO_REUSEADDR on BSD and Linux platforms, as you said, I doesn't make sense for a TCP port.

franz-hoepfinger-4diac:
> On the contrary, it hardens port ownership where even a second bind with SO_REUSEADDR won't give you access to that interface/port combination, which equals Linux standard behavior with no option set.


my main topic was, to have a console output: "hey user: you are starting 4diac the 2nd Time on this Machine". This was there on BSD, but not on Windows.

franz-hoepfinger-4diac:
> I understood this. But if the Linux behavior is already wrong your proposed Windows behavior will be equally wrong. So your PR is a nice chance to discuss this.

understand. so i could live with: allow re-use of the Port, and find another Implementation to give the User Information about the re-used Port !

azoitl:
@MartinMelikMerkumians then I'll remove the `SO_REUSEADDR` on BSD and Linux platforms. What should we do on Windows then? 

I've googled a bit and people where removing the `SO_EXCLUSIVEADDRUSE`. What I read is that the reason this flag was added is not really existing anymore. But I'm not as good in Windows API as you are.

azoitl:
Clean-up of BSD socket impl is done in PR #834.

MartinMelikMerkumians:
> @MartinMelikMerkumians then I'll remove the `SO_REUSEADDR` on BSD and Linux platforms. What should we do on Windows then?
> 
> I've googled a bit and people where removing the `SO_EXCLUSIVEADDRUSE`. What I read is that the reason this flag was added is not really existing anymore. But I'm not as good in Windows API as you are.

It is still mentioned [here ](https://learn.microsoft.com/en-us/windows/win32/winsock/so-exclusiveaddruse) and [here](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse), which is why I assume it is still in use, but with Microsoft you never know.

Fast googling brought up two Github issues, [one ](https://github.com/python-trio/trio/issues/928) seems to say that Windows fixed SO_REUSEADDR behavior, but the linked MSDN article contradicts this opinion. The[ other one](https://github.com/yhirose/cpp-httplib/issues/2011) removes it for functional reasons.
If you have addition sources, please link them, I will take a look on them

Perhaps to give a short intro to the issue:
`SO_REUSEADDR `on BSD/Linux TCP is mostly harmless, it just allows you to bind a port at `TIME_WAIT `state, but does not allow real reuse. It is more a convenient option for not having to wait for the timeout, as we actually did and currently do it.

On Windows however `SO_REUSEADDR` allows to really share a port and can be a security issue due to port hijacking. Earlier this was, even without `SO_REUSEADDR` , the standard Winsock behavior on a second bind, leading to `SO_EXCLUSIVEADDRUSE` being introduced, to suppress this behavior. Then Winsock added `SO_REUSEADDR` to remove this from standard behavior, but it still can be used for port hijacking. Only ports protected by `SO_EXCLUSIVEADDRUSE` are not susceptible for port hijacking.

According to the linked MSDN articles this seems to still be the case, so I would still add the option. 


Rethinking the whole issue and checking a bit (with the help of AI chats, but the points below are written by me):

- If we want to have the same consistent behavior on all platforms, remove `SO_REUSEADDR `on Linux/BSD, add `SO_EXCLUSIVEADDRUSE` -> Drawback: We have this 60s wait time again, Benefit: No port hijacking
- Keep `SO_REUSEADDR `on Linux/BSD, add `SO_EXCLUSIVEADDRUSE` -> Drawback: No consistent behavior: No delay on BSD/Linux, but timeout on Windows, Benefit: No port hijacking
- Keep it as it is: Ports can be hijacked on Windows, but no wait time on all plattforms
- Rewrite the Windows version, adding a startup layer, which most like wont crash and holds the sockets, with this no wait time and no port hijacking is possible (as long as the layer does not crash). Obvious drawback: Quite involved rewrite

azoitl:
> * If we want to have the same consistent behavior on all platforms, remove `SO_REUSEADDR `on Linux/BSD, add `SO_EXCLUSIVEADDRUSE` -> Drawback: We have this 60s wait time again, Benefit: No port hijacking

I think I would then go with this version. First of all it is saver and more correct. Second by now we check for ctrl+c and similar kills on Linux (and I think on Windows) and cleanly shut down and close all sockets. Therefore the wait would only affect people debugging and hard killing 4diac FORTE from the debugger.

core/arch/win32/src/win32socketinterf.cpp
```diff
@@ -13,6 +13,7 @@
  *  Jörg Walter - Windows XP compatibility, improve UDP multicast support
  *  Zhao Xin - fix socket resource leakage
  *  Ketut Kumajaya - switch to the Unicode version of WSAStringToAddress
+ *..Franz Höpfinger - set SO_EXCLUSIVEADDRUSE to avoid 2 FORTE instances can access the port.
```
azoitl: please remove the dots here.

core/arch/win32/src/win32socketinterf.cpp
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2023 ACIN, Profactor GmbH, AIT, fortiss GmbH, OFFIS e.V., HIT robot group
- *               2024 Samator Indo Gas
+ * Copyright (c) 2010, 2026 ACIN, Profactor GmbH, AIT, fortiss GmbH, OFFIS e.V., HIT robot group
```
azoitl: as with the other PRs this is number can be removed

azoitl: as you are anyhow touching the copyright header please add a `,` after `HIT robot group`

azoitl (changes requested):

core/arch/win32/src/win32socketinterf.cpp
MartinMelikMerkumians: I would also change the comment to  "set SO_EXCLUSIVEADDRUSE to avoid port hijacking", as this is the term MSDN uses

MartinMelikMerkumians (changes requested):

ernstblechaPT:
> > * If we want to have the same consistent behavior on all platforms, remove `SO_REUSEADDR `on Linux/BSD, add `SO_EXCLUSIVEADDRUSE` -> Drawback: We have this 60s wait time again, Benefit: No port hijacking
> 
> I think I would then go with this version. First of all it is saver and more correct. Second by now we check for ctrl+c and similar kills on Linux (and I think on Windows) and cleanly shut down and close all sockets. Therefore the wait would only affect people debugging and hard killing 4diac FORTE from the debugger.

That will affect everyone using the "Local 4diac FORTE" Launcher inside 4diac-ide when the forte is closed via the "Terminate" option - there already is a "Stop 4diac FORTEs" Launcher - but that one also just kills the process and does not send the correct signal.

azoitl:
> > > * If we want to have the same consistent behavior on all platforms, remove `SO_REUSEADDR `on Linux/BSD, add `SO_EXCLUSIVEADDRUSE` -> Drawback: We have this 60s wait time again, Benefit: No port hijacking
> > 
> > 
> > I think I would then go with this version. First of all it is saver and more correct. Second by now we check for ctrl+c and similar kills on Linux (and I think on Windows) and cleanly shut down and close all sockets. Therefore the wait would only affect people debugging and hard killing 4diac FORTE from the debugger.
> 
> That will affect everyone using the "Local 4diac FORTE" Launcher inside 4diac-ide when the forte is closed via the "Terminate" option - there already is a "Stop 4diac FORTEs" Launcher - but that one also just kills the process and does not send the correct signal.

@ernstblechaPT I just tested on Linux there the "Stop 4diac FORTE" Launcher does the right thing. I tried manually and also in a launch group. The only thing that you need to do with the Launch group is to check the wait for termination for the "Stop 4diac FORTE" Launcher. I have not yet tested under Windows with `SO_EXCLUSIVEADDRUSE`. But I hope it would be the same.

ernstblechaPT:
> > > > * If we want to have the same consistent behavior on all platforms, remove `SO_REUSEADDR `on Linux/BSD, add `SO_EXCLUSIVEADDRUSE` -> Drawback: We have this 60s wait time again, Benefit: No port hijacking
> > > 
> > > 
> > > I think I would then go with this version. First of all it is saver and more correct. Second by now we check for ctrl+c and similar kills on Linux (and I think on Windows) and cleanly shut down and close all sockets. Therefore the wait would only affect people debugging and hard killing 4diac FORTE from the debugger.
> > 
> > 
> > That will affect everyone using the "Local 4diac FORTE" Launcher inside 4diac-ide when the forte is closed via the "Terminate" option - there already is a "Stop 4diac FORTEs" Launcher - but that one also just kills the process and does not send the correct signal.
> 
> @ernstblechaPT I just tested on Linux there the "Stop 4diac FORTE" Launcher does the right thing. I tried manually and also in a launch group. The only thing that you need to do with the Launch group is to check the wait for termination for the "Stop 4diac FORTE" Launcher. I have not yet tested under Windows with `SO_EXCLUSIVEADDRUSE`. But I hope it would be the same.

Tried on Windows: when stopping FORTE with CTRL+C on the command line I get
```
INFO: DT#2026-04-08-04:21:59.035: FORTE is up and running
INFO: DT#2026-04-08-04:21:59.036: Using provided bootfile location set in CMake: forte.fboot
INFO: DT#2026-04-08-04:21:59.036: Boot file forte.fboot could not be opened. Skipping...
INFO: DT#2026-04-08-04:21:59.036: CWin32SocketInterface: Opening TCP-Server connection at: localhost:61499
DEBUG: DT#2026-04-08-04:22:01.065: CSocketBaseLayer::closeConnection()
INFO: DT#2026-04-08-04:22:01.066: FORTE finished
```

the ouput inside 4diac-ide after termination is only
```
INFO: DT#2026-04-08-04:23:41.000: FORTE is up and running
INFO: DT#2026-04-08-04:23:41.000: Using provided bootfile location set in CMake: forte.fboot
INFO: DT#2026-04-08-04:23:41.000: Boot file forte.fboot could not be opened. Skipping...
INFO: DT#2026-04-08-04:23:41.000: CWin32SocketInterface: Opening TCP-Server connection at: localhost:61499
```

franz-hoepfinger-4diac:
i see: 

`https://github.com/eclipse-4diac/4diac-forte/pull/830 
#830 opened last week by [franz-hoepfinger-4diac](https://github.com/eclipse-4diac/4diac-forte/issues?q=is%3Apr+is%3Aopen+author%3Afranz-hoepfinger-4diac) Contributor
• [Changes requested](https://github.com/eclipse-4diac/4diac-forte/pull/830#partial-pull-merging)`

BUT for me it is unclear what Changes requested. pls give a Hint.

azoitl:
> i see:
> 
> `https://github.com/eclipse-4diac/4diac-forte/pull/830 #830 opened last week by [franz-hoepfinger-4diac](https://github.com/eclipse-4diac/4diac-forte/issues?q=is%3Apr+is%3Aopen+author%3Afranz-hoepfinger-4diac) Contributor • [Changes requested](https://github.com/eclipse-4diac/4diac-forte/pull/830#partial-pull-merging)`
> 
> BUT for me it is unclear what Changes requested. pls give a Hint.

You already addressed the requested changes. However with the concerned raised by @ernstblechaPT I have to think about what your change means for 4diac IDE. As I had little time this week for Eclipse 4diac and not yet a windows machine for testing it is a bit hard to provide a quick response.

franz-hoepfinger-4diac:
> You already addressed the requested changes. However with the concerned raised by @ernstblechaPT I have to think about what your change means for 4diac IDE. As I had little time this week for Eclipse 4diac and not yet a windows machine for testing it is a bit hard to provide a quick response.

@azoitl @ernstblechaPT 
BTW: i have a Windows machine with the relevant Compiler now. 
i did tests the SO_EXCLUSIVEADDRUSE https://github.com/eclipse-4diac/4diac-forte/issues/924 and documented it here. 
i will now do the other test as well. 

without https://github.com/eclipse-4diac/4diac-forte/pull/830 opening 2 Windows makes this: 


1.
```
Windows PowerShell
Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

Installieren Sie die neueste PowerShell für neue Funktionen und Verbesserungen! https://aka.ms/PSWindows

PS C:\git\4d\4diac-forte\bin\windows> .\forte.exe
INFO: T#1779018669301801900: FORTE is up and running
INFO: T#1779018669302122300: Using provided bootfile location set in CMake: forte.fboot
INFO: T#1779018669302721600: Boot file forte.fboot could not be opened. Skipping...
INFO: T#1779018669303172700: CWin32SocketInterface: Opening TCP-Server connection at: localhost:61499
```
2.
```
Windows PowerShell
Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

Installieren Sie die neueste PowerShell für neue Funktionen und Verbesserungen! https://aka.ms/PSWindows

PS C:\git\4d\4diac-forte\bin\windows> .\forte.exe
INFO: T#1779018663390405300: FORTE is up and running
INFO: T#1779018663391080100: Using provided bootfile location set in CMake: forte.fboot
INFO: T#1779018663391553800: Boot file forte.fboot could not be opened. Skipping...
INFO: T#1779018663391819900: CWin32SocketInterface: Opening TCP-Server connection at: localhost:61499
```

No Error Message, nothing !

azoitl:

PR #831, opened 2026-04-05
Bugfix: logging in BSD socket interface

franz-hoepfinger-4diac:
Update the preprocessor logic to ensure that connection information is logged and variables are not marked as unused when FORTE_LOGDEBUG is active.

### DIFF
--- a/core/arch/common/src/bsdsocketinterf.cpp
+++ b/core/arch/common/src/bsdsocketinterf.cpp
@@ -31,11 +31,7 @@ namespace forte::arch {
                                                                                       unsigned short paPort) {
     TSocketDescriptor nRetVal = -1;
 
-#ifndef FORTE_LOGINFO
-    (void) paIPAddr;
-#else
     DEVLOG_INFO("CBSDSocketInterface: Opening TCP-Server connection at: %s:%d\n", paIPAddr, paPort);
-#endif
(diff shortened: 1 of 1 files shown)


azoitl (changes requested):
It took me a while to understand the core problem what you like to solve here. Unfortunately neither the PR message nor the commit message where helping here. Because this is not an improvement this is a bug fix. 

The other reason I had to think over is your proposed solution. The problem I have with it is that we know have to change at least two places if we change when DEVLOG_INFO shall log. 

I think the cleaner and more correct solution is to remove the `#ifndef .. #else`  as a whole and add a commit to `(void) paIPAddr;` telling that this line is here to have no warning of `DEVLOG_INFO` is implemented empty. 

Finally as this is a very small change that is not really adding a copyright an update of the copyright header is not necessary.

azoitl (changes requested):
I'm sorry this is not what I had in mind. The part `(void) paIPAddr;` needs to stay as otherwise you get warnings on log levels where info is not logged. Therefore also some comment should be there why this line is needed.

Finally you ignored my feedback to the commit message.

azoitl:
As my comments where ignored and another PR was changing similar lines I'm closing this in favor of PR #884.

PR #833, opened 2026-04-07
Add IEC 61131-3 endian conversion function blocks

franz-hoepfinger-4diac:
Introduces F_FROM_BIG_ENDIAN, F_FROM_LITTLE_ENDIAN, F_TO_BIG_ENDIAN,
and F_TO_LITTLE_ENDIAN function blocks to handle endianness conversions
for IEC 61131-3 data types.

These blocks were generated using the 4DIAC FORTE Export Filter.

https://github.com/eclipse-4diac/4diac-ide/pull/2084


azoitl (commented):
I don't think that the IEC 61131-3 functions should have variant support. Am I right @mx990 and @MartinMelikMerkumians?

Furthermore to copyright headers of your newly added files have the wrong history information

franz-hoepfinger-4diac:
> I don't think that the IEC 61131-3 functions should have variant support. Am I right @mx990 and @MartinMelikMerkumians?
> 
i do not understand ? 
do you mean: `func_` should not be touched ? 

how should the Solution look then ? the generated Code does not compile without these modifications, but there might be another way to solve it. 



> Furthermore to copyright headers of your newly added files have the wrong history information

i do not understand this either. 
it was said in FBs, the top-most entry should be the newest. This is the case here, as it is generated by the IDE.

MartinMelikMerkumians:
For the FB as a SFB I think the VARIANT type is necessary. As the VARIANT types are a somehow parallel type hierarchy, those functions should not be chosen by the compiler (at least I think so).

However, introducing new functions, also calling variants, should also always provide tests, testing the function if the results/behavior is as expected. Especially if templates are involved too.

mx990:
> I don't think that the IEC 61131-3 functions should have variant support. Am I right @mx990 and @MartinMelikMerkumians?

I also believe the IEC 61131-3 functions should not have variant support. While we do support working with generic interface variables to some degree, overall support for generic data types in ST code is still very limited. This reflects the IEC 61131-3 standard, which considers the usage of generic data types in user-declared program organization units beyond its scope[^1]. I would therefore avoid adding variant support to IEC 61131-3 functions, at least until we have a clear understanding on how to support generic data types in ST code overall.

> how should the Solution look then ? the generated Code does not compile without these modifications, but there might be another way to solve it.

The idiomatic way when dealing with generic interface variables is to declare it as an SIFB and implement the `std::visit` pattern in the `executeEvent(...)` function[^2].

[^1]: see also *IEC 61131-3 Ed. 3, 6.4.3 Generic data types*
[^2]: see also [modules/IEC61131-3/src/iec61131/charString/F_CONCAT_fbt.cpp:59-78](https://github.com/eclipse-4diac/4diac-forte/blob/6fec7499185fe05784ad43afaee4f857e1719f0a/modules/IEC61131-3/src/iec61131/charString/F_CONCAT_fbt.cpp#L59-L78)

azoitl:
@mx990 thx for your assessment this is what I had guessed. 

@franz-hoepfinger-4diac the reason why the generated code is not working and should not work is that the ST code in your SFB is not correct ST code. I originally thought just adjusting the generated C++ code would be the way to go. So the code that you currently have in your new functions should be moved into the FB code. 

@mx990 would Franz's current SFB be interpretable in 4diac IDE?  If yes I would prefer to keep them as SFBs.

franz-hoepfinger-4diac:
> @franz-hoepfinger-4diac the reason why the generated code is not working and should not work is that the ST code in your SFB is not correct ST code.

do we need an issue in the IDE, that non-correct ST Code is not marked as an Error ?

mx990:
> @mx990 would Franz's current SFB be interpretable in 4diac IDE? If yes I would prefer to keep them as SFBs.

That is a good point. Yes, they are interpretable in 4diac IDE. I agree, it makes sense to keep them as SFBs.

> do we need an issue in the IDE, that non-correct ST Code is not marked as an Error ?

If we did that, then we would have to remove the implementation. I would therefore keep the status quo for now.

azoitl (commented):
This PR does not look right. why are two of the 4 FBs and functions implemented differently? Please also clean the commit history.

diplfranzhoepfinger:
> This PR does not look right. why are two of the 4 FBs and functions implemented differently? 

because they compile. 
i would also preffer to have them implemented 1:1 same, because otherwise if we look to it in 4 years, we will not recognize. 

> Please also clean the commit history.

done

azoitl (changes requested):
all code of these functionblocks should be in the function blocks by adding stuff to the generic functions you enable use-cases that are not permitted by IEC 61131.3.

diplfranzhoepfinger:
> all code of these functionblocks should be in the function blocks by adding stuff to the generic functions you enable use-cases that are not permitted by IEC 61131.3.

you mean: these 2 Files should remain unchanged ? 
func_FROM_BIG_ENDIAN.h
func_TO_BIG_ENDIAN.h

azoitl:
> > all code of these functionblocks should be in the function blocks by adding stuff to the generic functions you enable use-cases that are not permitted by IEC 61131.3.
> 
> you mean: these 2 Files should remain unchanged ? func_FROM_BIG_ENDIAN.h func_TO_BIG_ENDIAN.h

yes

modules/IEC61131-3/src/iec61131/conversion/F_FROM_BIG_ENDIAN_fbt.cpp
```diff
@@ -0,0 +1,157 @@
...
+        [](auto &&paIN) -> CIEC_ANY_VARIANT {
+          using T = std::decay_t<decltype(paIN)>;
+          if constexpr (!std::is_same_v<T, CIEC_ANY_UNIQUE_PTR<CIEC_ARRAY>> &&
+                        !std::is_same_v<T, CIEC_ANY_UNIQUE_PTR<CIEC_STRUCT>>) {
+            return func_FROM_BIG_ENDIAN(paIN);
+          } else {
+            return paIN;
+          }
+        },
+        static_cast<CIEC_ANY_VARIANT::variant &>(var_IN));
```
MartinMelikMerkumians: Perhaps I am wrong, but won't that exclude arrays and structs from the call to change endianess? Have you tested this?

franz-hoepfinger-4diac: exactly. 
i did it other way round before (extended `func_FROM_BIG_ENDIAN`) so it accepts array and struct, and i was told this is not allowed. not i exclude it here, so the call never goes to `func_FROM_BIG_ENDIAN`. 

I am very open to any other solution! What do you think ?

azoitl: @franz-hoepfinger-4diac, as we discussed either above or in one of your other PRs,  the reason for not changing the implementation of the functions is that this would open up the use of ANY variables in ST Code. This is strictly forbidden in IEC 6131-3 . Even though we check that in 4diac IDE we want to have this second line of defense here in 4diac FORTE.

However this does not mean you can implement the FBs here have right. It means that you have to do the right means here to call the right functions from the FBs C++ code.

azoitl: thinking a bit as this PR requires FB specific fixes I think also unit tests for these FBs would be cool

mx990: I would suggest using the following implementation in the FB type:
```cpp
#include "forte/iec61131_functions/endianess.h"
[...]
    var_OUT = var_IN;
    if constexpr (std::endian::native == std::endian::little) {
      std::visit(
          []<typename T>(T &&paOUT) {
            if constexpr (std::is_same_v<std::remove_reference_t<T>, CIEC_ANY_UNIQUE_PTR<CIEC_ARRAY>> ||
                          std::is_same_v<std::remove_reference_t<T>, CIEC_ANY_UNIQUE_PTR<CIEC_STRUCT>>) {
              swapEndianessInplace(*paOUT);
            } else {
              paOUT = swapEndianess(paOUT);
            }
          },
          static_cast<CIEC_ANY_VARIANT::variant &>(var_OUT));
    }
```

This avoids the `std::visit` entirely when the endianness matches, thus saving on binary size. It also avoids unnecessary copies of the array or struct by swapping them inplace.

franz-hoepfinger-4diac:
just studying over the Tests: 

in IEC 61131-3 Table 37 all 4 Endian Conversion Functions are defined as "ANY". 
so shouldn't also our func_ accept any Subtype of "ANY" (not ANY itself !) ??

EDIT: i guess this is already the case.

modules/IEC61131-3/src/iec61131/conversion/F_FROM_BIG_ENDIAN_fbt.cpp
franz-hoepfinger-4diac: > thinking a bit as this PR requires FB specific fixes I think also unit tests for these FBs would be cool

made SOME Tests already, but need to do Array and Struct still.

azoitl:
> just studying over the Tests:
> 
> in IEC 61131-3 Table 37 all 4 Endian Conversion Functions are defined as "ANY". so shouldn't also our func_ accept any Subtype of "ANY" (not ANY itself !) ??
> 
> EDIT: i guess this is already the case.

Yes it is but in a type save manner. Which is important!

franz-hoepfinger-4diac:
Ready for Review. 
pls tell if we should add "Struct" Tests as well. 
Array Tests are in.

modules/IEC61131-3/src/iec61131/conversion/F_TO_LITTLE_ENDIAN_fbt.cpp
```diff
@@ -0,0 +1,153 @@
...
+      std::visit(
+          []<typename T>(T &&paOUT) {
+            if constexpr (std::is_same_v<std::remove_reference_t<T>, CIEC_ANY_UNIQUE_PTR<CIEC_ARRAY>> ||
+                          std::is_same_v<std::remove_reference_t<T>, CIEC_ANY_UNIQUE_PTR<CIEC_STRUCT>>) {
+              paOUT = func_TO_LITTLE_ENDIAN(paOUT);
+            } else {
+              paOUT = func_TO_LITTLE_ENDIAN(paOUT);
+            }
+          },
+          static_cast<CIEC_ANY_VARIANT::variant &>(var_OUT));
```
franz-hoepfinger-4diac, line 150: is ist better to have 1:1 the same Code as in Big Endian ? 

should a else case be added with `std::endian::native == std::endian::big`

franz-hoepfinger-4diac: we need. Just realize that 4diac also works on PPC (see [setup_posix-ppc.sh](https://github.com/eclipse-4diac/4diac-forte/blob/release/setup_posix-ppc.sh), so it must run as well on a Big-Endian architecture. 
so, we also need the code to work correctly there. While PowerPC in the PC Area is not more used, it is still used in Control Systems: [BODAS controller RC5-6/40](https://www.boschrexroth.com/media/685fdf2e-ec8a-4a5e-91b4-4df5ab86e4c3) e.g. is a recent ECU. Also i was using PowerPC many many years with CODESYS (Berghof DC1000) No Idea if forte will still build on PPC, maybe nobody has tried for a long time.

azoitl:

PR #834, opened 2026-04-07
Remove SO_REUSEADDR from TCP Server Ports in BSD Socket Implementation

azoitl:
This socket option was added for quicker restarts when 4diac FORTE crashes or is killed during debug sessions on Windows. As we have a specific Windows socket implementation it should not be in the generic code anymore.


MartinMelikMerkumians (approved):
LGTM

mx990 (approved):

PR #835, opened 2026-04-07
Pins actions to hash

MartinMelikMerkumians:
Pins the action in accordance to the Eclipse security advice


azoitl (changes requested):
thx for the pinning. Can you please submit this for release as this is the default branch of this repo and this is also where we have to workflow changes first.

MartinMelikMerkumians:
Of course, I will rebase it to release

azoitl:
For future reference, with a force push and editing the PR you can move it to a different branch.

PR #836, opened 2026-04-07
Pins actions to hash

MartinMelikMerkumians:
@azoitl it was easier to cherry-pick and recreate the pull request. This is the follow up to #835

Sorry for the inconvenience


azoitl (approved):
I would then use this PR in favor of the dependabot PR #832.

PR #839, opened 2026-04-08
remove format

### LINKED ISSUE #514: Commit "Change all toString methods to use std::string", d161dc26, relies on format header (c++ 20), this breaks Zephyr target
The Zephyr SDK for ESP32 doen't use a version of GCC yet which supports `<format>`.
Hence, FORTE does not build.

antongus:


core/src/datatypes/forte_any_chars.cpp
```diff
@@ -29,7 +28,11 @@ void CIEC_ANY_CHARS::dollarEscapeChar(std::string &paTargetBuf, char paSymbol, c
     case '\"': paTargetBuf += (paTypeID == e_WSTRING) ? "$\"" : "\""s; break;
     default:
       if (!isprint(static_cast<unsigned char>(paSymbol)) && paTypeID == e_STRING) {
-        std::format_to(std::back_inserter(paTargetBuf), "${:02X}", static_cast<unsigned char>(paSymbol));
+        static const char scmHex[] = "0123456789ABCDEF";
+        const unsigned char c = static_cast<unsigned char>(paSymbol);
+        paTargetBuf += '$';
+        paTargetBuf += scmHex[(c >> 4) & 0xF];
+        paTargetBuf += scmHex[c & 0xF];
```
mx990: I believe simply calculating the ASCII value would be faster than a lookup in a string constant:
```cpp
digit < 10 ? '0' + digit : 'A' - 10 + digit
```
I would also suggest extracting this into a common method `toHexDigit`.

core/src/datatypes/forte_date.cpp
```diff
@@ -72,8 +73,9 @@ namespace forte {
   void CIEC_DATE::toString(std::string &paTargetBuf) const {
     tm ptm;
     if (nullptr != getTimeStruct(&ptm)) {
-      std::format_to(std::back_inserter(paTargetBuf), "D#{:04}-{:02}-{:02}", 1900 + ptm.tm_year, ptm.tm_mon + 1,
-                     ptm.tm_mday);
+      char buf[32];
+      snprintf(buf, sizeof(buf), "D#%04d-%02d-%02d", 1900 + ptm.tm_year, ptm.tm_mon + 1, ptm.tm_mday);
+      paTargetBuf.append(buf);
```
mx990: I believe we may be able to optimize this even further by appending individual digits, thus avoiding any formatting code altogether.

core/src/util/mainparam_utils.cpp
```diff
@@ -56,9 +56,9 @@ namespace forte::util {
                      const std::string_view paHelp,
                      std::size_t maxShortNameLength,
                      std::size_t maxLongNameLength) {
-      std::cout << std::format("  -{:{}} --{:{}} {} {}", std::string(paShortName) + ",", maxShortNameLength + 1,
-                               paLongName, maxLongNameLength, paArgumentName, paHelp)
-                << std::endl;
+      std::cout << "  -" << std::string(paShortName) + "," << std::string(maxShortNameLength + 1 - paShortName.size() - 1, ' ')
+                << " --" << paLongName << std::string(maxLongNameLength - paLongName.size(), ' ')
+                << " " << paArgumentName << " " << paHelp << std::endl;
```
mx990: I would suggest setting the field width for `paShortName` and `paLongName` using [`std::setw`](https://en.cppreference.com/w/cpp/io/manip/setw.html) instead of adding a padding string.

core/src/datatypes/forte_lreal.cpp
```diff
@@ -52,8 +52,11 @@ namespace forte {
 
   void CIEC_LREAL::toString(std::string &paTargetBuf) const {
     auto startPos = paTargetBuf.size();
-    std::format_to(std::back_inserter(paTargetBuf), "{:.{}g}", getTDFLOAT(),
-                   std::numeric_limits<TValueType>::max_digits10);
+    char buf[32];
+    snprintf(buf, sizeof(buf), "%.*g",
+             std::numeric_limits<TValueType>::max_digits10,
+             static_cast<double>(getTDFLOAT()));
+    paTargetBuf.append(buf);
```
mx990: This may lead to problems since `snprintf` is locale-dependent. I would therefore again recommend to use `std::to_chars` as outlined in https://github.com/eclipse-4diac/4diac-forte/issues/514#issuecomment-4204764262.

mx990 (changes requested):

core/src/datatypes/forte_any_chars.cpp
antongus: > I believe simply calculating the ASCII value would be faster than a lookup in a string constant:

Nothing can be faster than a lookup: [godbolt x86-64](https://godbolt.org/z/cKx45seK6), [godbolt arm](https://godbolt.org/z/cE6v1h8j8) :)

> I would also suggest extracting this into a common method `toHexDigit`.

Agree. Where should I place this common method? `string_utils.h`?

core/src/datatypes/forte_date.cpp
antongus: You mean something like

```
  void CIEC_DATE::toString(std::string &paTargetBuf) const {
    tm ptm;
    if (nullptr != getTimeStruct(&ptm)) {
      char buf[] = "D#0000-00-00";

      auto w = 1900 + ptm.tm_year;
      buf[5] = (w % 10) + '0';
      w /= 10;
      buf[4] = (w % 10) + '0';
      w /= 10;
      buf[3] = (w % 10) + '0';
      w /= 10;
      buf[2] = (w % 10) + '0';

      w = ptm.tm_mon + 1;
      buf[8] = (w % 10) + '0';
      w /= 10;
      buf[7] = (w % 10) + '0';

      w = ptm.tm_mday;
      buf[11] = (w % 10) + '0';
      w /= 10;
      buf[10] = (w % 10) + '0';

      paTargetBuf.append(buf);
    }
  }
```
?

core/src/datatypes/forte_lreal.cpp
antongus: When I change this code to:

```cpp
#include <charconv>

  void CIEC_LREAL::toString(std::string &paTargetBuf) const {
    char buf[32];
    auto [ptr, ec] = std::to_chars(std::begin(buf), std::end(buf), static_cast<double>(getTDFLOAT()), 
        std::chars_format::general, std::numeric_limits<TValueType>::max_digits10);
    auto startPos = paTargetBuf.size();
    paTargetBuf.append(buf);
    normalizeToStringRepresentation(paTargetBuf, startPos);
  }
```

Code size increased by 111KB. Not good. Maybe just replace comma (if any) with dot after printf to ensure locale-independent representation?

azoitl: Have you tried getting rid of snprintf as a whole. From previous work with ECos on arm 7 I know that snprintf is also requiring a lot of flash. Given all the other problems snprintf has I would vote for getting rid of it.

core/src/datatypes/forte_any_chars.cpp
azoitl: > > I believe simply calculating the ASCII value would be faster than a lookup in a string constant:
> 
> Nothing can be faster than a lookup: [godbolt x86-64](https://godbolt.org/z/cKx45seK6), [godbolt arm](https://godbolt.org/z/cE6v1h8j8) :)

I'm here with @mx990 even if the lookup is fat it requires an additional memory access. So to paraphrase you: nothing beats a simple addition.

mx990: > > I would also suggest extracting this into a common method `toHexDigit`.
> 
> Agree. Where should I place this common method? `string_utils.h`?

Yes, I also believe `string_utils.h` would be best.

core/src/datatypes/forte_date.cpp
mx990: Yes, basically. I would suggest to add a helper function to append a fixed number of digits to a string:
```cpp
  template<int N, std::integral T>
  void appendDigits(std::string &paTargetBuf, T paValue) {
    char buf[N];
    for (int i = 0; i < N; i++) {
      buf[N - i - 1] = '0' + (paValue % 10);
      paValue /= 10;
    }
    paTargetBuf.append(buf, N);
  }
```
which would result in:
```cpp
  void CIEC_DATE::toString(std::string &paTargetBuf) const {
    tm ptm;
    if (nullptr != getTimeStruct(&ptm)) {
      paTargetBuf.reserve(paTargetBuf.size() + 12); // "D#1970-01-01"
      paTargetBuf += "D#";
      appendDigits<4>(paTargetBuf, ptm.tm_year + 1900);
      paTargetBuf += "-";
      appendDigits<2>(paTargetBuf, ptm.tm_mon + 1);
      paTargetBuf += "-";
      appendDigits<2>(paTargetBuf, ptm.tm_mday);
    }
  }
```

azoitl:
@antongus I really like the hexchar version now. Would you mind extracting it into an own PR?

antongus:
> @antongus I really like the hexchar version now. Would you mind extracting it into an own PR?

done: https://github.com/eclipse-4diac/4diac-forte/pull/843

core/src/datatypes/forte_lreal.cpp
antongus: > Have you tried getting rid of snprintf as a whole.

@azoitl , printf functions are used not only by forte, but also by LwIP, FreeRTOS. So getting rid of this is almost impossible (at least for my case).

core/src/util/string_utils.cpp
```diff
@@ -346,4 +346,21 @@ namespace forte {
     paTargetBuf += ":="s;
     variable->toString(paTargetBuf);
   }
+
+void util::appendInt(std::string &paTargetBuf, size_t width, int value) {
```
mx990: Passing the size via a template parameter, as suggested in #839 (comment), allows the compiler to better optimize, for example, by unrolling the for loop. Please also extract this and the calling locations into a separate PR.

core/src/util/string_utils.cpp
```diff
@@ -346,4 +346,21 @@ namespace forte {
...
+  if (width) {
+    if (value < 0) {
+      value = -value;
+      paTargetBuf += '-';
+      --width;
+    }
```
mx990: There should be no need to handle negative values here. Any negative values would also produce an invalid string value anyhow.

core/include/forte/datatypes/forte_any_date.h
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 #include "forte/datatypes/forte_any_elementary.h"
+#include "forte/arch/forte_architecture_time.h"
```
mx990, line 19: Why did you add this additional include?

core/src/datatypes/forte_date.cpp
```diff
@@ -72,8 +74,12 @@ namespace forte {
...
     tm ptm;
     if (nullptr != getTimeStruct(&ptm)) {
-      std::format_to(std::back_inserter(paTargetBuf), "D#{:04}-{:02}-{:02}", 1900 + ptm.tm_year, ptm.tm_mon + 1,
-                     ptm.tm_mday);
+      paTargetBuf += "D#";
+      forte::util::appendInt(paTargetBuf, 4, 1900 + ptm.tm_year);
+      paTargetBuf += "-";
+      forte::util::appendInt(paTargetBuf, 2, ptm.tm_mon + 1);
+      paTargetBuf += "-";
+      forte::util::appendInt(paTargetBuf, 2, ptm.tm_mday);
```
mx990: Please [`reserve`](https://en.cppreference.com/w/cpp/string/basic_string/reserve.html) the expected size at the beginning to avoid fragmented allocations. This also goes for the other locations.

core/src/util/mainparam_utils.cpp
```diff
@@ -56,9 +56,9 @@ namespace forte::util {
                      const std::string_view paHelp,
                      std::size_t maxShortNameLength,
                      std::size_t maxLongNameLength) {
-      std::cout << std::format("  -{:{}} --{:{}} {} {}", std::string(paShortName) + ",", maxShortNameLength + 1,
-                               paLongName, maxLongNameLength, paArgumentName, paHelp)
-                << std::endl;
+      std::cout << "  -" << std::left << std::setw(maxShortNameLength + 1) << std::string(paShortName) + ","
+                << " --" << std::setw(maxLongNameLength) << paLongName
+                << " " << paArgumentName << " " << paHelp << std::endl;
```
mx990: Please extract this into a separate PR.

core/src/util/string_utils.cpp
```diff
@@ -346,4 +346,21 @@ namespace forte {
...
+    paTargetBuf.append(width, ' ');
```
mx990: You can just use `resize`, since we are overwriting all the added characters anyhow.

core/src/datatypes/forte_lreal.cpp
mx990: > Maybe just replace comma (if any) with dot after printf to ensure locale-independent representation?

Accommodating the differences and quirks from all potential locales seems very fragile. I would also point out that, in terms of performance, `std::to_chars` from libc++ was measured to be **17 times** faster than `sprintf` in a recent benchmark[^1]. I was also able to confirm those numbers with a quick benchmark on my machine.

[^1]: see https://vitaut.net/posts/2025/to-chars/

mx990 (changes requested):
Please use `clang-format` on the changed files to fix the formatting[^1].

[^1]: see also https://clang.llvm.org/docs/ClangFormat.html

core/include/forte/datatypes/forte_any_date.h
antongus: > Why did you add this additional include?

This is because of `struct tm` usage.  Without this include compilation fails for freertos arch.

core/src/datatypes/forte_date.cpp
antongus: `reserve()` (in most implementations) just checks capacity and doubles it if it less than requested new capacity. It is exactiy the same behavior as when I just append something to the string.

antongus:
> Accommodating the differences and quirks from all potential locales seems very fragile.

@mx990 I totally agree. But my overall point in this PR is code size reduction. All other variants of float/double to string conversion leads to code grow (`format` - +170KB, `to_chars` - +110-120KB).

I have an idea how to handle this: move string formatting routines to arch. But it is for future.

core/src/util/string_utils.cpp
antongus: > Please also extract this and the calling locations into a separate PR.

Done: https://github.com/eclipse-4diac/4diac-forte/pull/853

core/src/util/mainparam_utils.cpp
antongus: > Please extract this into a separate PR.

Done: https://github.com/eclipse-4diac/4diac-forte/pull/852

azoitl:
> > Accommodating the differences and quirks from all potential locales seems very fragile.
> 
> @mx990 I totally agree. But my overall point in this PR is code size reduction. All other variants of float/double to string conversion leads to code grow (`format` - +170KB, `to_chars` - +110-120KB).
> 
> I have an idea how to handle this: move string formatting routines to arch. But it is for future.

@antongus I have mixed feelings about this change. I'm a big fan of reducing code size and even more reducing ram consumption of 4diac FORTE. In my experience as already written RAM usage is nowadays still the more important fact. With all the work we spent on 4diac FORTE 3.0 to remove old C code for more correct and often faster C++ code I'm not happy that we should now go back. Especially when also the performance and the RAM usage strongly speak against it. 

These are all important functions needed for correct function of 4diac FORTE. They are on an execution critical path when monitoring is turned on. Therefore I'm against moving these functions to arch. This would make it very easy to break 4diac FORTE when porting 4diac FORTE. Also it increases the porting effort.

antongus:
> @antongus I have mixed feelings about this change. I'm a big fan of reducing code size and even more reducing ram consumption of 4diac FORTE. In my experience as already written RAM usage is nowadays still the more important fact.

@azoitl , as for RAM usage there is no penalties in current PR code. I eliminated local buffers and writes are now preformed directly to string.

> These are all important functions needed for correct function of 4diac FORTE. They are on an execution critical path when monitoring is turned on. Therefore I'm against moving these functions to arch. This would make it very easy to break 4diac FORTE when porting 4diac FORTE. Also it increases the porting effort.

My idea was just adding something like `forte/arch/forte_architecture_format.h` with wrappers around `std::format` and use it in all architectures which are no code size restricted. Something like:

```c++
namespace forte::arch {
  template <typename ... Args>
  std::string format(const std::format_string<Args...>fmt, Args&&... args)
  {
    return std::format(fmt, std::forward<Args>(args)...);
  }

  template <typename It, typename ... Args>
  void format_to(It it, const std::format_string<Args...>fmt, Args&&... args)
  {
    std::format_to(it, fmt, std::forward<Args>(args)...);
  }
  ... etc
} // namespace forte::arch
```

This allows fully use modern std::format in most architectures, and also allows customization for space-restricted arches.

antongus:
Please see this: https://github.com/eclipse-4diac/4diac-forte/pull/869

(I think new solution much better than this PR).

mx990 (approved):

azoitl (changes requested):
@antongus could you please update the commit message so that it better reflects the intent of the pr

azoitl (approved):
thx

PR #850, opened 2026-04-14
Add generic event rendezvous function block (GEN_E_REND)

franz-hoepfinger-4diac:
This function block provides event synchronization for a configurable number of inputs. The count of event inputs (EI) is determined dynamically from the FB instance name (e.g., GEN_E_REND_3). An output event (EO) is fired once all EIs have been received, and a reset input (R) clears the internal state.

https://github.com/eclipse-4diac/4diac-ide/pull/2087


azoitl (changes requested):
As the output event interface is fixed it would be better to have the event output interface related fields and functions also fixed. 

Also for this kind of generic FB you should also provide some FB tests.

stdfblib/events/src/GEN_E_REND_fbt.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+using namespace forte::literals;
+
+namespace forte::iec61499::events {
+  // Registers the generic FB with the name "GEN_E_REND" in the FORTE type library.
+  DEFINE_GENERIC_FIRMWARE_FB(GEN_E_REND, "iec61499::events::GEN_E_REND"_STRID)
+
+  namespace {
+    // Defines the constant name for the single event output.
+    const auto cEventOutputNames = std::array{"EO"_STRID};
```
franz-hoepfinger-4diac, line 28: this is a fixed Name, isn't it ?

franz-hoepfinger-4diac:
> Also for this kind of generic FB you should also provide some FB tests.

I added tests for E_REND, E_REND_3 and RT_E_REND

do we need also Tests for E_REND_2 and E_REND_4 ? it would not be a big work to make them as well !

stdfblib/events/src/GEN_E_REND_fbt.cpp
azoitl: yes but also add the event output connection you can look at the generic muxers for reference.

azoitl:
> > Also for this kind of generic FB you should also provide some FB tests.
> 
> I added tests for E_REND, E_REND_3 and RT_E_REND
> 
> do we need also Tests for E_REND_2 and E_REND_4 ? it would not be a big work to make them as well !

Please only add tests for the FBs you added. I would expect to have at least some tests for GEN_REND blocks with more then 2 events so that we see if the GEN block is really working.

franz-hoepfinger-4diac:
> Please only add tests for the FBs you added. I would expect to have at least some tests for GEN_REND blocks with more then 2 events so that we see if the GEN block is really working.

made seperate PRs for the other Blocks, to keep this Commit clean. 

now we are testing _2, _3, _4

azoitl (changes requested):
Thx for the clean-up and also for adding testcases. 

However similar to the other tests I think it would be good to extract constants for the event numbers into an anonymous namesspace to make the code better read- and maintainable.

azoitl (approved):

PR #857, opened 2026-04-16
Add tests for E_REND function block

franz-hoepfinger-4diac:
Introduces unit tests for the standard (E_REND) event rendezvous function block. These tests verify proper event synchronization, sequence independence, and reset behavior.


azoitl (changes requested):
Would it make sense to add constants for `EI1`, `EI2` and `EO`. I think that the code would be more readable.

I think it would also be good to have a test with several `EI1` after each other followed by several `EI2` to show that that is also correctly working.

azoitl (approved):

PR #863, opened 2026-04-17
Use static FreeRTOS task creation for forte threads

franz-hoepfinger-4diac:
Migrates CFreeRTOSThread to use xTaskCreateStatic, explicitly allocating stack memory and providing a StaticTask_t buffer.

This enhances determinism and robustness by reducing reliance on FreeRTOS's internal dynamic heap, mitigating potential fragmentation. Improved error logging for task creation is also included.

Also on Systems with PSRAM, it can be more granulary controlled if Task should use DRAM or PSRAM in the settings of the RTOS.

this PR makes it hugely consistent to the other implementations in arch, using `mStack` as the Stack.

### DIFF
--- a/core/arch/common/include/forte/arch/threadbase.h
+++ b/core/arch/common/include/forte/arch/threadbase.h
@@ -90,6 +90,10 @@ namespace forte::arch {
       }
 
     protected:
+      constexpr long getStackSize() const {
+        return mStackSize;
+      }
+
       explicit CThreadBase(long paStackSize);
 
       virtual ~CThreadBase();
@@ -136,6 +140,7 @@ namespace forte::arch {
 
       /*! \brief create the thread and return a handle to it
        *
+       * @param paStackSize size of the stack in bytes to allocate for this thread
        * @return handle to the newly created thread
        */
       virtual TThreadHandleType createThread(long paStackSize) = 0;
--- a/core/arch/freeRTOS/include/forte/arch/forte_thread.h
+++ b/core/arch/freeRTOS/include/forte/arch/forte_thread.h
@@ -1,5 +1,5 @@
 /************************************************************************************
- * Copyright (c) 2016 fortiss GmbH
+ * Copyright (c) 2016 fortiss GmbH, HR Agartechnik GmbH
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -15,6 +15,8 @@
 
 #include <freertos/FreeRTOS.h>
 #include <freertos/task.h>
+#include <memory>
+#include <vector>
 
 #include "forte/datatype.h"
 #include "forte/util/devlog.h"
@@ -63,7 +65,6 @@ namespace forte::arch {
        *
        * @param pa_miliSeconds The miliseconds for the thread to sleep
        */
-
       static void sleepThread(unsigned int paMilliSeconds);
 
     protected:
@@ -87,7 +88,23 @@ namespace forte::arch {
        */
       virtual TThreadHandleType createThread(long paStackSize);
 
+      /*! \brief Convert a stack size in bytes to a FreeRTOS stack depth in words.
+       *
+       * FreeRTOS expects the stack depth as the number of StackType_t words, not bytes.
+       * @param paStackSize stack size in bytes
+       * @return stack depth in StackType_t words
+       */
+      constexpr size_t getStackDepth() {
+        return static_cast<size_t>(getStackSize()) / sizeof(StackType_t);
+      }
+
+      void deleteFreeRTOSTask();
+
       static const int scmForteTaskPriority;
+
+      std::vector<StackType_t> mFreeRTOSStack;
+      std::unique_ptr<StaticTask_t, decltype([](StaticTask_t *paTask) { vPortFree(paTask); })> mTask;
+      TaskHandle_t mFreeRTOSThreadHandle = nullptr;
   };
 
   typedef CFreeRTOSThread CThread; // allows that doxygen can generate better documenation
--- a/core/arch/freeRTOS/src/forte_thread.cpp
+++ b/core/arch/freeRTOS/src/forte_thread.cpp
@@ -1,5 +1,5 @@
 /************************************************************************************
- * Copyright (c) 2016 - 2018 fortiss GmbH
+ * Copyright (c) 2016 fortiss GmbH, HR Agartechnik GmbH
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -20,32 +20,65 @@ namespace forte::arch {
    */
   const int CFreeRTOSThread::scmForteTaskPriority = tskIDLE_PRIORITY + 4;
 
-  CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize) {
+  CFreeRTOSThread::CFreeRTOSThread(long paStackSize) :
+      CThreadBase(paStackSize),
+      mFreeRTOSStack(getStackDepth()),
+      /* Allocate the TCB via pvPortMalloc(), not as a member or via new, because
+       * on systems with PSRAM (e.g. ESP32-S3) the CFreeRTOSThread object itself
+       * may reside in PSRAM. FreeRTOS requires the TCB to be in internal SRAM;
+       * pvPortMalloc() guarantees MALLOC_CAP_INTERNAL on ESP-IDF. */
+      mTask{static_cast<StaticTask_t *>(pvPortMalloc(sizeof(StaticTask_t)))} {
+  }
+
+  void CFreeRTOSThread::deleteFreeRTOSTask() {
+    if (mFreeRTOSThreadHandle != nullptr) {
+      /* On SMP (dual-core ESP32) there is a narrow window between runThread() signalling
+       * mJoinSem and threadFunction() actually reaching vTaskDelay(). Calling vTaskDelete
+       * on a Running task across cores hangs on ESP-IDF SMP, so we wait. */
+      while (eTaskGetState(mFreeRTOSThreadHandle) == eRunning) {
+        vTaskDelay(1);
+      }
+      vTaskDelete(mFreeRTOSThreadHandle);
+      mFreeRTOSThreadHandle = nullptr;
+    }
   }
 
   CFreeRTOSThread::~CFreeRTOSThread() {
+    end();
+    /* After end(), run() has returned and join() completed, but the FreeRTOS task is
+     * still alive and parked in vTaskDelay(). Delete it now that TCB and stack are no
+     * longer needed. */
+    deleteFreeRTOSTask();
   }
 
   void CFreeRTOSThread::threadFunction(void *paData) {
-    CThreadBase::runThread(static_cast<CFreeRTOSThread *>(paData));
-    /* Tasks must not attempt to return from their implementing
-     function or otherwise exit.
-     https://www.freertos.org/implementing-a-FreeRTOS-task.html */
-    vTaskDelete(nullptr);
+    auto paThreadInst = static_cast<CFreeRTOSThread *>(paData);
+    CThreadBase::runThread(paThreadInst);
+    /* join() is notified via mJoinSem.inc() inside runThread(), before we reach here.
+     * Park the task so the TCB and stack remain valid until ~CFreeRTOSThread()
+     * calls vTaskDelete() after join() has returned. */
+    vTaskDelay(portMAX_DELAY);
   }
 
   CThreadBase<TaskHandle_t, TaskHandle_t(0), CFreeRTOSThread>::TThreadHandleType
   CFreeRTOSThread::createThread(long paStackSize) {
-    TaskHandle_t handle = 0;
-
-    if (pdPASS != xTaskCreate(threadFunction, "FORTE", paStackSize, this, scmForteTaskPriority, &handle)) {
-      DEVLOG_ERROR("Error: Could not create FreeRTOS Task thread!");
+    if (mFreeRTOSStack.empty() || !mTask) {
+      DEVLOG_ERROR("Error: Could not allocate memory for FreeRTOS Task!\n");
+      return TaskHandle_t(0);
     }
 
-    return handle;
+    /* Clean up any previously parked task before creating a new one (restart case). */
+    deleteFreeRTOSTask();
+    mFreeRTOSThreadHandle = xTaskCreateStatic(threadFunction, "FORTE", mFreeRTOSStack.size(), this,
+                                              scmForteTaskPriority, mFreeRTOSStack.data(), mTask.get());
+    if (mFreeRTOSThreadHandle == nullptr) {
+      DEVLOG_ERROR("Error: Could not create FreeRTOS Task thread!");
+    }
+    return mFreeRTOSThreadHandle;
   }
 
   void CFreeRTOSThread::sleepThread(unsigned int paMilliSeconds) {
     vTaskDelay(pdMS_TO_TICKS(paMilliSeconds));
   }
+
 } // namespace forte::arch
--- a/core/arch/freeRTOS/src/main.cpp
+++ b/core/arch/freeRTOS/src/main.cpp
@@ -1,5 +1,5 @@
 /************************************************************************************
- * Copyright (c) 2016, 2024 fortiss GmbH, Jose Cabral
+ * Copyright (c) 2016, 2024 fortiss GmbH, Jose Cabral, HR Agartechnik GmbH
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -14,6 +14,7 @@
 #include <freertos/task.h>
 
 #include "../c_interface/forte_c.h"
+#include "forte/util/devlog.h"
 
 namespace {
   const unsigned forteTaskPriority = tskIDLE_PRIORITY + 1;
@@ -43,7 +44,9 @@ int main() {
     return result;
   }
 
-  xTaskCreate(vForteTask, "forte", stackDepth, nullptr, forteTaskPriority, nullptr);
+  if (pdPASS != xTaskCreate(vForteTask, "forte", stackDepth, nullptr, forteTaskPriority, nullptr)) {
+    DEVLOG_ERROR("Failed to create forte task");
+  }
 
   vTaskStartScheduler();
 


core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,6 +21,7 @@ namespace forte::arch {
   const int CFreeRTOSThread::scmForteTaskPriority = tskIDLE_PRIORITY + 4;
 
   CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize) {
+    mStack = new char[paStackSize];
```
azoitl: somthing is of here. the type of `mStack` is `SaticTask_t` but you allocate a char array.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -38,7 +39,14 @@ namespace forte::arch {
...
     TaskHandle_t handle = 0;
 
-    if (pdPASS != xTaskCreate(threadFunction, "FORTE", paStackSize, this, scmForteTaskPriority, &handle)) {
+    if (mStack == nullptr) {
+      DEVLOG_ERROR("Error: Could not allocate memory for FreeRTOS Task!\n");
+      return TaskHandle_t(0);
+    }
+
+    handle = xTaskCreateStatic(threadFunction, "FORTE", paStackSize, this, scmForteTaskPriority,
+                               reinterpret_cast<StackType_t *>(mStack), &mTask);
```
azoitl: again why is the cast need why can the variable not be the right type. Without comments or fixing this is totally unmaintainable code.

azoitl (changes requested):

core/arch/freeRTOS/src/forte_thread.cpp
diplfranzhoepfinger: StaticTask_t mTask;
mTask/mStack

core/arch/freeRTOS/src/forte_thread.cpp
diplfranzhoepfinger: mStack is defined in the superclass; and it is here because signed/unsigned

core/arch/freeRTOS/src/forte_thread.cpp
azoitl: thx for the clarification. Mixed that up. However this allocation is wrong. after checking the freeRTOS documentation the stacksize needs to be given in number of `StackType_t` entries.  So to be correct you need to:

`
  mStack = new char[paStackSize* sizeof(StackType_t)];
`

core/arch/freeRTOS/src/forte_thread.cpp
azoitl: Thx mixed that up.

azoitl (changes requested):

core/arch/freeRTOS/src/forte_thread.cpp
franz-hoepfinger-4diac: almost. then the Stack gets 4 times the Size.

franz-hoepfinger-4diac:
```
    size_t stackDepth = static_cast<size_t>(paStackSize) / sizeof(StackType_t);
    mStack = reinterpret_cast<char*>(new StackType_t[stackDepth]);
```
i guess that way it is the most correct way, right ? 
avoiding misalignments on some ARM Platforms!

franz-hoepfinger-4diac: to be MOST correct, the delete then should be: 

`delete[] reinterpret_cast<StackType_t*>(mStack);`

azoitl (changes requested):
> almost. then the Stack gets 4 times the Size.

It depends how you want to interpret the  `paStackSize` parameter. 

But what I now don't like is that delete is wrong and that you need to cast stuff around. I think the better solution is that you defined your own 

`StackType_t *mFreeRTOSStack` 

field and not use `mStack` at all.

franz-hoepfinger-4diac:
> It depends how you want to interpret the `paStackSize` parameter.

```
       *  @param paStackSize the Size of the stack the thread is allowed to use. this class will
       *         allocate the stack size in bytes from the heap
```
i guess we should stay in bytes.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -38,7 +41,14 @@ namespace forte::arch {
   CFreeRTOSThread::createThread(long paStackSize) {
     TaskHandle_t handle = 0;
 
-    if (pdPASS != xTaskCreate(threadFunction, "FORTE", paStackSize, this, scmForteTaskPriority, &handle)) {
+    if (mStack == nullptr) {
```
azoitl: I guess this should now be `mFreeRTOSSTack`

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -38,7 +41,14 @@ namespace forte::arch {
...
+      DEVLOG_ERROR("Error: Could not allocate memory for FreeRTOS Task!\n");
+      return TaskHandle_t(0);
+    }
+
+    handle = xTaskCreateStatic(threadFunction, "FORTE", paStackSize / sizeof(StackType_t), this, scmForteTaskPriority,
```
azoitl: I think you should extract that into a `getStackDepth()` method, which would make the code better readable and clearer state what why we have that.

azoitl (changes requested):
> > It depends how you want to interpret the `paStackSize` parameter.
> 
> ```
>        *  @param paStackSize the Size of the stack the thread is allowed to use. this class will
>        *         allocate the stack size in bytes from the heap
> ```
> 
> i guess we should stay in bytes.

good point.

core/arch/freeRTOS/src/forte_thread.cpp
franz-hoepfinger-4diac: done

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,25 +21,85 @@ namespace forte::arch {
...
+      delete[] mFreeRTOSStack;
+    }
   }
 
   void CFreeRTOSThread::threadFunction(void *paData) {
-    CThreadBase::runThread(static_cast<CFreeRTOSThread *>(paData));
-    /* Tasks must not attempt to return from their implementing
-     function or otherwise exit.
-     https://www.freertos.org/implementing-a-FreeRTOS-task.html */
-    vTaskDelete(nullptr);
```
franz-hoepfinger-4diac, line 34: vTaskDelete(nullptr); does NOT call prvDeleteTCB, it is only called from the idle Task.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -48,4 +108,40 @@ namespace forte::arch {
...
+
+    /* Wait for the task to be suspended */
+    while (eRunning == eTaskGetState(xTaskToDelete)) {
+      vTaskDelay(1);
+    }
+
+    /* We can delete the task
+     * First, we must call `vTaskDelete` so that the port task delete callback is called.
+     */
+    vTaskDelete(xTaskToDelete);
```
franz-hoepfinger-4diac: vTaskDelete from another Task Context calls prvDeleteTCB immediately, no need to wait for the idle Task.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,25 +21,85 @@ namespace forte::arch {
...
+    }
+
+    if (nullptr != mDeleteSem) {
+      vSemaphoreDelete(mDeleteSem);
+    }
+
+    // it is safe to delete mTask here, because if vTaskDelete was called from not itself,
+    // FreeRTOS calls prvDeleteTCB. (see tasks.c)
+    if (nullptr != mTask) {
+      vPortFree(mTask);
```
franz-hoepfinger-4diac: if the destructor deletes mTask, and the idle Task later comes over the TCB, we get a 

`  assert failed: prvDeleteTCB tasks.c:4939 (pxTCB->ucStaticallyAllocated == ( ( uint8_t ) 2 ))    `

so we need to get sure, the TCB is deleted by prvDeleteTCB *before* the Destructor is running.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,25 +21,90 @@ namespace forte::arch {
...
-    CThreadBase::runThread(static_cast<CFreeRTOSThread *>(paData));
-    /* Tasks must not attempt to return from their implementing
-     function or otherwise exit.
-     https://www.freertos.org/implementing-a-FreeRTOS-task.html */
-    vTaskDelete(nullptr);
+    auto paThreadInst = static_cast<CFreeRTOSThread *>(paData);
+    CThreadBase::runThread(paThreadInst);
+    // do not call vTaskDelete(nullptr)
+    TaskHandle_t xTaskToDelete = xTaskGetCurrentTaskHandle();
+    auto *params = new (std::nothrow) SDeleteTaskParams{xTaskToDelete, paThreadInst->mDeleteSem};
```
franz-hoepfinger-4diac: If somebody finds a better idea than using this struct, i am happy for suggestions !

diplfranzhoepfinger:
it is TESTED on a ESP32S3 with PSRAM, putting the Stack into the PSRAM. 
Goal was not to use ESP32 freeRTOS addtions, but stay at vanilla freeRTOS.

franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-forte/pull/863/changes/103c04f4742be5dbd9863977d83ad74f3b35a750 this commit simplifies the first one massively. 

- not require a extra semaphore.
- only minimal change to the base-class
- reusable for similar Platforms.

core/arch/freeRTOS/src/forte_thread.cpp
franz-hoepfinger-4diac: solved with https://github.com/eclipse-4diac/4diac-forte/commit/103c04f4742be5dbd9863977d83ad74f3b35a750

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +98,22 @@ namespace forte::arch {
        */
       virtual TThreadHandleType createThread(long paStackSize);
 
+      /*! \brief Convert a stack size in bytes to a FreeRTOS stack depth in words.
+       *
+       * FreeRTOS expects the stack depth as the number of StackType_t words, not bytes.
+       * @param paStackSize stack size in bytes
+       * @return stack depth in StackType_t words
+       */
+      static constexpr size_t getStackDepth(long paStackSize) {
```
azoitl: why is this a static method? if it is a normal private method it wouldn't need a parameter and would be much easier to use.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,31 +21,69 @@ namespace forte::arch {
   const int CFreeRTOSThread::scmForteTaskPriority = tskIDLE_PRIORITY + 4;
 
   CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize) {
+    /* Allocate memory for the task's TCB. We use pvPortMalloc() here as the
+     * TCB must be in internal memory. */
+    mTask = static_cast<StaticTask_t *>(pvPortMalloc(sizeof(StaticTask_t)));
+    /* Allocate memory for the task's stack using C++ new to place it on the Heap
+     * std::nothrow because -fno-exceptions leads to undefined behaviour otherwise.
+     * */
+    mFreeRTOSStack = new (std::nothrow) StackType_t[getStackDepth(paStackSize)];
```
azoitl: we have nowhere the std::nothrow why here?

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,31 +21,69 @@ namespace forte::arch {
...
-    CThreadBase::runThread(static_cast<CFreeRTOSThread *>(paData));
-    /* Tasks must not attempt to return from their implementing
-     function or otherwise exit.
-     https://www.freertos.org/implementing-a-FreeRTOS-task.html */
-    vTaskDelete(nullptr);
+    auto paThreadInst = static_cast<CFreeRTOSThread *>(paData);
+    CThreadBase::runThread(paThreadInst);
+    /* Park the task. destroyThread() (called from end() or ~CFreeRTOSThread()) will
+     * delete this FreeRTOS task via vTaskDelete once it is safe to free the TCB/stack. */
+    vTaskDelay(portMAX_DELAY);
```
azoitl, line 60: how is now the join function notified that the task has completed?

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +97,21 @@ namespace forte::arch {
...
+        return static_cast<size_t>(paStackSize) / sizeof(StackType_t);
+      }
+
       static const int scmForteTaskPriority;
+
+      StackType_t *mFreeRTOSStack = nullptr;
+      StaticTask_t *mTask = nullptr;
```
azoitl: why cant this just be a not a pointer, this would reduce the code complexity a lot.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -21,31 +21,69 @@ namespace forte::arch {
...
+      return TaskHandle_t(0);
+    }
 
-    if (pdPASS != xTaskCreate(threadFunction, "FORTE", paStackSize, this, scmForteTaskPriority, &handle)) {
+    TaskHandle_t handle = xTaskCreateStatic(threadFunction, "FORTE", getStackDepth(paStackSize), this,
+                                            scmForteTaskPriority, mFreeRTOSStack, mTask);
+    if (handle == nullptr) {
       DEVLOG_ERROR("Error: Could not create FreeRTOS Task thread!");
+    } else {
+      mThreadCreated = true;
```
azoitl: why do you need this boolean? it is always equal to handle != nullptr

core/src/ecet.cpp
```diff
@@ -30,7 +30,13 @@ namespace forte {
     clear();
   }
 
-  CEventChainExecutionThread::~CEventChainExecutionThread() = default;
+  CEventChainExecutionThread::~CEventChainExecutionThread() {
+    /* Call end() while the vtable is still CEventChainExecutionThread's, so that
+     * onAliveChanged() correctly dispatches to resumeSelfSuspend() and wakes the
+     * thread from selfSuspend(). If we leave this to ~CFreeRTOSThread(), the vtable
+     * has already been reverted and onAliveChanged() no longer reaches ECET's override. */
+    end();
```
azoitl: have you tested this on all other platforms. this makes me extremely nervouse

azoitl (changes requested):
In addition to the comments below. Please clean-up the history and provide an explaining commit message what is the reason for this change and what it solves.

core/arch/freeRTOS/src/forte_thread.cpp
franz-hoepfinger-4diac: 100% agree, this must be discussed first. Starting a Discussion here: https://github.com/eclipse-4diac/4diac-forte/discussions/874

core/arch/freeRTOS/include/forte/arch/forte_thread.h
franz-hoepfinger-4diac: no, it can not. 
Explanation: 
On systems with PSRAM (e.g. ESP32-S3), the CFreeRTOSThread object itself may be allocated in PSRAM when created via new. FreeRTOS requires the TCB to reside in internal SRAM — placing it in PSRAM triggers an assertion in prvDeleteTCB. pvPortMalloc() on FreeRTOS guarantees MALLOC_CAP_INTERNAL, so the TCB always lands in SRAM regardless of where the surrounding object lives. This is now documented in the code comment.

core/arch/freeRTOS/src/forte_thread.cpp
franz-hoepfinger-4diac: you are right.

core/arch/freeRTOS/src/forte_thread.cpp
franz-hoepfinger-4diac: `join()` is notified via `mJoinSem.inc()` inside `runThread()`, which is called before `vTaskDelay()`. The task then parks itself so that the TCB and stack remain valid until `destroyThread()` calls `vTaskDelete()` after `join()` has returned. `vTaskDelay(portMAX_DELAY);` does handle over Control to FreeRtos here, so the Task itself will be in the 'eBlocked' state.

core/src/ecet.cpp
franz-hoepfinger-4diac: This is not FreeRTOS-specific — it's a general C++ destruction issue. When `end()` is called from `~CThreadBase()`, the vtable has already been reverted to `CThreadBase`, so `onAliveChanged()` is a no-op and `resumeSelfSuspend()` is never reached. The thread stays      
  blocked in `selfSuspend()` and `join()` blocks forever. By calling `end()` explicitly here, while the vtable is still `CEventChainExecutionThread`, `onAliveChanged()` correctly wakes the thread. The subsequent `end()` call in `~CThreadBase() `is a safe no-op because        
  `mThreadHandle` is null after the first call completes.       


Calling e`nd()` in the destructor of a `CThreadBase` subclass is an established pattern in 4diac-forte — nearly all subclasses do it (`CPCTimeHandler`, `CFDSelectHandler`, `CWin32SerComHandler`, etc.). The reason it is needed here specifically is the C++ vtable: if we  rely on `~CThreadBase()` to call `end()`, the vtable has already reverted and `onAliveChanged()` no longer reaches `resumeSelfSuspend()`, causing `join()` to block forever. The subsequent `end()` in `~CThreadBase()` is a safe no-op since `mThreadHandle` is null after the     
  first call.

core/arch/freeRTOS/src/forte_thread.cpp
azoitl: thx for the clarification. I totally forgot what infrastructure I have implemented here by now.

core/arch/freeRTOS/include/forte/arch/forte_thread.h
azoitl: Ok that is bad. As any new pointer is not good.

core/src/ecet.cpp
azoitl: Thx you for this feedback. But this is what I allready understood from the comment, so this is not answering my question.

The reason for my question is that we had an end at the same place and had problems with it. That is also the reason that the normal case is that all event chain execution threads are stopped before they are deleted. If you found a place where this is not the case I would like to know. Because with this change you are maybe masking another problem.

core/arch/common/include/forte/arch/threadbase.h
```diff
@@ -89,6 +89,10 @@ namespace forte::arch {
         return mThreadHandle;
       }
 
+      constexpr long getStackSize() const {
```
azoitl, line 93: why did you make this method public? 
If you add this method I would prefere that `mStackSize` private.

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +97,20 @@ namespace forte::arch {
...
+       * @param paStackSize stack size in bytes
+       * @return stack depth in StackType_t words
+       */
+      constexpr size_t getStackDepth() {
+        return static_cast<size_t>(getStackSize()) / sizeof(StackType_t);
+      }
+
       static const int scmForteTaskPriority;
+
+      StackType_t *mFreeRTOSStack = nullptr;
```
azoitl: After thinking more about the new topic and also considering Scott Meyers advise in his Effective STL book this would be much cleaner as `std:vector`.

azoitl (changes requested):

core/src/ecet.cpp
franz-hoepfinger-4diac: aaaaaaaaaaaaaaah ...OK, yes, it was happening when i had using the RT_E_ Blocks. 
when testing this i wanted to have more Threads to be spawned. 

Yes, i see, https://github.com/eclipse-4diac/4diac-forte/commit/6fb3aa6bdaf43f05bbd8d16b3d1c72da75ec64f1
but noticing, the commit Message does not really match what was done. With this Commit, there was not `end()` removed. 

But: sure you are much deeper in this Topic than me. 
I only can say: without this `end()` i got wrong Object destructions - and crashes - , with the `end()` it runs good. 
Happy to do more testing.

core/arch/common/include/forte/arch/threadbase.h
franz-hoepfinger-4diac: Moved to protected — only subclasses need it. mStackSize was and remains private.

core/src/ecet.cpp
azoitl: Thx for the feedback. I would say this is a different topic and should not be part of this commit.

franz-hoepfinger-4diac: i remove it, and make another PR for this!

franz-hoepfinger-4diac: https://github.com/eclipse-4diac/4diac-forte/pull/877 here you go.

franz-hoepfinger-4diac: and removed from this PR.

core/arch/freeRTOS/include/forte/arch/forte_thread.h
franz-hoepfinger-4diac: done.

core/arch/freeRTOS/include/forte/arch/forte_thread.h
franz-hoepfinger-4diac: Agree. 
do you have a better Idea ? 

i wanted to stay "pure" FreeRTOS. 
another way on ESP32 would be to use C++ Pthread API, but this is a lot adopted by espressif, and not portable anymore.

azoitl (commented):
Sorry to bother you again, but I found one or two issues in code when I did the final review:

 - What should happen if someone is only calling join()?
 - What happens for multiple join() calls?
 - What should if someone calls stop() and no join()?
 - What happens for multiple end() calls?

and in both cases no end()?

franz-hoepfinger-4diac:
> Sorry to bother you again, but I found one or two issues in code when I did the final review:
> 
> * What should happen if someone is only calling join()?
> * What happens for multiple join() calls?
> * What should if someone calls stop() and no join()?
> * What happens for multiple end() calls?
> 
> and in both cases no end()?

  join() only: join() blocks until mJoinSem is signalled by runThread(). At that point mThreadHandle has already been cleared to nullHandle by runThread(). If destroyThread() is never called afterwards, the FreeRTOS task remains parked in vTaskDelay() and
  TCB/stack are leaked. This is a real issue we need to fix.                                                                                                                                                                                                          
                                                            
  Multiple join() calls: Safe — after the first join(), mThreadHandle == nullHandle, so subsequent calls return immediately (no-op).                                                                                                                                  
                                                                                                                                                                                                                                                                      
  stop() without join(): The thread will eventually exit its run() loop and runThread() will clear mThreadHandle. If the destructor runs before that happens, end() correctly calls stop(), join(), and destroyThread(). If runThread() completes first, the          
  destructor's end() sees mThreadHandle == nullHandle and skips destroyThread() — task leaked.                                                                                                                                                                        
                                                                                                                                                                                                                                                                      
  Multiple end() calls: Safe — after the first end(), mThreadHandle == nullHandle, so subsequent calls are no-ops.
                                                   
  No end() at all: The destructor ~CFreeRTOSThread() calls end(). Works correctly as long as runThread() has not already cleared mThreadHandle.                                                                                                                       
  
  Root cause & fix: runThread() clears mThreadHandle before signalling mJoinSem, which can cause destroyThread() to be skipped.

core/arch/common/include/forte/arch/threadbase.h
```diff
@@ -241,7 +259,6 @@ namespace forte::arch {
       paThread->setAlive(true);
       paThread->run();
       paThread->setAlive(false);
-      paThread->mThreadHandle = nullHandle;
```
franz-hoepfinger-4diac, line 244: here the Handle was "nulled" too early.

core/arch/common/include/forte/arch/threadbase.h
```diff
@@ -222,6 +238,8 @@ namespace forte::arch {
     if (nullHandle != mThreadHandle) {
       stop();
       join();
+      destroyThread(mThreadHandle);
+      mThreadHandle = nullHandle;
```
franz-hoepfinger-4diac: now it is nulled AFTER destroyThread so the destructor can destroy safely.

franz-hoepfinger-4diac:
The root cause was that runThread() cleared mThreadHandle before signalling mJoinSem, which meant end() could find a null handle and skip destroyThread(). We fixed this in the base class: mThreadHandle is now only cleared in end() after destroyThread() completes.                                                                                                                                                                                                                                                          
            
  This makes all edge cases correct:                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                      

-   join() only: mThreadHandle remains set after join() returns, so the destructor's end() can still call destroyThread() correctly.                                                                                                                                  
-   stop() without join(): Same — mThreadHandle stays set until end() clears it.                                                                                                                                                                                      
-   Multiple end() calls: After the first end(), mThreadHandle == nullHandle, so subsequent calls are safe no-ops.                                                                                                                                                    
-   No explicit end(): The destructor calls end() and finds mThreadHandle still set — cleanup works correctly.

azoitl:
I don't think this is the way to go. nulling the threadhandle was long thought over and doing it in different places caused many issues. 

But thx to your elaborate explanation I think I now better understand when you need to call your destroythread method:

1. in the freertos thread's destructor after end()
2. in freertos createThread when the thread is restarted

in order that this works you need to cash the threadhandle in the freertos thread but you can implement all your requirements without the need to introduce new api. Which is always the better approach as a learned from the book Practical API Design. Especially for APIs that need to last.

franz-hoepfinger-4diac:
> 1. in the freertos thread's destructor after end()
> 2. in freertos createThread when the thread is restarted

done.

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -20,32 +20,64 @@ namespace forte::arch {
...
 
   CFreeRTOSThread::~CFreeRTOSThread() {
+    end();
+    /* If the task finished naturally before end() was called, runThread() had already
+     * cleared mThreadHandle, so end() skipped the cleanup. Clean up the parked task here. */
+    if (mFreeRTOSThreadHandle != nullptr) {
+      while (eTaskGetState(mFreeRTOSThreadHandle) == eRunning) {
+        vTaskDelay(1);
+      }
+      vTaskDelete(mFreeRTOSThreadHandle);
```
franz-hoepfinger-4diac, line 41: > 1. in the freertos thread's destructor after end()

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -20,32 +20,61 @@ namespace forte::arch {
...
   CFreeRTOSThread::createThread(long paStackSize) {
-    TaskHandle_t handle = 0;
+    if (mFreeRTOSStack.empty() || !mTask) {
+      DEVLOG_ERROR("Error: Could not allocate memory for FreeRTOS Task!\n");
+      return TaskHandle_t(0);
+    }
 
-    if (pdPASS != xTaskCreate(threadFunction, "FORTE", paStackSize, this, scmForteTaskPriority, &handle)) {
+    /* Clean up any previously parked task before creating a new one (restart case). */
+    if (mFreeRTOSThreadHandle != nullptr) {
```
azoitl: why is this duplicated and not a method to be called from here and from the destructor

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +88,27 @@ namespace forte::arch {
...
+      struct FreeRTOSDeleter {
+          void operator()(StaticTask_t *paTask) const {
+            vPortFree(paTask);
+          }
+      };
+
       static const int scmForteTaskPriority;
+
+      std::vector<StackType_t> mFreeRTOSStack;
+      std::unique_ptr<StaticTask_t, FreeRTOSDeleter> mTask;
```
azoitl: can't you use a lamdba?

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -20,32 +20,61 @@ namespace forte::arch {
    */
   const int CFreeRTOSThread::scmForteTaskPriority = tskIDLE_PRIORITY + 4;
 
-  CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize) {
+  CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize), mFreeRTOSStack(getStackDepth()) {
+    /* Allocate the TCB via pvPortMalloc(), not as a member or via new, because
+     * on systems with PSRAM (e.g. ESP32-S3) the CFreeRTOSThread object itself
+     * may reside in PSRAM. FreeRTOS requires the TCB to be in internal SRAM;
+     * pvPortMalloc() guarantees MALLOC_CAP_INTERNAL on ESP-IDF. */
+    mTask.reset(static_cast<StaticTask_t *>(pvPortMalloc(sizeof(StaticTask_t))));
```
azoitl: it would be cleaner to move that to the initializer list and use the constructor of the uniqupointer

azoitl (changes requested):
This is no going in a much better direction. At this point I think it is also sensible to clean the history and bring the commit message up to date.

franz-hoepfinger-4diac:
> This is no going in a much better direction. 

why no ?

i feel it is better now ...

azoitl:
> > This is no going in a much better direction.
> 
> why no ?
> 
> i feel it is better now ...

I meant now

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +88,23 @@ namespace forte::arch {
...
+       *
+       * FreeRTOS expects the stack depth as the number of StackType_t words, not bytes.
+       * @param paStackSize stack size in bytes
+       * @return stack depth in StackType_t words
+       */
+      constexpr size_t getStackDepth() {
+        return static_cast<size_t>(getStackSize()) / sizeof(StackType_t);
+      }
+
+      void deleteFreeRTOSTask();
```
azoitl, line 101: this should definitely be private

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +88,23 @@ namespace forte::arch {
...
+
       static const int scmForteTaskPriority;
+
+      std::vector<StackType_t> mFreeRTOSStack;
```
azoitl, line 105: after looking more detailed the fields should also be for shure not protected

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +88,23 @@ namespace forte::arch {
        */
       virtual TThreadHandleType createThread(long paStackSize);
 
+      /*! \brief Convert a stack size in bytes to a FreeRTOS stack depth in words.
+       *
+       * FreeRTOS expects the stack depth as the number of StackType_t words, not bytes.
+       * @param paStackSize stack size in bytes
+       * @return stack depth in StackType_t words
+       */
+      constexpr size_t getStackDepth() {
```
azoitl, line 97: i guess this one should also be private

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -20,32 +20,64 @@ namespace forte::arch {
...
 
-  CFreeRTOSThread::CFreeRTOSThread(long paStackSize) : CThreadBase(paStackSize) {
+  CFreeRTOSThread::CFreeRTOSThread(long paStackSize) :
+      CThreadBase(paStackSize),
+      mFreeRTOSStack(getStackDepth()),
+      /* Allocate the TCB via pvPortMalloc(), not as a member or via new, because
+       * on systems with PSRAM (e.g. ESP32-S3) the CFreeRTOSThread object itself
+       * may reside in PSRAM. FreeRTOS requires the TCB to be in internal SRAM;
+       * pvPortMalloc() guarantees MALLOC_CAP_INTERNAL on ESP-IDF. */
+      mTask{static_cast<StaticTask_t *>(pvPortMalloc(sizeof(StaticTask_t))), vPortFree} {
```
azoitl: if you have vPortFree here, why do you need in the h file line 106?

azoitl (changes requested):

core/arch/freeRTOS/src/forte_thread.cpp
azoitl: I find it interesting that you keep this deleter and not the one in the header. If you use this one it needs memory. The other, template parameter, can be compile time optimized, as you don not have different instances with different deleters I thought the template version is better. 

Any reason why you opted for the construtor parameter that I missed?

core/arch/freeRTOS/src/forte_thread.cpp
```diff
@@ -20,32 +20,64 @@ namespace forte::arch {
...
+        vTaskDelay(1);
+      }
+      vTaskDelete(mFreeRTOSThreadHandle);
+      mFreeRTOSThreadHandle = nullptr;
+    }
   }
 
   CFreeRTOSThread::~CFreeRTOSThread() {
+    end();
+    /* If the task finished naturally before end() was called, runThread() had already
```
azoitl: this comment is wrong

azoitl (changes requested):

core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -87,7 +88,29 @@ namespace forte::arch {
...
+      struct FreeRTOSDeleter {
+          void operator()(StaticTask_t *paTask) const {
+            vPortFree(paTask);
+          }
+      };
+
       static const int scmForteTaskPriority;
+
+      std::vector<StackType_t> mFreeRTOSStack;
+      std::unique_ptr<StaticTask_t, FreeRTOSDeleter> mTask;
```
azoitl: why now no lambda anymore?

franz-hoepfinger-4diac: like this:

`std::unique_ptr<StaticTask_t, decltype([](StaticTask_t *paTask) { vPortFree(paTask); })> mTask;`
?

diplfranzhoepfinger: now again.

azoitl (approved):

franz-hoepfinger-4diac:
i now get the following Warning: 

```
/home/franz/git/hr/LOGIBUS_integration_datapanel/4diac-forte/core/arch/freeRTOS/include/forte/arch/forte_thread.h:48:9: warning: 'forte::arch::CFreeRTOSThread' has a field 'std::unique_ptr<xSTATIC_TCB, forte::arch::CFreeRTOSThread::<lambda(StaticTask_t*)> > forte::arch::CFreeRTOSThread::mTask' whose type has internal linkage [-Wsubobject-linkage]
   48 |   class CFreeRTOSThread : public CThreadBase<TaskHandle_t> {
      |         ^~~~~~~~~~~~~~~

```


https://github.com/eclipse-4diac/4diac-forte/pull/903
does solve it. 
still all in the .h File, so maximum Compiler Optimisation possible. what do you think ?

PR #869, opened 2026-04-20
add forte::arch::format

antongus:
replace all `std::format_to` calls to `forte::arch::format_to`

depending on FORTE_CUSTOM_FORMAT flag redirect these calls to std::format or to custom format implementation provided by arch.

### DIFF
--- a/core/arch/common/include/forte/arch/forte_format.h
+++ b/core/arch/common/include/forte/arch/forte_format.h
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Insolsoft
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Anton Gusev  - initial API and implementation and/or initial documentation
+ *******************************************************************************/
+
+#pragma once
+
+#ifdef FORTE_CUSTOM_FORMAT
+#include "forte/arch/forte_architecture_format.h"
+#else
+
+#include <format>
+
+namespace forte::arch {
+  template<typename... Args>
+  std::string format(const std::format_string<Args...> fmt, Args &&...args) {
+    return std::format(fmt, std::forward<Args>(args)...);
+  }
+
+  template<typename It, typename... Args>
+  void format_to(It it, const std::format_string<Args...> fmt, Args &&...args) {
+    std::format_to(it, fmt, std::forward<Args>(args)...);
+  }
+} // namespace forte::arch
+
+#endif
--- a/core/arch/CMakeLists.txt
+++ b/core/arch/CMakeLists.txt
@@ -26,6 +26,13 @@ target_compile_options(forte-core PUBLIC
         $<$<COMPILE_LANGUAGE:CXX>:$<$<BOOL:${HAVE_EXCEPTIONS_FLAG}>:$<IF:$<BOOL:${FORTE_RTTI_AND_EXCEPTIONS}>,-fexceptions,-fno-exceptions>>>
 )
 
+option(FORTE_CUSTOM_FORMAT "Use custom format implementation" FALSE)
+#mark_as_advanced(FORTE_CUSTOM_FORMAT)
+if(FORTE_CUSTOM_FORMAT)
+  target_compile_definitions(forte-core PUBLIC "FORTE_CUSTOM_FORMAT")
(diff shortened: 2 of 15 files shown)


core/arch/common/include/forte/arch/forte_format.h
```diff
@@ -0,0 +1,34 @@
...
+namespace forte::arch {
+  template<typename... Args>
+  std::string format(const std::format_string<Args...> fmt, Args &&...args) {
+    return std::format(fmt, std::forward<Args>(args)...);
+  }
+
+  template<typename It, typename... Args>
+  void format_to(It it, const std::format_string<Args...> fmt, Args &&...args) {
+    std::format_to(it, fmt, std::forward<Args>(args)...);
+  }
```
mx990, line 31: While this appears to be very straightforward in simply wrapping `std::format`, you are still relying on `std::format_string` in your abstraction.

How would an independent implementation that does not rely on `std::format` look like then?

mx990 (changes requested):
I am against adding architecture-dependent abstractions for standard C++ facilities. The whole point of having and using standard C++ facilities is to avoid having to deal with differences between individual architectures in the first place.

I am also worried this will end up bloating the arch layer with code that is seldom used and thus poorly tested, threatening the stability of 4diac FORTE on different platforms.

core/arch/common/include/forte/arch/forte_format.h
antongus: > How would an independent implementation that does not rely on std::format look like then?

It can be something like

```c++
  template <typename Iterator, typename ... Args>
  void format_to(Iterator it, const char* fmt, Args&&... args) {
    ...
  }
```
But `std::format_string` does all the checks in compile-time, so I prefer it (if it is available).

antongus:
> The whole point of having and using standard C++ facilities is to avoid having to deal with differences between individual architectures in the first place.

I agree we should ideally just use `std::format` everywhere. However, it bloats the flash footprint by 160KB (for STM32 + FreeRTOS target). A customization point would solve this: unconstrained architectures can use `std::format` through the `forte::arch::format` wrapper, while flash-limited architectures can plug in lightweight formatters.

core/arch/common/include/forte/arch/forte_format.h
mx990: I believe the crux lies in the `...` of your suggestion. The implementation would have to be capable of handling all possible format strings to be a proper substitute for `std::format`. If it handles only specific cases or a subset, it would break as soon as someone used a different format string that has not been anticipated.

In other words, we would be playing catch up with the standard C++ implementation. I do not believe this is a viable approach for such a basic functionality.

mx990:
I understand your point, but please understand mine. Introducing a customization point that constitutes a nightmare from a maintenance standpoint is not a solution. I am therefore closing this.

azoitl:
I totally agree here with @mx990. I had raised my concerns already the first time you brought up the idea. Also I think we should first minimize the usage of format_to as much as possible.

antongus:
> Also I think we should first minimize the usage of format_to as much as possible.

There is no point in reducing the use of `format_to`. Even a single use instantly increases the firmware size by the exact same 160KB. Because of this, I am looking for a way to eliminate it entirely. For integers, dates, and times, we already have replacements like `hexChar()` and `appendInt()`. 
 
For floating-point numbers, we haven't found anything acceptable. `sprintf`/`snprintf` were rejected due to locale dependencies, and `to_chars()` also causes a significant firmware size increase.
 
That's why I decided to propose this approach. I want to emphasize that I didn't mean to add an alternative `format_to` implementation into the `forte` codebase. It was simply intended as a way to customize `forte` during the build process (similar to `forte_fileio.h`, `forte_printer.h`, etc.).
 
 It's a shame this option didn't work out.

azoitl:
@antongus the reason I mentioned to reduce the format_to is that when we have only one or two specific use cases. We can introduce dedicated methods in arch that would then internally use format_to or something similar e.g. to_chars() but not replicate the full C++ API. 

Please also note my comment to your other PR so that we can also merge this one.

antongus:
@azoitl you mean adding to arch something like `forte_ftoa()`? 

> Please also note my comment to your other PR so that we can also merge this one.

Ok, I will update that PR soon.

azoitl:
> @azoitl you mean adding to arch something like `forte_ftoa()`?

Yes I was thinking in that direction. I'm not happy about it but if this is the only way that it be.

mx990:
I have created a PR https://github.com/eclipse-4diac/4diac-forte/pull/871 that adds an `appendFloat` for floating-point numbers. It uses `std::to_chars` by default but can be changed to use `printf` using a new CMake option `FORTE_FORMAT_REAL_PRINTF`.

PR #858, opened 2026-04-16
Add tests for RT_E_REND function blocks

franz-hoepfinger-4diac:
Introduces unit tests for the real-time (RT_E_REND) event rendezvous function blocks. These tests verify proper event synchronization, sequence independence, and reset behavior.

PR #870, opened 2026-04-21
Adds pre-commit code style checks

franz-hoepfinger-4diac:
Introduces pre-commit hooks to enforce code style and quality standards locally, mirroring existing CI checks.

*   Configures various checks, including `clang-format` for C++ files, tab checks, trailing whitespace, end-of-file, YAML syntax, and mixed line endings.
*   Updates the `CONTRIBUTING.md` with detailed instructions on installing and utilizing pre-commit, outlining the configured hooks and their corresponding CI workflows.

This enhancement aims to improve code consistency and catch common issues earlier in the development cycle, reducing CI failures and streamlining developer workflow.


CONTRIBUTING.md
```diff
@@ -48,6 +48,32 @@ To build and develop the 4diac FORTE you need:
 
 A detailed description of setting up a 4diac FORTE development environment and build 4diac FORTE from source can be found in our [development documentation](https://eclipse.dev/4diac/doc/installation/#_4diacforte)
 
+## Code Style and Pre-commit Hooks
+
+This project uses [pre-commit](https://pre-commit.com/) to enforce code style locally before pushing.
```
azoitl: I definitely do not want to push pre-comit hooks on all of our contributors. With modern editors and the clang-format file you can easily achieve the same result. This shall be accommodated in the docs. also this section should be a subsection of the previous one.

azoitl (changes requested):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
+# https://pre-commit.com/
+# Install: pip install pre-commit && pre-commit install
+# Run manually: pre-commit run --all-files
+
+repos:
+  # General code quality checks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
```
mx990: I believe the latest tag is `v6.0.0`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+        name: Check for Tabs
+        entry: tests/findTabs.sh
+        language: script
+        pass_filenames: false
+        always_run: true
+
+  # clang-format — identical to .github/workflows/cpp-linter.yml
+  # version: 20, style: file (.clang-format), tidy: disabled
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v20.1.0
```
mx990: I believe the latest tag is `v22.1.3`.

CONTRIBUTING.md
mx990: I also believe we should phrase it as an offer from the start. I would suggest the following:
```
We provide Git [pre-commit](https://pre-commit.com/) hooks that can automatically check your code style locally in order to avoid CI failures.
```

mx990 (changes requested):

mx990 (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+  # Check for Tabs — identical to .github/workflows/tab_check.yml (tests/findTabs.sh)
+  - repo: local
+    hooks:
+      - id: check-tabs
+        name: Check for Tabs
+        entry: tests/findTabs.sh
+        language: script
+        pass_filenames: false
+        always_run: true
```
MartinMelikMerkumians: arent those handled by clang-format anyway? @azoitl @mx990 
if not, could this be added to the clang-format config?

CONTRIBUTING.md
```diff
@@ -48,6 +48,33 @@ To build and develop the 4diac FORTE you need:
...
+
+We provide Git [pre-commit](https://pre-commit.com/) hooks that can automatically check your code style locally in order to avoid CI failures.
+The hooks mirror the CI checks in `.github/workflows/`.
+
+**Install pre-commit:**
+```bash
+pip install pre-commit
+pre-commit install
+```
+The Installation of the pre-commit hooks is **optional** for any Contributors.
```
MartinMelikMerkumians: installation starts with a lower-case "i"
constributors starts with a lower-case "c"

.pre-commit-config.yaml
gravemalte: Yes you can configure clang-format with this option [^1].
And I don't think we did it already.

[^1]: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#usetab

azoitl: @MartinMelikMerkumians thx for bringing this up and @gravemalte thx for the clarification. 

For me it is important to have a single source of truth otherwise it will be a maintenance nightmare. As the clang-format file is the better place for that we should use it for this.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+# Run manually: pre-commit run --all-files
+
+repos:
+  # General code quality checks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v6.0.0
+    hooks:
+      - id: trailing-whitespace
+        exclude: \.patch$
+      - id: end-of-file-fixer
```
gravemalte: I'm unsure if we should include this config. Running your configuration local gives me a lot of changes.

CONTRIBUTING.md
azoitl: This still only talks about the pre-commit hooks. As written below the main source of truth and for me the better solution is the clang-format file. Therefore this should be mentioned first. Furthermore, we have a full documentation file on code style which is also not mentioned. Therefore the new headline is totally misleading. Can you please fix that.

.pre-commit-config.yaml
mx990: I also believe we should avoid redundant checks. I would then suggest to also let `clang-format` handle `modules`, which is currently explicitly ignored.

> Yes you can configure clang-format with this option. And I don't think we did it already.

It is already set to `None` by default, but I am open to making this explicit in our `.clang-format` file.

.pre-commit-config.yaml
diplfranzhoepfinger: Clear. If you let run over all files. But I thought it only changes Files at commit; unless you start it on command line and let it run over all Files. 

Isn't same true for the clang Formatting thing?

CONTRIBUTING.md
franz-hoepfinger-4diac: thanks, fixed.

.pre-commit-config.yaml
MartinMelikMerkumians: I would prefer having all formatting in clang-format

.pre-commit-config.yaml
gravemalte: `clang-format` just touches C++ related files while this hook extension touches every file.

@mx990, @MartinMelikMerkumians @azoitl What do you say?

.pre-commit-config.yaml
franz-hoepfinger-4diac: trailing-whitespace is partially covered by clang-format for C++ files, but this hook also applies to non-C++ files (YAML, CMake, shell scripts). The same applies to end-of-file-fixer and mixed-line-ending. check-yaml, check-added-large-files, and check-merge-conflict have no clang-format equivalent at all.

Should we remove the hooks that overlap with clang-format and only keep the ones with no equivalent?

CONTRIBUTING.md
```diff
@@ -48,6 +48,35 @@ To build and develop the 4diac FORTE you need:
 
 A detailed description of setting up a 4diac FORTE development environment and build 4diac FORTE from source can be found in our [development documentation](https://eclipse.dev/4diac/doc/installation/#_4diacforte)
 
+### Code Style
+
+The coding guidelines are documented in the [4diac FORTE Code Quality](https://eclipse.dev/4diac/doc/development/forte_codequality.html) page.
```
azoitl: This is the wrong link. It should point to the coding rules in the doc directory of this repo

azoitl (changes requested):
Following the discussion I think for a first PR a pre-commit hook should check only one thing and for me this would be the clangd style on C and C++ files. 

Also I'm feeling very uncomfortable that this pre-commit fill pulls in some repos with versions. From a security perspective with all the supplychain attacks that happened recently this makes me nervous. 

As I have no clue how git pre-commit hooks work I don't know if there are better solutions but at least pinning versions with hashes should be done. But maybe I'm to over sensitive here.

.pre-commit-config.yaml
azoitl: As I worked on other stuff the last days I noticed that I definitely don't want to have these changes on all files. For example some markdown files use trailing whitespaces for formatting. Some other files may require tabs.

CONTRIBUTING.md
MartinMelikMerkumians: Perhaps we then just start with the clang-format and skip the rest, and add stuff one step at a time? I think formatting h/cpp files according to the clang-format should be a common accepted standard, right?

franz-hoepfinger-4diac: done with https://github.com/eclipse-4diac/4diac-forte/pull/870/commits/6f3854d1189efb0f0b00af750ffaabb99ed7c012

azoitl:

PR #877, opened 2026-04-23
Fix potential deadlock in CEventChainExecutionThread destructor

franz-hoepfinger-4diac:
Call end() explicitly in the destructor while the vtable is still
CEventChainExecutionThread's, ensuring onAliveChanged() dispatches to
resumeSelfSuspend() and wakes the thread from selfSuspend(). Without
this, ~CThreadBase() calls end() with the base-class vtable, making
onAliveChanged() a no-op and causing join() to block forever.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/core/src/ecet.cpp
+++ b/core/src/ecet.cpp
@@ -30,7 +30,16 @@ namespace forte {
     clear();
   }
 
-  CEventChainExecutionThread::~CEventChainExecutionThread() = default;
+  CEventChainExecutionThread::~CEventChainExecutionThread() {
+    /* Call end() while the vtable is still CEventChainExecutionThread's, so that
+     * onAliveChanged() dispatches to our resumeSelfSuspend() and wakes the thread
+     * from selfSuspend(). Without this, ~CThreadBase() would call end() with the
+     * base-class vtable, onAliveChanged() would be a no-op, and join() would block
+     * forever. This applies to all platforms, not just FreeRTOS.
+     * The subsequent end() call in ~CThreadBase() is a safe no-op because
+     * mThreadHandle is null after the first end() completes. */
+    end();
+  }
 
   void CEventChainExecutionThread::run() {
     while (isAlive()) { // thread is allowed to execute


azoitl:
you where for me a bit to quick with the PR. I expected to first to find the root cause that got you the deadlock in the first place so that we can see if this is the right place to fix the problem. As you know where it happend it would be great if you could further analyse and investigate.

franz-hoepfinger-4diac:
> you where for me a bit to quick with the PR. I expected to first to find the root cause that got you the deadlock in the first place so that we can see if this is the right place to fix the problem. As you know where it happend it would be great if you could further analyse and investigate.

Unterstand. 
I will try to get stack-Traces later today. 
But i also want to avoid, the PR https://github.com/eclipse-4diac/4diac-forte/pull/863 will be merged, and this thing forgotten.

azoitl:
But this stack trace is now a clear problem of your #863 PR because you get the exception in `deleteFreeRTOSTask`.

franz-hoepfinger-4diac:
> But this stack trace is now a clear problem of your #863 PR because you get the exception in `deleteFreeRTOSTask`.

YOU ARE RIGHT. 
because adding the `end()` did also not fix it. 
also i found out i have dirty merges, the Code in the PR was not the code i was running! i deleted the Trace. 
`destroyThread` was the Reason. 
i deleted, so the Code is same as #863 now it runs even without the `end()` at least when there is no RT Block. 
i now test with RT Block.

azoitl:
So #863 works as it should?

franz-hoepfinger-4diac:
> So #863 works as it should?

true, but i want to make some RT_E_ tests now to fully confirm.

azoitl:
> > So #863 works as it should?
> 
> true, but i want to make some RT_E_ tests now to fully confirm.

thx for the clarification. Could not fully follow your initial comment that you gave when closing this PR.

PR #882, opened 2026-04-27
fix: CGatheringDataConnection::disconnect does not null destination DI slot

Klaurx:
When a DeleteConnection command targets a gathering connection using a port-only path, CGatheringDataConnection::disconnect returns without clearing the destination FB data input slot. 
The caller in CResource::deleteConnection then frees the object, leaving the slot with a stale pointer.

Root cause: 
connectDI(port, nullptr) was never called in CGatheringDataConnection::disconnect, unlike CDataConnection::disconnect which upholds this correctly.

Fix: 
Call connectDI(dstPortId, nullptr) before returning, consistent with how other connection types handle cleanup.

```
EMGMResponse CGatheringDataConnection::disconnect(CFunctionBlock &paDstFB,
                                                   std::span<const StringId> paDstPortNameId) {
  for (const auto [member, connection, memberName] : mGatheringData) {
    connection->disconnect(paDstFB, memberName);
    if (connection->isDelegating()) {
      delete connection;
    }
  }
  mGatheringData.clear();

  const TPortId dstPortId =
      paDstFB.getFBInterfaceSpec().getDIID(paDstPortNameId.front());
  paDstFB.connectDI(dstPortId, nullptr);

  return EMGMResponse::Ready;
}
```

PR #883, opened 2026-04-27
fix: CGatheringDataConnection::disconnect does not null destination DI slot

Klaurx:
When a DeleteConnection command targets a gathering connection using a
port-only path, CGatheringDataConnection::disconnect returned without
clearing the destination FB data input slot. The caller in
CResource::deleteConnection then freed the object, leaving the slot
with a stale pointer.

Fix: Call connectDI(dstPortId, nullptr) before returning, consistent
with how CDataConnection::disconnect handles cleanup.

### DIFF
--- a/core/src/gatherdataconn.cpp
+++ b/core/src/gatherdataconn.cpp
@@ -49,6 +49,8 @@ namespace forte::internal {
     }
     mGatheringData.clear();
 
+    paDstFB.connectDI(dstPortId, nullptr);
+
     return EMGMResponse::Ready;
   }
 


Klaurx:
@mx990 no rush, i was just wondering when it will be merged, and if there is anything else i should do.

core/src/gatherdataconn.cpp
```diff
@@ -47,10 +47,11 @@ namespace forte::internal {
         delete connection;
       }
     }
-    mGatheringData.clear();
+mGatheringData.clear();
```
mx990: It looks like you may have accidentially messed up the formatting. Please restore this line.

core/src/gatherdataconn.cpp
```diff
@@ -47,10 +47,11 @@ namespace forte::internal {
...
+
+    paDstFB.connectDI(dstPortId, nullptr);
 
     return EMGMResponse::Ready;
-  }
```
mx990, line 53: It looks like you may have accidentially deleted this line. Please restore it the way it was.

mx990 (changes requested):
As mentioned in https://github.com/eclipse-4diac/4diac-forte/pull/882#issuecomment-4330040937, please use a more descriptive commit message. The title and description of this PR would be a good example.

You can simply amend your commit and then force push your changes. There is no need to create another PR. If you are unsure or need assistance, please let me know.

Klaurx:
Hi @mx990, Ive squashed everything into a single clean commit with the correct message and formatting. The duplicate line has been removed and the closing brace is restored. Should be ready for review now. 
Please let me know if there are any mistakes or misunderstandings on my side.

mx990 (approved):
Looks good. Thank you again for contributing to Eclipse 4diac.

Klaurx:
Thank you very much for the review and for the guidance throughout the process.

I really appreciate the time and patience, especially with the branch cleanup and commit corrections on my side.

I am glad the fix could be merged, and I appreciate the opportunity to contribute to Eclipse 4diac.

Please let me know if anything is needed.

PR #884, opened 2026-04-28
Display the actual port being used in BSD and Win32 Socket Interface

m-meingast:
Also fixes #831, where connection information was not shown for log levels below `FORTE_LOGINFO`

### DIFF
--- a/core/arch/common/include/forte/arch/bsdsocketinterf.h
+++ b/core/arch/common/include/forte/arch/bsdsocketinterf.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2023 ACIN, fortiss GmbH, OFFIS e.V.
+ * Copyright (c) 2010 ACIN, fortiss GmbH, OFFIS e.V.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
--- a/core/arch/common/src/bsdsocketinterf.cpp
+++ b/core/arch/common/src/bsdsocketinterf.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2026 ACIN, Profactor GmbH, AIT, fortiss GmbH, OFFIS e.V.
+ * Copyright (c) 2010 ACIN, Profactor GmbH, AIT, fortiss GmbH, OFFIS e.V.
  *                          Primetals Technologies Austria GmbH
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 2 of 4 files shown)


core/arch/common/src/bsdsocketinterf.cpp
```diff
@@ -48,17 +51,24 @@ namespace forte::arch {
...
-      if (0 == bind(nSocket, (struct sockaddr *) &stSockAddr, sizeof(struct sockaddr))) {
-        if (-1 == listen(nSocket, 1)) { // for the classic IEC 61499 server only one connection at the same time is
+      if (bind(nSocket, (struct sockaddr *) &stSockAddr, sizeof(struct sockaddr)) == 0) {
+        if (listen(nSocket, 1) == -1) { // for the classic IEC 61499 server only one connection at the same time is
           // accepted TODO mayb make this adjustable for future extensions
           DEVLOG_ERROR("CBSDSocketInterface: listen() failed: %s\n", strerror(errno));
         } else {
           nRetVal = nSocket;
         }
+        socklen_t addressLen = sizeof(stSockAddr);
```
azoitl: this should only be done if the given port was 0

also please extract into a method.

core/arch/common/src/bsdsocketinterf.cpp
```diff
@@ -37,7 +40,7 @@ namespace forte::arch {
     DEVLOG_INFO("CBSDSocketInterface: Opening TCP-Server connection at: %s:%d\n", paIPAddr, paPort);
```
azoitl, line 51: as we may have a wrong port number here this should now be moved to the place when we have the right portnumber.

azoitl (changes requested):
The similar problems should also be fixed for windows

core/arch/common/src/bsdsocketinterf.cpp
m-meingast: Done

core/arch/common/src/bsdsocketinterf.cpp
m-meingast: As discussed, this message will still be displayed and for `port = 0` an additional message is logged with the actual port being used.

core/arch/common/include/forte/arch/bsdsocketinterf.h
```diff
@@ -42,5 +42,6 @@ namespace forte::arch {
 
     private:
       static int handleError(int nRetVal, const char *msg);
+      static void displayActualPort(TSocketDescriptor paSocket, sockaddr_in &paSockAddr);
```
azoitl: I would prefer to have this as function in an anonymouse namespace in the cpp only.

core/arch/win32/include/forte/arch/win32socketinterf.h
```diff
@@ -40,5 +40,6 @@ namespace forte::arch {
       CWin32SocketInterface(); // this function is not implemented as we don't want instances of this class
 
       static LPSTR getErrorMessage(int paErrorNumber);
+      static void displayActualPort(TSocketDescriptor paSocket, sockaddr_in &paSockAddr);
```
azoitl: I would prefer to have this as function in an anonymouse namespace in the cpp only.

azoitl (changes requested):

core/arch/common/src/bsdsocketinterf.cpp
```diff
@@ -12,13 +13,28 @@
...
+ *     - add logging of actual tcp port being used
  *******************************************************************************/
 #include "forte/arch/sockhand.h" //needs to be first pulls in the platform specific includes
 #include "forte/arch/bsdsocketinterf.h"
 #include "forte/util/devlog.h"
 #include <string.h>
 
 namespace forte::arch {
+  namespace {
+    void displayActualPort(CBSDSocketInterface::TSocketDescriptor paSocket, const sockaddr_in &paSockAddr) {
```
azoitl: remove the paSockAddr paramater it is not needed.

core/arch/win32/src/win32socketinterf.cpp
```diff
@@ -21,6 +22,29 @@
...
 namespace forte::arch {
+  namespace {
+    LPSTR getErrorMessage(int paErrorNumber) {
+      LPSTR pacErrorMessage = nullptr;
+      FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                    nullptr, paErrorNumber, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), pacErrorMessage, 0, nullptr);
+      return pacErrorMessage;
+    }
+
+    void displayActualPort(CWin32SocketInterface::TSocketDescriptor paSocket, const sockaddr_in &paSockAddr) {
```
azoitl: remove the paSockAddr paramater it is not needed.

core/arch/win32/src/win32socketinterf.cpp
```diff
@@ -58,45 +58,46 @@ namespace forte::arch {
     DEVLOG_INFO("CWin32SocketInterface: Opening TCP-Server connection at: %s:%d\n", paIPAddr, paPort);
 
     TSocketDescriptor nSocket = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    if (nSocket == INVALID_SOCKET) {
+      int nLastError = WSAGetLastError();
```
azoitl, line 62: should we extract the error message logging into an own method to remove the duplicated WSAGetLastError, getErrorMessage and LocalFree duplication?

azoitl (changes requested):

azoitl (approved):

PR #886, opened 2026-04-29
Add isComparable check for CIEC_ANY_ELEMENTARY_VARIANT in comparison FBs

franz-hoepfinger-4diac:
- Add isComparable() method to CIEC_ANY_ELEMENTARY_VARIANT to check type compatibility
- Update F_EQ, F_NE, F_LT, F_LE, F_GT, F_GE FBs to use isComparable() before comparison
- Log error with FB instance name only when types are incompatible (deferred getFullQualifiedApplicationInstanceName)
- Avoid duplicate error messages by skipping compare() call when types are incompatible

### DIFF
--- a/core/include/forte/datatypes/forte_any_elementary_variant.h
+++ b/core/include/forte/datatypes/forte_any_elementary_variant.h
@@ -127,6 +127,8 @@ namespace forte {
 
       [[nodiscard]] const CIEC_ANY_ELEMENTARY &unwrap() const override;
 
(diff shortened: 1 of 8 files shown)


azoitl (commented):
I don't think you want this change. This is a performance nightmare.

franz-hoepfinger-4diac:
no i dont. 
this will replace it: 
https://github.com/eclipse-4diac/4diac-forte/pull/889

Issue #888, opened 2026-04-29
CIEC_ANY_ELEMENTARY_VARIANT::compare() returns ambiguous -1 for both error and valid less-than result


franz-hoepfinger-4diac:
CIEC_ANY_ELEMENTARY_VARIANT::compare() uses -1 as sentinel for incompatible types (error), but -1 is also a valid return value for legitimate comparisons:                               

  - Numeric types: (primitiveValue > primitiveOther) - (primitiveValue < primitiveOther) returns exactly -1 when left < right
  - CIEC_STRING: std::string::compare() can return -1
  - CIEC_WSTRING: strcmp() can return -1                                                                                                                                                                                                               
            
  This means callers cannot distinguish between "types are incompatible" and "left operand is less than right operand".
                                                                                                                                                                                                                                                       
  Impact:
  - operator!= currently returns true for incompatible types (since -1 != 0)                                                                                                                                                                           
  - operator< and operator<= currently return true for incompatible types (since -1 < 0)                                                                                                                                                             
  - Any future code that tries to detect the error case via the return value of compare() will produce incorrect results for string comparisons


possible Solution: 
To be discussed. 


the nicest way would maybe be: 

```
#include <expected>

std::expected<double, ErrorCode> divide(double a, double b) {
    if (b == 0) return std::unexpected(ErrorCode::InvalidInput);
    return a / b;
}
```
this in my knowledge is only possible in C++23. 

the other way would be to use `std::partial_ordering`

```
return std::partial_ordering::unordered;                                                                                                                                                                                             
return std::partial_ordering::equivalent;                                                                                                                                                                                            
return std::partial_ordering::less;                                                                                                                                                                                                  
return std::partial_ordering::greater;   
```

would be very clean in my eyes.

franz-hoepfinger-4diac:
https://github.com/eclipse-4diac/4diac-forte/pull/889 i guess returning `return std::partial_ordering::unordered` instead of `-1` for non-compatible Types makes perfectly sense.

PR #889, opened 2026-04-29
Use std::partial_ordering in CIEC_ANY_ELEMENTARY_VARIANT::compare and F_EQ

franz-hoepfinger-4diac:
compare() now returns std::partial_ordering instead of int, using
unordered for incompatible types to avoid the ambiguity with -1 as both
a valid less-than result and an error sentinel. F_EQ detects unordered
and logs the FB instance name alongside the type mismatch.

https://github.com/eclipse-4diac/4diac-forte/issues/888

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/core/include/forte/datatypes/forte_any_elementary_variant.h
+++ b/core/include/forte/datatypes/forte_any_elementary_variant.h
@@ -14,6 +14,7 @@
  *******************************************************************************/
 #pragma once
 
+#include <compare>
 #include <variant>
 
 #include "forte/datatypes/forte_any.h"
@@ -135,33 +136,17 @@ namespace forte {
         return unwrap().equals(paOther.unwrap());
       }
 
-      [[nodiscard]] static int compare(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
-                                       const CIEC_ANY_ELEMENTARY_VARIANT &paOther);
+      [[nodiscard]] static std::partial_ordering compare(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
+                                                         const CIEC_ANY_ELEMENTARY_VARIANT &paOther);
   };
 
-  inline bool operator==(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) == 0;
-  }
+  // operator<=> replaces <, <=, >, >=
+  // operator== is still needed explicitly (not automatically generated from <=>)
+  // operator!= is automatically generated from operator==
+  std::partial_ordering operator<=>(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
+                                    const CIEC_ANY_ELEMENTARY_VARIANT &paOther);
 
-  inline bool operator!=(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) != 0;
-  }
-
-  inline bool operator<(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) < 0;
-  }
-
-  inline bool operator<=(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) <= 0;
-  }
-
-  inline bool operator>(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) > 0;
-  }
-
-  inline bool operator>=(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) >= 0;
-  }
+  bool operator==(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther);
 
   static_assert(std::is_copy_constructible_v<CIEC_ANY_ELEMENTARY_VARIANT>);
   static_assert(std::is_move_constructible_v<CIEC_ANY_ELEMENTARY_VARIANT>);
--- a/core/src/datatypes/forte_any_elementary_variant.cpp
+++ b/core/src/datatypes/forte_any_elementary_variant.cpp
@@ -150,10 +150,10 @@ namespace forte {
     }
   }
 
-  int CIEC_ANY_ELEMENTARY_VARIANT::compare(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
-                                           const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
+  std::partial_ordering CIEC_ANY_ELEMENTARY_VARIANT::compare(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
+                                                             const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
     return std::visit(
-        [](auto &&value, auto &&other) -> int {
+        [](auto &&value, auto &&other) -> std::partial_ordering {
           using T = std::decay_t<decltype(value)>;
           using U = std::decay_t<decltype(other)>;
           using commonType = std::conditional_t<std::is_same_v<T, U>, T, typename mpl::get_castable_type<T, U>::type>;
@@ -161,22 +161,29 @@ namespace forte {
             if constexpr (std::is_same_v<T, U> && std::is_same_v<T, CIEC_STRING>) {
               return static_cast<CIEC_STRING>(value).compare(static_cast<CIEC_STRING>(other));
             } else if constexpr (std::is_same_v<T, U> && std::is_same_v<T, CIEC_WSTRING>) {
-              return strcmp(static_cast<commonType>(value).getValue(), static_cast<commonType>(other).getValue());
+              const auto wstringValue = static_cast<commonType>(value).getValue();
+              const auto wstringOther = static_cast<commonType>(other).getValue();
+              return wstringValue <=> wstringOther;
             } else {
-              DEVLOG_ERROR("Comparing incompatible types %s and %s\n", value.getTypeNameID().data(),
-                           other.getTypeNameID().data());
-              return -1;
+              return std::partial_ordering::unordered;
             }
           } else if constexpr (!std::is_same_v<commonType, mpl::NullType>) {
-            auto primitiveValue = static_cast<typename commonType::TValueType>(static_cast<commonType>(value));
-            auto primitiveOther = static_cast<typename commonType::TValueType>(static_cast<commonType>(other));
-            return (primitiveValue > primitiveOther) - (primitiveValue < primitiveOther);
+            const auto primitiveValue = static_cast<typename commonType::TValueType>(static_cast<commonType>(value));
+            const auto primitiveOther = static_cast<typename commonType::TValueType>(static_cast<commonType>(other));
+            return primitiveValue <=> primitiveOther;
           } else {
-            DEVLOG_ERROR("Comparing incompatible types %s and %s\n", value.getTypeNameID().data(),
-                         other.getTypeNameID().data());
-            return -1;
+            return std::partial_ordering::unordered;
           }
         },
         static_cast<const variant &>(paValue), static_cast<const variant &>(paOther));
+  }
+
+  std::partial_ordering operator<=>(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
+                                    const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
+    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther);
+  }
+
+  bool operator==(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
+    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) == std::partial_ordering::equivalent;
   }
 } // namespace forte
--- a/tests/core/datatypes/CIEC_ANY_ELEMENTARY_VARIANT_test.cpp
+++ b/tests/core/datatypes/CIEC_ANY_ELEMENTARY_VARIANT_test.cpp
@@ -133,6 +133,12 @@ namespace forte::test {
     BOOST_CHECK(!(test1 < test2));
     BOOST_CHECK(test1 >= test2);
     BOOST_CHECK(!(test1 > test2));
+    BOOST_CHECK(std::is_eq(test1 <=> test2));
+    BOOST_CHECK(!std::is_neq(test1 <=> test2));
+    BOOST_CHECK(!std::is_lt(test1 <=> test2));
+    BOOST_CHECK(std::is_lteq(test1 <=> test2));
+    BOOST_CHECK(!std::is_gt(test1 <=> test2));
+    BOOST_CHECK(std::is_gteq(test1 <=> test2));
 
     // same types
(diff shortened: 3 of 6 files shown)


core/src/datatypes/forte_any_elementary_variant.cpp
```diff
@@ -150,31 +150,40 @@ namespace forte {
...
             if constexpr (std::is_same_v<T, U> && std::is_same_v<T, CIEC_STRING>) {
-              return static_cast<CIEC_STRING>(value).compare(static_cast<CIEC_STRING>(other));
+              const int cmp = static_cast<CIEC_STRING>(value).compare(static_cast<CIEC_STRING>(other));
+              return cmp < 0   ? std::partial_ordering::less
+                     : cmp > 0 ? std::partial_ordering::greater
+                               : std::partial_ordering::equivalent;
             } else if constexpr (std::is_same_v<T, U> && std::is_same_v<T, CIEC_WSTRING>) {
-              return strcmp(static_cast<commonType>(value).getValue(), static_cast<commonType>(other).getValue());
+              const int cmp =
+                  strcmp(static_cast<commonType>(value).getValue(), static_cast<commonType>(other).getValue());
```
franz-hoepfinger-4diac: this can be even improved, as strcmp is "old" and there is a Version wich returns `std::strong_ordering` which is implicitly convertable to `std::partial_ordering` i will do soon.

core/include/forte/datatypes/forte_any_elementary_variant.h
```diff
@@ -135,16 +136,17 @@ namespace forte {
...
 
-      [[nodiscard]] static int compare(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
-                                       const CIEC_ANY_ELEMENTARY_VARIANT &paOther);
+      [[nodiscard]] static std::partial_ordering compare(const CIEC_ANY_ELEMENTARY_VARIANT &paValue,
+                                                         const CIEC_ANY_ELEMENTARY_VARIANT &paOther);
   };
 
   inline bool operator==(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) == 0;
+    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) == std::partial_ordering::equivalent;
```
mx990: This should anyways use `CIEC_ANY_ELEMENTARY_VARIANT::equals` instead.

core/include/forte/datatypes/forte_any_elementary_variant.h
```diff
@@ -135,16 +136,17 @@ namespace forte {
...
   }
 
   inline bool operator!=(const CIEC_ANY_ELEMENTARY_VARIANT &paValue, const CIEC_ANY_ELEMENTARY_VARIANT &paOther) {
-    return CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther) != 0;
+    const auto result = CIEC_ANY_ELEMENTARY_VARIANT::compare(paValue, paOther);
+    return result == std::partial_ordering::less || result == std::partial_ordering::greater;
   }
```
mx990: With C++20, this can now be automatically synthesized by the compiler and should thus be removed.

core/include/forte/datatypes/forte_any_elementary_variant.h
mx990: With C++20, we now have the [three-way comparison](https://en.cppreference.com/cpp/language/operator_comparison#Three-way_comparison) `operator<=>` that can return a `std::partial_ordering` directly. The two way comparison operators can then be automatically synthesized by the compiler and should thus be removed.

core/src/datatypes/forte_any_elementary_variant.cpp
```diff
@@ -150,31 +150,40 @@ namespace forte {
...
     return std::visit(
-        [](auto &&value, auto &&other) -> int {
+        [](auto &&value, auto &&other) -> std::partial_ordering {
           using T = std::decay_t<decltype(value)>;
           using U = std::decay_t<decltype(other)>;
           using commonType = std::conditional_t<std::is_same_v<T, U>, T, typename mpl::get_castable_type<T, U>::type>;
           if constexpr (std::is_base_of_v<CIEC_ANY_STRING, commonType>) {
             if constexpr (std::is_same_v<T, U> && std::is_same_v<T, CIEC_STRING>) {
-              return static_cast<CIEC_STRING>(value).compare(static_cast<CIEC_STRING>(other));
+              const int cmp = static_cast<CIEC_STRING>(value).compare(static_cast<CIEC_STRING>(other));
```
mx990: We should add an `operator<=>` in `CIEC_STRING` that returns a `std::strong_ordering`.

core/src/datatypes/forte_any_elementary_variant.cpp
```diff
@@ -150,31 +150,40 @@ namespace forte {
...
+              return cmp < 0   ? std::partial_ordering::less
+                     : cmp > 0 ? std::partial_ordering::greater
+                               : std::partial_ordering::equivalent;
             } else if constexpr (std::is_same_v<T, U> && std::is_same_v<T, CIEC_WSTRING>) {
-              return strcmp(static_cast<commonType>(value).getValue(), static_cast<commonType>(other).getValue());
+              const int cmp =
+                  strcmp(static_cast<commonType>(value).getValue(), static_cast<commonType>(other).getValue());
+              return cmp < 0   ? std::partial_ordering::less
+                     : cmp > 0 ? std::partial_ordering::greater
+                               : std::partial_ordering::equivalent;
```
mx990: We should add an `operator<=>` in `CIEC_WSTRING` that returns a `std::strong_ordering`.

core/src/datatypes/forte_any_elementary_variant.cpp
```diff
@@ -150,31 +150,40 @@ namespace forte {
...
             } else {
-              DEVLOG_ERROR("Comparing incompatible types %s and %s\n", value.getTypeNameID().data(),
+              DEVLOG_ERROR("Comparing incompatible types %s and %s", value.getTypeNameID().data(),
                            other.getTypeNameID().data());
```
mx990: Now that we can indicate an incomparable result, we should remove the error message.

core/src/datatypes/forte_any_elementary_variant.cpp
```diff
@@ -150,31 +150,40 @@ namespace forte {
...
-            return (primitiveValue > primitiveOther) - (primitiveValue < primitiveOther);
+            const auto primitiveValue = static_cast<typename commonType::TValueType>(static_cast<commonType>(value));
+            const auto primitiveOther = static_cast<typename commonType::TValueType>(static_cast<commonType>(other));
+            return primitiveValue < primitiveOther   ? std::partial_ordering::less
+                   : primitiveValue > primitiveOther ? std::partial_ordering::greater
+                                                     : std::partial_ordering::equivalent;
           } else {
-            DEVLOG_ERROR("Comparing incompatible types %s and %s\n", value.getTypeNameID().data(),
+            DEVLOG_ERROR("Comparing incompatible types %s and %s", value.getTypeNameID().data(),
                          other.getTypeNameID().data());
```
mx990: Now that we can indicate an incomparable result, we should remove the error message.

core/src/datatypes/forte_any_elementary_variant.cpp
```diff
@@ -150,31 +150,40 @@ namespace forte {
...
             }
           } else if constexpr (!std::is_same_v<commonType, mpl::NullType>) {
-            auto primitiveValue = static_cast<typename commonType::TValueType>(static_cast<commonType>(value));
-            auto primitiveOther = static_cast<typename commonType::TValueType>(static_cast<commonType>(other));
-            return (primitiveValue > primitiveOther) - (primitiveValue < primitiveOther);
+            const auto primitiveValue = static_cast<typename commonType::TValueType>(static_cast<commonType>(value));
+            const auto primitiveOther = static_cast<typename commonType::TValueType>(static_cast<commonType>(other));
+            return primitiveValue < primitiveOther   ? std::partial_ordering::less
+                   : primitiveValue > primitiveOther ? std::partial_ordering::greater
+                                                     : std::partial_ordering::equivalent;
```
mx990: Please use `<=>` instead to perform a three-way comparsion directly.

mx990 (changes requested):
I like the idea of using the new C++ three-way comparison to better indicate incomparable results and also reduce implementation overhead. However, we should make sure to consistently apply these changes to all data types. I also believe we should split changes in the data types into a separate PR from adapting the FB types.

core/include/forte/datatypes/forte_any_elementary_variant.h
franz-hoepfinger-4diac: i already did before reading your comment, thanks for it anyhow.

core/src/datatypes/forte_any_elementary_variant.cpp
franz-hoepfinger-4diac: true. and `std::strong_ordering` implicitly casts to `std::partial_ordering`

core/src/datatypes/forte_any_elementary_variant.cpp
franz-hoepfinger-4diac: how does the Customer then know what Error he has done ? for me this Error Message is the worthiest think in the whole Setup, (together with the FB Name)

core/include/forte/datatypes/forte_any_elementary_variant.h
franz-hoepfinger-4diac: NOW `::equals` and `::compare` are different. the first make a Strong Compare (only if Type and Value are same it is same) but the second allow compatible types to be compared. 
this has to be discussed/directed by the boss first before adopting!

tests/core/datatypes/CIEC_ANY_ELEMENTARY_VARIANT_test.cpp
```diff
@@ -215,13 +234,16 @@ namespace forte::test {
...
+    // Previously these asserted test1 < test2 / test1 <= test2 etc.
+    // That was wrong: incomparable types (e.g. DINT vs STRING) are not ordered.
+    // They are simply unequal (unordered in std::partial_ordering terms).
     test1 = CIEC_DINT(17);
     test2 = "abc"_STRING;
 
     BOOST_CHECK(!(test1 == test2));
     BOOST_CHECK(test1 != test2);
-    BOOST_CHECK(test1 <= test2);
-    BOOST_CHECK(test1 < test2);
```
franz-hoepfinger-4diac, line 224: these 2 lines in my Opinion was a wrong test before.

tests/core/datatypes/CIEC_ANY_ELEMENTARY_VARIANT_test.cpp
```diff
@@ -230,11 +252,34 @@ namespace forte::test {
...
```
franz-hoepfinger-4diac, line 234: these 2 lines in my Opinion was a wrong test before.

tests/core/datatypes/CIEC_ANY_ELEMENTARY_VARIANT_test.cpp
```diff
@@ -165,6 +184,15 @@ namespace forte::test {
     BOOST_CHECK(test1 >= test2);
     BOOST_CHECK(test1 > test2);
 
+    // same value, different types - should be equal via type promotion
+    test2 = CIEC_LINT(17);
+    BOOST_CHECK(test1 == test2);
```
franz-hoepfinger-4diac, line 189: @mx990 TODAY this test is green. if we would change == to use .equals then it would no more be green. this has to be wisely decided.

franz-hoepfinger-4diac:
> I like the idea of using the new C++ three-way comparison to better indicate incomparable results and also reduce implementation overhead. However, we should make sure to consistently apply these changes to all data types. I also believe we should split changes in the data types into a separate PR from adapting the FB types.

YES ! let us check if this one is green, and make the final decisions, then we should/could make a 2-step approach: 

1. change the Datatypes (in 1 or n PRs)
2. change the FBs using the Compare Functions seperately.

MartinMelikMerkumians:
@franz-hoepfinger-4diac I got 10 E-Mail notifications from you only in this issue in the last 15 minutes!

Could you please use issue comments less like a instant chat and with more care? This feels like spam

franz-hoepfinger-4diac:
> @franz-hoepfinger-4diac I got 10 E-Mail notifications from you only in this issue in the last 15 minutes!
> 
> Could you please use issue comments less like a instant chat and with more care? This feels like spam

i am sorry. how can i respond to every Message otherwise ? I have no Idea ?

core/src/datatypes/forte_any_elementary_variant.cpp
mx990: I meant that we should produce the error message entirely in the FB type and not split it between the data and FB type.

mx990:
For replies to review comments, a way that I found works well is to go into the *Files changed* tab and then make sure to use *Start a review* with your first reply. You can then use *Submit review* and *Comment* to send all your replies at the same time.

franz-hoepfinger-4diac:
> For replies to review comments, a way that I found works well is to go into the _Files changed_ tab and then make sure to use _Start a review_ with your first reply. You can then use _Submit review_ and _Comment_ to send all your replies at the same time.

i saw this, but i was always anxious to use that, as i wanted only to give hubmual comments, not a "big" review. OK, i will test this next time. !!!! good tip !

diplfranzhoepfinger:
NOW: 

the log looks like this: 

```
ERROR: T#38059025000: Comparing incompatible types BYTE and UDINT ERROR: T#38059398000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
ERROR: T#38160072000: Comparing incompatible types BYTE and UDINT ERROR: T#38160406000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
ERROR: T#38355580000: Comparing incompatible types BYTE and UDINT ERROR: T#38356006000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
ERROR: T#38457183000: Comparing incompatible types BYTE and UDINT ERROR: T#38457567000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
ERROR: T#38874208000: Comparing incompatible types BYTE and UDINT ERROR: T#38874653000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
ERROR: T#39079026000: Comparing incompatible types BYTE and UDINT ERROR: T#39079409000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
ERROR: T#39180209000: Comparing incompatible types BYTE and UDINT ERROR: T#39180606000:  in App.Baumsetzer.Start_Baumsetzer.F_EQ_1!
```

core/include/forte/datatypes/forte_any_elementary_variant.h
franz-hoepfinger-4diac: @mx990 @azoitl pls give a guidance about this topic. THANKS !

mx990: I have been thinking a bit more on this. We should probably move the definition of `operator<=>` into the `cpp` file and inline `compare` into it. We should then also separately implement `operator==` in the `cpp` file similar to `compare`.

franz-hoepfinger-4diac: OK, we will go that way ! cool idea !

franz-hoepfinger-4diac: this is done with https://github.com/eclipse-4diac/4diac-forte/pull/889/changes/0006052c60ff7d2fa96478abf970d4fae8d7f259

core/src/datatypes/forte_any_elementary_variant.cpp
franz-hoepfinger-4diac: done with https://github.com/eclipse-4diac/4diac-forte/pull/889/changes/d8105d3b9bcf5678c629141d3e02a177d18d7010

core/src/datatypes/forte_any_elementary_variant.cpp
franz-hoepfinger-4diac: done with https://github.com/eclipse-4diac/4diac-forte/pull/889/changes/d8105d3b9bcf5678c629141d3e02a177d18d7010

core/src/datatypes/forte_any_elementary_variant.cpp
franz-hoepfinger-4diac: done

core/src/datatypes/forte_any_elementary_variant.cpp
franz-hoepfinger-4diac: i will do for all Datatypes once this is approved yes.

franz-hoepfinger-4diac (commented):
this is now Ready to review. If it is approved, i will make all Datatypes using `std::partial_ordering` OR `std::strong_ordering`, and meaningful Loggings in all Compare FBs.

azoitl:

PR #890, opened 2026-04-29
Introduce optional RAM disk for file I/O

franz-hoepfinger-4diac:
Adds a new build option, `FORTE_FILEIO_RAMDISK`, to enable an in-memory file system using `fmemopen`. This allows files to be loaded into RAM and accessed via standard file I/O functions.

This feature is particularly useful for embedded systems or performance-critical scenarios where traditional file system access is not feasible or desired. Enabling the option requires the `fmemopen` function to be available on the target system.

### DIFF
--- a/core/arch/common/src/genforte_fileio_ramdisk.cpp
+++ b/core/arch/common/src/genforte_fileio_ramdisk.cpp
@@ -50,11 +50,15 @@ int forte_ramdisk_load(const char *filename) {
 
   std::ifstream file(filename, std::ios::binary | std::ios::ate);
   if (!file) {
+    std::lock_guard<std::mutex> lock(g_ramdiskMutex);
+    g_ramdiskEntries[filename] = std::vector<char>();
     return -1;
   }
 
   std::streamsize size = file.tellg();
   if (size < 0) {
+    std::lock_guard<std::mutex> lock(g_ramdiskMutex);
+    g_ramdiskEntries[filename] = std::vector<char>();
     return -1;
   }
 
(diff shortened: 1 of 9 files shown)


core/arch/common/src/genforte_fileio_ramdisk.cpp
```diff
@@ -0,0 +1,139 @@
...
+      if (mode[0] == 'r') {
+        FILE *file = fmemopen(it->second.data(), it->second.size(), mode);
+        if (file != nullptr) {
+          g_ramdiskOpenFiles[file] = {filename, mode[0]};
+        }
+        return file;
+      }
+    }
+  }
+  return fopen(filename, mode);
```
franz-hoepfinger-4diac, line 105: in a 2nd Stage we will have to care about Writers, like the CSV Logger, 
it does not need to load the whole file to RAM, but just write the Append chunks to the Flash. 
this will be next step. 
for now the bootfile.

core/arch/common/src/genforte_fileio_ramdisk.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <mutex>
+#include <string>
+#include <vector>
+
+struct RamdiskFileInfo {
+    std::string filename;
+    char mode;
+};
+
+static std::map<std::string, std::vector<char>> g_ramdiskEntries;
```
franz-hoepfinger-4diac, line 29: I initally wanted to use vector instead of malloc/free.

franz-hoepfinger-4diac:
this IS working, i tested on a ESP32, and it solves the Problem; but there are many other Problems coming up. NVS, INI etc.. any File access must be buffered. As ESP32 is 10 years old, and ESP32S3 and ESP32P4 is available, and we did not use ESP32 in Products for some years now, we will discontinue the use of ESP32 and 4diac. 

this PR is written in PURE POSIX, and it could stay here or get merged, it is fully tested and proven functional, so it may serve somebody in the Future.

azoitl:

PR #903, opened 2026-05-11
Fix: Extract unique_ptr deleter to named struct

franz-hoepfinger-4diac:
Replaces an inline lambda deleter with a named struct `STCBDeleter` for the `mTask` unique_ptr. This avoids a Compiler Warning about internal linking.

https://github.com/eclipse-4diac/4diac-forte/pull/863

this fixes this Warning: 

```
/home/franz/git/hr/LOGIBUS_integration_datapanel/4diac-forte/core/arch/freeRTOS/include/forte/arch/forte_thread.h:48:9: warning: 'forte::arch::CFreeRTOSThread' has a field 'std::unique_ptr<xSTATIC_TCB, forte::arch::CFreeRTOSThread::<lambda(StaticTask_t*)> > forte::arch::CFreeRTOSThread::mTask' whose type has internal linkage [-Wsubobject-linkage]
   48 |   class CFreeRTOSThread : public CThreadBase<TaskHandle_t> {
      |         ^~~~~~~~~~~~~~~
```

### DIFF
--- a/core/arch/freeRTOS/include/forte/arch/forte_thread.h
+++ b/core/arch/freeRTOS/include/forte/arch/forte_thread.h
@@ -103,7 +103,15 @@ namespace forte::arch {
       static const int scmForteTaskPriority;
 
       std::vector<StackType_t> mFreeRTOSStack;
-      std::unique_ptr<StaticTask_t, decltype([](StaticTask_t *paTask) { vPortFree(paTask); })> mTask;
+
+      // use a functor instead a lambda, to avoid a "internal linkage warning" at compile-time.
+      struct STCBDeleter {
+          void operator()(StaticTask_t *paTask) const {
+            vPortFree(paTask);
+          }
+      };
+
+      std::unique_ptr<StaticTask_t, STCBDeleter> mTask;
       TaskHandle_t mFreeRTOSThreadHandle = nullptr;
   };
 


core/arch/freeRTOS/include/forte/arch/forte_thread.h
```diff
@@ -74,6 +74,12 @@ namespace forte::arch {
       }
 
     private:
+      struct CTCBDeleter {
```
azoitl: Please move directly to the user of this functor. 
Please note that our coding guidelines demand for an S as prefix not a C.

I guess it would also be good to have a comment there explaining why a functor and not a lambda is used so that the next contributor is changing it back.

azoitl (changes requested):
Referring to our contribution guidelines: Commit messages should be self contained and explain the change and especially for fixes like yours the reason for this change.

Your commit message is definitely not that. It is even plain wrong because this change is definitely not increasing readability and maintainability. Also I would not call it a refactor as it is fixing something. Namely the problem that you mention in your issue #863. And after looking up the explanation why this warning is a real problem (something I would have been happy if you could have provided in your issue description) I totally agree that this change is really needed. 

All this information is missing here.

azoitl (approved):

PR #907, opened 2026-05-14
Replace E_PERMIT with E_R_TRIG in E_TP timer function block

franz-hoepfinger-4diac:
Update the internal logic of the E_TP pulse timer to use E_R_TRIG for rising edge detection. This replaces E_PERMIT, ensuring more precise triggering of the timer's internal state.

https://github.com/eclipse-4diac/4diac-ide/pull/2455


azoitl (approved):
LGTM

PR #918, opened 2026-05-16
Fix: Correct data type in DWORD splitting function

franz-hoepfinger-4diac:
Change the data type for the second word extraction from CIEC_BOOL to CIEC_WORD to ensure correct data representation in the DWORD splitting process.

https://github.com/eclipse-4diac/4diac-ide/pull/2476


azoitl:

PR #921, opened 2026-05-17
Added JSON Parser FBs

mayrthom:
Created a JSON Encoder and Decoder, useful for connecting 4diac to Node-RED. The JSON Encoder is modular and can be connected in series for multiple JSON outputs.

### DIFF
--- a/modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
+++ b/modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
@@ -64,7 +64,6 @@ void FORTE_ReadJSON::executeEvent(const TEventID paEIID, CEventChainExecutionThr
     switch(paEIID) {
         case scmEventREQID: {
             alg_REQ();
-            writeOutputData(scmEventCNFID);
             sendOutputEvent(scmEventCNFID, paECET);
             break;
         }
@@ -134,8 +133,8 @@ void FORTE_ReadJSON::executeEvent(const TEventID paEIID, CEventChainExecutionThr
   }
 
 void FORTE_ReadJSON::alg_REQ(void){
-    std::string JSONInStr = var_JSONIn.c_str();
-    std::string field = var_FieldName.c_str();
+    std::string JSONInStr = var_JSONIn.getStorage();
+    std::string field = var_FieldName.getStorage();
 
     std::string pattern = "\""+field+"\"\\s*:\\s*(\"[^\"]*\"|\\{[^}]*\\}|\\d+\\.?\\d*|true|false|null)";
     std::regex re(pattern);
@@ -152,9 +151,7 @@ if(valueStr == "true" || valueStr == "false") {
             if(isString) {
                 valueStr = valueStr.substr(1, valueStr.size()-2);
             }
-            CIEC_STRING tmpString;
-            tmpString.assign(valueStr.c_str(), static_cast<TForteUInt16>(valueStr.size()));
-            var_Value = CIEC_ANY_VARIANT(tmpString);
+            var_Value = CIEC_ANY_VARIANT(CIEC_STRING(valueStr));
         }
 
     } else {
--- a/modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
+++ b/modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
@@ -1,0 +1,165 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Thomas Mayr
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Thomas Mayr - Initial implementation.
+ *******************************************************************************/
+
+#include "forte/eclipse4diac/convert/ReadJSON_fbt.h"
+
+#include "forte/forte_st_util.h"
+#include <string>
+#include <regex>
+
+using namespace std::literals;
+using namespace forte::literals;
+
+namespace forte::eclipse4diac::convert {
+  namespace {
+    constexpr std::string_view TypeHash =""sv;
+
+    const auto cEventInputNames = std::array{"REQ"_STRID};
+    const auto cEventOutputNames = std::array{"CNF"_STRID};
+    const auto cDataInputNames = std::array{"JSONIn"_STRID, "FieldName"_STRID};
+    const auto cDataOutputNames = std::array{"Value"_STRID};
+    const SFBInterfaceSpec cFBInterfaceSpec = {
+        .mEINames = cEventInputNames,
+        .mEITypeNames = {},
+        .mEONames = cEventOutputNames,
+        .mEOTypeNames = {},
+        .mDINames = cDataInputNames,
+        .mDONames = cDataOutputNames,
+        .mDIONames = {},
+        .mSocketNames = {},
+        .mPlugNames = {},
+    };
+  }
+
+  DEFINE_FIRMWARE_FB(FORTE_ReadJSON, "eclipse4diac::convert::ReadJSON"_STRID, TypeHash)
+
+  FORTE_ReadJSON::FORTE_ReadJSON(const StringId paInstanceNameId, CFBContainer &paContainer) :
+      CFunctionBlock(paContainer, cFBInterfaceSpec, paInstanceNameId),
+      var_JSONIn(""_STRING),
+      var_FieldName(""_STRING),
+      var_Value(CIEC_ANY_VARIANT()),
+      conn_CNF(*this, 0),
+      conn_JSONIn(nullptr),
+      conn_FieldName(nullptr),
+      conn_Value(*this, 0, var_Value) {
+  };
+
+  void FORTE_ReadJSON::setInitialValues() {
+    CFunctionBlock::setInitialValues();
+    var_JSONIn = ""_STRING;
+    var_FieldName = ""_STRING;
+    var_Value = CIEC_ANY_VARIANT();
+  }
+
+void FORTE_ReadJSON::executeEvent(const TEventID paEIID, CEventChainExecutionThread *const paECET) {
+    switch(paEIID) {
+        case scmEventREQID: {
+            alg_REQ();
+            writeOutputData(scmEventCNFID);
+            sendOutputEvent(scmEventCNFID, paECET);
+            break;
+        }
+        default:
+            break;
+    }
+}
+
+  void FORTE_ReadJSON::readInputData(const TEventID paEIID) {
+    switch(paEIID) {
+      case scmEventREQID: {
+        readData(0, var_JSONIn, conn_JSONIn);
+        readData(1, var_FieldName, conn_FieldName);
+        break;
+      }
+      default:
+        break;
+    }
+  }
+
+  void FORTE_ReadJSON::writeOutputData(const TEventID paEIID) {
+    switch(paEIID) {
+      case scmEventCNFID: {
+        writeData(2, var_Value, conn_Value);
+        break;
+      }
+      default:
+        break;
+    }
+  }
+
+  CIEC_ANY *FORTE_ReadJSON::getDI(const size_t paIndex) {
+    switch(paIndex) {
+      case 0: return &var_JSONIn;
+      case 1: return &var_FieldName;
+    }
+    return nullptr;
+  }
+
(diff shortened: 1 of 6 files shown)


mayrthom:
Changed it to the convert package now.

modules/convert/src/eclipse4diac/convert/ToJSON_fbt.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  CDataConnection *FORTE_ToJSON::getDOConUnchecked(const TPortId paIndex) {
+    switch(paIndex) {
+      case 0: return &conn_JSONOut;
+    }
+    return nullptr;
+  }
+
+void FORTE_ToJSON::alg_REQ(void){
+    std::string tmp(func_ANY_AS_STRING(var_Value).c_str(), func_ANY_AS_STRING(var_Value).length());
```
azoitl: you can use here `CIEC_STRING::getStorage()` this will be better readable and more performant.

modules/convert/src/eclipse4diac/convert/ToJSON_fbt.cpp
```diff
@@ -0,0 +1,205 @@
...
+    if(var_JSONIn == ""){
+        JSONOutStd = "{\"" + std::string(var_FieldName.c_str(), var_FieldName.length()) + "\":" + valueStr + "}";
+    } else {
+        JSONOutStd = std::string(var_JSONIn.c_str(), var_JSONIn.length());
+        if(!JSONOutStd.empty() && JSONOutStd.back() == '}'){
+            JSONOutStd.pop_back();
+            JSONOutStd += ", \"" + std::string(var_FieldName.c_str(), var_FieldName.length()) + "\":" + valueStr + "}";
+        }
+    }
+    var_JSONOut.assign(JSONOutStd.c_str(), static_cast<TForteUInt16>(JSONOutStd.size()));
```
azoitl: Even if it looks a bit strange you can change that to:

`var_JSONOut = CIEC_STRING(JSONOutStd);`

modules/convert/src/eclipse4diac/convert/ToJSON_fbt.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return nullptr;
+  }
+
+void FORTE_ToJSON::alg_REQ(void){
+    std::string tmp(func_ANY_AS_STRING(var_Value).c_str(), func_ANY_AS_STRING(var_Value).length());
+
+    if(!tmp.empty() && tmp.front() == '\'') tmp.erase(0,1);
+    if(!tmp.empty() && tmp.back() == '\'') tmp.pop_back();
+
+    auto escapeJSONString = [](const std::string &input) -> std::string {
```
azoitl: I would move this to a function in an anonymous namespace. This would make the code better readable.

modules/convert/src/eclipse4diac/convert/ToJSON_fbt.cpp
```diff
@@ -0,0 +1,205 @@
...
+    var_Value = CIEC_ANY_VARIANT();
+    var_FieldName = ""_STRING;
+    var_JSONOut = ""_STRING;
+  }
+
+  void FORTE_ToJSON::executeEvent(const TEventID paEIID, CEventChainExecutionThread *const paECET) {
+    switch(paEIID) {
+        case scmEventREQID: {
+            alg_REQ();
+            writeOutputData(scmEventCNFID);
```
azoitl: do not call `writeOutputData()` here. This is handled correctly by `sendOutputEvent`.

modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  CDataConnection *FORTE_ReadJSON::getDOConUnchecked(const TPortId paIndex) {
+    switch(paIndex) {
+      case 0: return &conn_Value;
+    }
+    return nullptr;
+  }
+
+void FORTE_ReadJSON::alg_REQ(void){
+    std::string JSONInStr = var_JSONIn.c_str();
```
azoitl: as below please use `getStorage` here

modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
```diff
@@ -0,0 +1,165 @@
...
+    std::string field = var_FieldName.c_str();
+
+    std::string pattern = "\""+field+"\"\\s*:\\s*(\"[^\"]*\"|\\{[^}]*\\}|\\d+\\.?\\d*|true|false|null)";
```
azoitl, line 139: I would move pattern and re as const strings into an anonymous namespace.

modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
```diff
@@ -0,0 +1,165 @@
...
+    var_JSONIn = ""_STRING;
+    var_FieldName = ""_STRING;
+    var_Value = CIEC_ANY_VARIANT();
+  }
+
+void FORTE_ReadJSON::executeEvent(const TEventID paEIID, CEventChainExecutionThread *const paECET) {
+    switch(paEIID) {
+        case scmEventREQID: {
+            alg_REQ();
+            writeOutputData(scmEventCNFID);
```
azoitl: please remove `writeOutputData`

azoitl (changes requested):

modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
mayrthom: How do you mean this?
The pattern string consists also of the field itself, which is not constant.

azoitl: Ah sorry I missed that.  Would it make sense then to move the pattern creation to an init event so that you do not need to do it in every parse step?

mx990 (changes requested):
I do not believe parsing JSON via regular expressions is a viable approach[^1]. Instead, I would suggest using either a recursive descent parser or one of the many existing libraries, such as https://github.com/nlohmann/json or https://github.com/Tencent/rapidjson.

[^1]: see https://community.notepad-plus-plus.org/topic/25304/faq-parsing-and-editing-json-with-regex-is-a-bad-idea for reasons why this does not work

mayrthom:
@mx990 Using a dedicated JSON Library was also my first thought, but this would make compiling FORTE more complicated, as the library also have to be referenced and I'm also not sure how this will work with the licence.

The idea was to just encode and decode basic datatypes into simple JSON Strings for MQTT messages.
I know that the design is limited, but it was never the idea to create complex nested JSON Files. This would also require a quite complex function block to configure it and MQTT message are mostly simple JSON Strings.

Implementing a recursive descent parser would work, but does this make really any benefit for such a simple case?

azoitl:
I would also prefer if we could get away without an additional dependency. What I liked on @mayrthom that the FBs are very simple rather straight forward. This allows to build and parse simple Json strings. It is definitely not suited for something as complex as Eclipse Sparkplug. But I would say this is not the intention of these blocks.  

But as we are targeting simple flat Json strings could we get away with searches and string_views?

mx990:
I understand this may be intended for a limited use case, but that is not clear from looking at those blocks. If we add generic blocks for parsing JSON to the standard library, we should support all valid JSON documents.

I also noticed that the example in https://github.com/eclipse-4diac/4diac-ide/pull/2479 in fact *does* seem to target nested JSON objects. This is already too much for a simple regex parser to handle properly. Consider the following JSON document:
```json
{
  "test": {
    "key": false
  },
  "key": true
}
```
When searching for `key`, a simple regex search will return the value `false` inside the nested object `test`, instead of the correct value `true` in the top-level object.

mayrthom:
@mx990 Where to you see that it targets nested JSON strings? I never used a nested JSON string and the example is also not a nested JSON?

As @azoitl said external libraries are not recommended, what is your solution?

azoitl:
> As @azoitl said external libraries are not recommended, what is your solution?

My classical answer is it depends. I originally thought that your block is taking the first json statement from the potential list of json statements if it matches the given field name its value is put on the output and the remaining string is put on the other output. 

As @mx990 points out json can be quite complex. So what complexity of nesting do you want to support? Can the value be a nested json statement? If this is the case I think we should go with an existing json library because that sounds quite complicated. But could still fit to your block. 

If you want to be simpler I thought with simple string search operations you could also get to your value (e.g., `find_first`).  However then some checks should be there if the json string is not as expected so that an error can be indicated to the user of your block.

modules/convert/src/eclipse4diac/convert/ReadJSON_fbt.cpp
```diff
@@ -0,0 +1,162 @@
...
+      case 0: return &conn_Value;
+    }
+    return nullptr;
+  }
+
+void FORTE_ReadJSON::alg_REQ(void){
+    std::string JSONInStr = var_JSONIn.getStorage();
+    std::string field = var_FieldName.getStorage();
+
+    std::string pattern = "\""+field+"\"\\s*:\\s*(\"[^\"]*\"|\\{[^}]*\\}|\\d+\\.?\\d*|true|false|null)";
```
ernstblechaPT, line 139: looking just at this regular expression and assuming that matching nested elements is not a target, based on [json.org](https://json.org):

 - field needs to be escaped, two examples for problems:
   - `.` is a possible character in a field-name, that needs to be escaped for the regular expressions (altough seldom used as it conflicts with simple use in json)
   - `"` is a possible character in a field name, that nees to be escaped for the json string
 - json-strings may contain `"`, if correctly escaped - that is quite a head-scratcher to implement in regular expressions
 - json-strings may not contain `\` followed by anything other than the defined escaped-symbols
 - numbers can be written in scientific notation (`-5.43E+21`)
 - this might also parse invalid entries: e.g. `"abc": truefalse` in my understanding will parse to `true` - even if the full json string is not even a valid object (i.e. missing `{` and `}`)

mayrthom:
@azoitl 
I think the most universal solution would be to use the same approach as Node-RED. There you can use nested JSON Files and go down the file like "payload.sensor1.temp1" which would return the value "temp1" of a nested json where temp1 is a field of sensor1.
However I think this would require an external lib, as implementing this all yourself wouldn't make much sense. So I must say I don't know what the best approach here is.

MartinMelikMerkumians:
Coming also to the party, as I also have a use case for JSON parsing.
I know of two approaches:
1. as @mayrthom mentioned, you specify the field you want to have. Guess this could be done as a comm layer (like) module with a full parser
2. like Beckhoff is doing it, the easy way - a struct specifies the JSON structure and can be directly used for in and out parsing.
3. like Beckhoff is doing it, the hard way - a JSON parsing lib as FBs which are applied to WSTRINGs

I personally like 1. and 2.
1. 20% for use cases where you need the full the flexibility and 2. for 80% of typical use cases (in my limited experience) where you just want to transmit data with a clear structure.

azoitl:
> @azoitl I think the most universal solution would be to use the same approach as Node-RED. There you can use nested JSON Files and go down the file like "payload.sensor1.temp1" which would return the value "temp1" of a nested json where temp1 is a field of sensor1. However I think this would require an external lib, as implementing this all yourself wouldn't make much sense. So I must say I don't know what the best approach here is.

@mayrthom the more comments I read here the more I think we need an to use a json parsing lib. In this case I would recommend that we move your blocks into a dedicated json lib. Do you or any of the people commented here have a lib in mind. I'm happy to support with checking license compatibility and handling the Eclipse process for dependency checks that would be required.

mayrthom:
@MartinMelikMerkumians 
I already had the idea with the structs.
But would it be possible to retrieve the variable names of the struct members during runtime?

azoitl:
> @MartinMelikMerkumians I already had the idea with the structs. But would it be possible to retrieve the variable names of the struct members during runtime?

Yes our struct class has `virtual const StringId *elementNames()` for that.

mayrthom:
It just came to my mind that the struct-based approach might be a bit inflexible, since any change to the JSON structure would require recompiling the source file containing the struct, or am I wrong?

MartinMelikMerkumians:
> It just came to my mind that the struct-based approach might be a bit inflexible, since any change to the JSON structure would require recompiling the source file containing the struct, or am I wrong?

yes, that' s essentially the drawback. But on the other hand, as the parsing/mapping is handled based in the structure, you do not need to change FBs or FB parameters for this.

For both cases the mechanism could be ignoring new fields in the JSON payload to provide some kind of backwards compatability.

PR #923, opened 2026-05-17
Fix incorrect parameter type in getErrorMessage for Win32 socket interface

franz-hoepfinger-4diac:
Corrects the parameter type for FormatMessage to ensure proper allocation of the error message buffer on Win32 systems.


azoitl:

PR #925, opened 2026-05-17
Fix memory leaks by avoiding unnecessary dynamic allocations

franz-hoepfinger-4diac:
Replaced dynamic allocation of type libraries with static allocation to prevent memory leaks. Modified destruction process in CFBContainer to properly deinitialize children before deletion.

This should fix thefrequent  "Detected memory leaks!" Error in the Test Pipeline.

### DIFF
--- a/core/src/fbcontainer.cpp
+++ b/core/src/fbcontainer.cpp
@@ -63,8 +63,11 @@ namespace forte {
   }
 
   CFBContainer::~CFBContainer() {
-    for (TFBContainerList::iterator itRunner(mChildren.begin()); itRunner != mChildren.end(); ++itRunner) {
-      delete (*itRunner);
+    for (auto *child : mChildren) {
+      child->deinitialize();
+    }
+    for (auto *child : mChildren) {
+      delete child;
     }
     mChildren.clear();
   }
--- a/core/src/typelib.cpp
+++ b/core/src/typelib.cpp
@@ -29,23 +29,23 @@ using namespace std::string_literals;
 namespace forte {
   namespace {
     std::vector<CFBTypeEntry *> &getFBTypeLib() {
-      static std::vector<CFBTypeEntry *> *fbTypeLib = new std::vector<CFBTypeEntry *>();
-      return *fbTypeLib;
+      static std::vector<CFBTypeEntry *> fbTypeLib;
+      return fbTypeLib;
     }
 
     std::vector<CAdapterTypeEntry *> &getAdapterTypeLib() {
-      static std::vector<CAdapterTypeEntry *> *adapterTypeLib = new std::vector<CAdapterTypeEntry *>();
-      return *adapterTypeLib;
+      static std::vector<CAdapterTypeEntry *> adapterTypeLib;
+      return adapterTypeLib;
     }
 
     std::vector<CDataTypeEntry *> &getDataTypeLib() {
-      static std::vector<CDataTypeEntry *> *dataTypeLib = new std::vector<CDataTypeEntry *>();
-      return *dataTypeLib;
+      static std::vector<CDataTypeEntry *> dataTypeLib;
+      return dataTypeLib;
     }
 
     std::vector<CGlobalConstEntry *> &getGlobalConstTypeLib() {
-      static std::vector<CGlobalConstEntry *> *globalConstTypeLib = new std::vector<CGlobalConstEntry *>();
-      return *globalConstTypeLib;
+      static std::vector<CGlobalConstEntry *> globalConstTypeLib;
+      return globalConstTypeLib;
     }
 
     CFunctionBlock *createGenericFB(StringId paInstanceNameId,


core/src/typelib.cpp
```diff
@@ -29,23 +29,23 @@ using namespace std::string_literals;
...
   namespace {
     std::vector<CFBTypeEntry *> &getFBTypeLib() {
-      static std::vector<CFBTypeEntry *> *fbTypeLib = new std::vector<CFBTypeEntry *>();
-      return *fbTypeLib;
+      static std::vector<CFBTypeEntry *> fbTypeLib;
+      return fbTypeLib;
     }
 
     std::vector<CAdapterTypeEntry *> &getAdapterTypeLib() {
-      static std::vector<CAdapterTypeEntry *> *adapterTypeLib = new std::vector<CAdapterTypeEntry *>();
```
franz-hoepfinger-4diac, line 37: this `new` made a Memory-Leak, as it was never deleted.

core/src/typelib.cpp
```diff
@@ -29,23 +29,23 @@ using namespace std::string_literals;
 namespace forte {
   namespace {
     std::vector<CFBTypeEntry *> &getFBTypeLib() {
-      static std::vector<CFBTypeEntry *> *fbTypeLib = new std::vector<CFBTypeEntry *>();
```
franz-hoepfinger-4diac, line 32: this `new` made a Memory-Leak, as it was never deleted.

core/src/typelib.cpp
```diff
@@ -29,23 +29,23 @@ using namespace std::string_literals;
...
 
     std::vector<CAdapterTypeEntry *> &getAdapterTypeLib() {
-      static std::vector<CAdapterTypeEntry *> *adapterTypeLib = new std::vector<CAdapterTypeEntry *>();
-      return *adapterTypeLib;
+      static std::vector<CAdapterTypeEntry *> adapterTypeLib;
+      return adapterTypeLib;
     }
 
     std::vector<CDataTypeEntry *> &getDataTypeLib() {
-      static std::vector<CDataTypeEntry *> *dataTypeLib = new std::vector<CDataTypeEntry *>();
```
franz-hoepfinger-4diac, line 42: this `new` made a Memory-Leak, as it was never deleted.

core/src/typelib.cpp
```diff
@@ -29,23 +29,23 @@ using namespace std::string_literals;
...
-      return *dataTypeLib;
+      static std::vector<CDataTypeEntry *> dataTypeLib;
+      return dataTypeLib;
     }
 
     std::vector<CGlobalConstEntry *> &getGlobalConstTypeLib() {
-      static std::vector<CGlobalConstEntry *> *globalConstTypeLib = new std::vector<CGlobalConstEntry *>();
```
franz-hoepfinger-4diac, line 47: this `new` made a Memory-Leak, as it was never deleted.

core/src/fbcontainer.cpp
```diff
@@ -63,8 +63,11 @@ namespace forte {
   }
 
   CFBContainer::~CFBContainer() {
-    for (TFBContainerList::iterator itRunner(mChildren.begin()); itRunner != mChildren.end(); ++itRunner) {
-      delete (*itRunner);
+    for (auto *child : mChildren) {
+      child->deinitialize();
```
franz-hoepfinger-4diac, line 67: first call `deinitialize()`

core/src/fbcontainer.cpp
```diff
@@ -63,8 +63,11 @@ namespace forte {
...
+    }
+    for (auto *child : mChildren) {
+      delete child;
```
franz-hoepfinger-4diac, line 70: then call `delete`

@azoitl should it be 1 for-loop, or better this way ?

franz-hoepfinger-4diac (commented):
Ready for Review. 
all 3 Builds and Tests are Green, 
and: the `Detected memory leaks!` which appeared frequently on Windows builds is not here anymore.

franz-hoepfinger-4diac:
# typelib.cpp — Static Vector Leak Fix


i asked my colleague if this is really safe, and here is the Answer. For me it sounds reasonable. 

## Change

**Before (leak):**
```cpp
static std::vector<CFBTypeEntry *> *fbTypeLib = new std::vector<CFBTypeEntry *>();
return *fbTypeLib;
```

**After (no leak):**
```cpp
static std::vector<CFBTypeEntry *> fbTypeLib;
return fbTypeLib;
```

## Why This Is Correct

The function-local `static` variable is automatically destroyed at program exit — C++ guarantees this.

The `new`-based trick was originally intended to avoid the *static initialization order fiasco*. However, it is unnecessary here because these functions (`getFBTypeLib()`, `getAdapterTypeLib()`, `getDataTypeLib()`, `getGlobalConstTypeLib()`) are always initialized on first call (Meyers Singleton pattern). Since C++11, the standard guarantees **thread-safe initialization** of function-local statics, so no manual heap allocation is needed.

azoitl:

PR #950, opened 2026-06-05
Fix out-of-bounds array access in COPC_UA_Local_Handler

mx990:
The parent lookup in `COPC_UA_Local_Handler::initializeCreateMethod` did read beyond the end of `presentNodes` when comparing elements to look for the parent. This simplifies the code to use simple index-based access.

### DIFF
--- a/com/opc_ua/src/opcua_local_handler.cpp
+++ b/com/opc_ua/src/opcua_local_handler.cpp
@@ -718,13 +718,8 @@ namespace forte::com_infra::opc_ua {
         UA_NodeId parent = UA_NODEID_NUMERIC(0, UA_NS0ID_ROOTFOLDER);
 
         // look for parent object
-        for (auto itPresentNodes = presentNodes.begin(); itPresentNodes != presentNodes.end();) {
-          auto currentIterator = itPresentNodes;
-          ++itPresentNodes;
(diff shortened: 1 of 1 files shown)


com/opc_ua/src/opcua_local_handler.cpp
```diff
@@ -718,13 +718,8 @@ namespace forte::com_infra::opc_ua {
...
-        for (auto itPresentNodes = presentNodes.begin(); itPresentNodes != presentNodes.end();) {
-          auto currentIterator = itPresentNodes;
-          ++itPresentNodes;
-          if (*itPresentNodes == presentNodes.back()) {
-            parent = **currentIterator;
-            break;
-          }
+        if (presentNodes.size() > 1) {
+          parent = *presentNodes[presentNodes.size() - 2];
         }
```
mx990, line 723: It seems the current implementation is overly complicated because `presentNodes` used to be a linked list, but it would be great if someone could confirm that this is really the intended behavior.

azoitl: I checked how the vector is filled and I'm pretty sure that your proposed solution is the right one and yes most probably originated from using our CSinglyLinkedList.

azoitl (approved):

PR #952, opened 2026-06-05
Fix memory leak in OPC UA logger

mx990:
The previous logger was not cleared properly when assigning a custom logger, which led to a memory leak. This first creates a config with the custom logger and then creates a client or server from it[^1], thus avoiding either memory leaks or dangling pointers.

[^1]: see also https://github.com/open62541/open62541/issues/6769#issuecomment-4621348961

### DIFF
--- a/com/opc_ua/src/opcua_local_handler.cpp
+++ b/com/opc_ua/src/opcua_local_handler.cpp
@@ -25,6 +25,7 @@
 #include "opcua_local_handler.h"
 
 #include <open62541/server.h>
+#include <open62541/server_config_default.h>
 
 #include "forte/iec61131_functions.h"
 #include "forte/cominfra/basecommfb.h"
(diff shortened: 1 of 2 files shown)


com/opc_ua/src/opcua_local_handler.cpp
```diff
@@ -97,20 +98,23 @@ namespace forte::com_infra::opc_ua {
...
-      UA_ServerStrings serverStrings;
-      generateServerStrings(gOpcuaServerPort.mArgument, serverStrings);
-      configureUAServer(serverStrings, *uaServerConfig);
+    UA_ServerConfig config{
+        config.logging = &getLogger(),
+    };
+
+    // set default config
+    if (const UA_StatusCode status = UA_ServerConfig_setDefault(&config); status != UA_STATUSCODE_GOOD) {
+      DEVLOG_ERROR("[OPC UA CLIENT]: Error setting server configuration. Error: %s\n", UA_StatusCode_name(status));
```
m-meingast:
```suggestion
      DEVLOG_ERROR("[OPC UA LOCAL]: Error setting server configuration. Error: %s\n", UA_StatusCode_name(status));
```

mx990: Done, thanks for noticing!

m-meingast (approved):

azoitl (approved):

PR #972, opened 2026-06-22
Restructure Command parsing

azoitl:
Base work to reduce state of command parser.

### DIFF
--- a/core/src/mgmcmdstruct.cpp
+++ b/core/src/mgmcmdstruct.cpp
@@ -1,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Johannes Kepler University Linz
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Alois Zoitl - initial implementation and rework communication infrastructure
+ *******************************************************************************/
+
+#include "forte/mgmcmdstruct.h"
+
+namespace forte {
+
+  void SManagementCMD::clear() {
+    mCMD = EMGMCommandType::INVALID;
+    mDestination = StringId{};
+    mFirstParam.clear();
+    mSecondParam.clear();
+    mAdditionalParams.clear();
+    mID = nullptr;
+    mMonitorResponse.clear();
+  }
+} // namespace forte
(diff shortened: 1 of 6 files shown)


core/src/mgmcmdstruct.cpp
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Contributors:
+ *   Alois Zoitl - initial implementation and rework communication infrastructure
+ *******************************************************************************/
+
+#include "forte/mgmcmdstruct.h"
+
+namespace forte {
+
+  void SManagementCMD::clear() {
```
cochicde, line 18: can't we have default values where needed, and then just reconstruct the object instead of clearing?

stdfblib/system/src/CommandParser.cpp
```diff
@@ -25,19 +25,10 @@ namespace forte::iec61499::system {
...
   EMGMResponse CommandParser::parseAndExecuteMGMCommand(const char *const paDest, char *paCommand) {
-    mLastResponse = EMGMResponse::InvalidObject;
-
     if (nullptr != strchr(paCommand, '>')) {
-      mCommand.mID = nullptr;
-      mCommand.mAdditionalParams.clear();
-      mCommand.mFirstParam.clear();
-      mCommand.mSecondParam.clear();
-      mCommand.mMonitorResponse.clear();
+      mCommand.clear();
```
cochicde, line 29: I see here `mCommand = SManagementCMD();`

core/src/mgmcmdstruct.cpp
cochicde: Or is this needed to avoid reallocating the buffers of the params?

azoitl: Exactly recreating the object would lead to a lot of allocation and reallocation that we need to avoid.

stdfblib/system/src/CommandParser.cpp
```diff
@@ -67,69 +58,69 @@ namespace forte::iec61499::system {
         }
 
         if (EMGMCommandType::INVALID != mCommand.mCMD) {
-          mLastResponse = mDevice.executeMGMCommand(mCommand);
+          return mDevice.executeMGMCommand(mCommand);
         }
```
mx990, line 62: Could we also decouple parsing and executing of commands, so we can more easily add unit tests for the command parser in the future?

mx990 (changes requested):

stdfblib/system/src/CommandParser.cpp
azoitl: yes this is on my todolist. I was not sure if I first improve parsing in general or first extract. But I think extracting the execution is better. I would do this then as next PR.

mx990 (approved):

PR #968, opened 2026-06-18
Add core EtherCAT library integration (PR 1 of 3)

wzj7531:
## Summary

First of three PRs split from #967 per maintainer feedback.

- Add `FORTE_MODULE_ETHERCAT` CMake option for Linux
- Add IgH EtherCAT Master userspace integration (bus handler, device layer)
- Update `dependencies.spdx` with IgH ethercat library entry for CQ process

**Review feedback addressed (force-pushed):**
- Renamed runtime layer from `slave/` to `device/` with controller/device terminology
- `ECBusDeviceHandler` / `ECDeviceHandle` replace former slave handler/handle names
- Process images use `std::vector<unsigned char>` (RAII) instead of manual `new[]`/`delete[]`
- Handle buffer views use `std::span<unsigned char>`
- `mUpdateMutex` and `mDevice` bound by reference in `ECDeviceHandle`

IgH API names (`ecrt_master_*`, `ecrt_slave_*`) and FORTE core IO base classes remain unchanged.

## Related PRs

- Closes #967 (superseded by this split series)
- eclipse-4diac/4diac-fbe#72 (build dependency recipes)
- eclipse-4diac/4diac-ide#2581 (ESI import, planned for PR 3)

## Planned follow-up PRs

2. IO function blocks (ECController, ECDevice, ECModule, ECCoupler)
3. ESI parsing and generic FB support

## Test plan

- [x] Build forte with `FORTE_MODULE_ETHERCAT=ON` on Linux (core-only module)
- [ ] Verify module links against IgH libethercat
- [ ] Confirm `dependencies.spdx` entry is sufficient for CQ

PR #967, opened 2026-06-16
Add modular EtherCAT IO module

wzj7531:
## Summary
- Add FORTE_MODULE_ETHERCAT for Linux with ECMaster, ECSlave, ECModule, and ECCoupler FBs
- Add ESI file parsing and IgH EtherCAT Master integration
- Add IOConfigHandlerFBMultiSlave for multi-slave IO handlers

## Related PRs
- eclipse-4diac/4diac-fbe#72
- eclipse-4diac/4diac-ide#2581

## Test plan
- [ ] Build forte with FORTE_MODULE_ETHERCAT=ON
- [ ] Verify EtherCAT master/slave FB deployment on Linux

PR #982, opened 2026-07-07
Add EtherCAT IO function blocks (PR 2 of 3)

wzj7531:
## Summary

Second PR in the **4diac FORTE** split series from #967 (approved by @azoitl), building on #968 (merged core library).

### Planned PR 2 scope (IO function blocks)

- Add `IOConfigHandlerFBMultiSlave` base class for multi-device IO handlers
- Add EtherCAT IO function blocks: `ECController`, `ECBusAdapter`, `ECDevice`, `ECModule`, `ECCoupler`
- Add device/module config structs and `ECModuleHandler` device layer
- Unify **ControllerId** naming on `ECController` and `ECBusAdapter` per @azoitl review

### Planned PR 3 scope (runtime ESI + generic FB) — included here

Per the original #967 split, FORTE **PR 3** was *ESI parsing and generic FB support*. That runtime code is included in this PR because `ECDevice` and `ECModule` depend on it at configuration time:

- Runtime `EsiFileParser` (ESI XML → PDO/handle setup)
- Generic firmware FBs: `GEN_ECDevice`, `GEN_ECModule`, `GEN_ECCoupler`
- Link `tinyxml` for ESI parsing

After this PR merges, the **4diac FORTE EtherCAT module** from the #967 split series should be complete on the Linux/IgH path.

IgH API names and FORTE core IO base classes (`IOConfigFBMultiMaster`, `var_MasterId`, …) remain unchanged.

## Related PRs (separate tracks)

| Track | PR | Repository |
|-------|-----|------------|
| FORTE PR 1 (core) | #968 (merged) | 4diac-forte |
| FORTE PR 2+3 (this PR) | #982 | 4diac-forte |
| Build dependency recipes | eclipse-4diac/4diac-fbe#72 | 4diac-fbe |
| **IDE** (typelibrary, ESI import wizard, deployment) | eclipse-4diac/4diac-ide#2581 | **4diac IDE** |

The **4diac IDE** work is a **separate repository and PR series**, not FORTE PR 3. IDE #2581 covers typelibrary entries, ESI import, and deployment alignment (e.g. `ControllerId` on `ECController.fbt`).

## Test plan

- [x] Build 4diac FORTE with `FORTE_MODULE_ETHERCAT=ON` on Linux
- [ ] Deploy application with `ECController` + `ECDevice` chain
- [ ] Verify IO handle mapping after controller activation

### DIFF
--- a/core/include/forte/io/configFB/CMakeLists.txt
+++ b/core/include/forte/io/configFB/CMakeLists.txt
@@ -1,5 +1,5 @@
 #############################################################################
-# Copyright (c) 2025 Martin Erich Jobst
+# Copyright (c) 2025 Martin Erich Jobst, Sichuan Qunyuan Technology Co., Ltd.
 #
 # This program and the accompanying materials are made available under the
 # terms of the Eclipse Public License 2.0 which is available at
@@ -10,6 +10,8 @@
 # Contributors:
 #    Martin Erich Jobst
 #      - initial API and implementation and/or initial documentation
+#    Zijun Tang
+#      - add createSlaveHandler() hook and generic FB constructor to io_slave_multi
 #############################################################################
 
 target_sources(forte-core PUBLIC
--- a/core/src/io/configFB/io_slave_multi.cpp
+++ b/core/src/io/configFB/io_slave_multi.cpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 - 2018 fortiss GmbH
+ * Copyright (c) 2017 fortiss GmbH, Sichuan Qunyuan Technology Co., Ltd.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -9,6 +9,7 @@
  * Contributors:
  *   Johannes Messmer - initial API and implementation and/or initial documentation
  *   Jose Cabral - Cleaning of namespaces
+ *   Zijun Tang - add createSlaveHandler() hook and generic FB constructor
  *******************************************************************************/
 
 #include "forte/io/configFB/io_slave_multi.h"
@@ -22,6 +23,7 @@ namespace forte::io {
   const char *const IOConfigFBMultiSlave::scmMasterNotFound("Master not found");
   const char *const IOConfigFBMultiSlave::scmNotFound("Module not found");
   const char *const IOConfigFBMultiSlave::scmIncorrectType("Module type is incorrect");
+  const char *const IOConfigFBMultiSlave::scmCreateSlaveHandlerFailed("Create slave handler failed");
 
   IOConfigFBMultiSlave::IOConfigFBMultiSlave(const TForteUInt8 *const paSlaveConfigurationIO,
                                              const TForteUInt8 paSlaveConfigurationIONum,
@@ -40,6 +42,16 @@ namespace forte::io {
     for (int i = 0; i < mSlaveConfigurationIONum; i++) {
       mSlaveConfigurationIOIsDefault[i] = false;
     }
+  }
+
+  IOConfigFBMultiSlave::IOConfigFBMultiSlave(CFBContainer &paContainer,
+                                             const SFBInterfaceSpec &paInterfaceSpec,
+                                             forte::StringId paInstanceNameId,
+                                             const TForteUInt8 *const paSlaveConfigurationIO,
+                                             TForteUInt8 paSlaveConfigurationIONum,
+                                             int paType) :
+      IOConfigFBMultiSlave(
+          paSlaveConfigurationIO, paSlaveConfigurationIONum, paType, paContainer, paInterfaceSpec, paInstanceNameId) {
   }
 
   IOConfigFBMultiSlave::~IOConfigFBMultiSlave() {
@@ -159,6 +171,11 @@ namespace forte::io {
     }
 
     IODeviceMultiController &controller = getController();
+
+    if (!createSlaveHandler()) {
+      DEVLOG_ERROR("[IOConfigFBMultiSlave] Create slave handler failed\n");
+      return scmCreateSlaveHandlerFailed;
+    }
 
     // Check if slave exists
     if (!controller.isSlaveAvailable(mIndex)) {
--- a/core/include/forte/io/configFB/io_slave_multi.h
+++ b/core/include/forte/io/configFB/io_slave_multi.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 - 2018 fortiss GmbH
+ * Copyright (c) 2017 fortiss GmbH, Sichuan Qunyuan Technology Co., Ltd.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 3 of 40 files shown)


azoitl (commented):
All in all this looks good. I only have to points:

 - I think ' IOConfigHandlerFBMultiSlave'  should be called ' IOConfigHandlerFBMultiDevice' to show how we would like to have the io module in the future
 - can the parser also be implemented as non singleton. It is caching a lot of data and this makes me a bit nervous.

wzj7531:
Hi @azoitl,

Thank you for the review — glad the overall direction looks good to you.

### 1. `IOConfigHandlerFBMultiSlave` → `IOConfigHandlerFBMultiDevice`

I agree with the naming direction. Before renaming, I would like to briefly explain the design choice and ask for your preference.

The new class is largely modeled after the existing `IOConfigFBMultiSlave` (`io_slave_multi.h` / used e.g. by Embrick), but was added as a **separate** base class so we would not change behaviour for existing modules:

| | `IOConfigFBMultiSlave` | `IOConfigHandlerFBMultiSlave` (new) |
|---|---|---|
| Purpose | Existing multi-slave IO FB base (Embrick, …) | EtherCAT device handler FB base |
| Extra constructor | — | Generic FB constructor for `CGenFunctionBlock` |
| Key extension | — | Pure virtual `createSlaveHandler()` |
| Where used | Called from existing `handleInitEvent()` flow | Called in `handleInitEvent()` **before** `init()` (see `io_slave_multi_handler.cpp`) |

EtherCAT `ECDevice` / `ECModule` need to create an `ECBusDeviceHandler` / `ECModuleHandler` on the bus during MAP, which is why `createSlaveHandler()` was added in a dedicated base class rather than modifying `IOConfigFBMultiSlave` directly.

We see two options:

**Option A (my current preference):** Keep the separate base class and rename it to `IOConfigHandlerFBMultiDevice` (and the source files accordingly). `IOConfigFBMultiSlave` stays untouched for Embrick and other existing IO modules; EtherCAT uses the new device-oriented base.

**Option B:** Remove `IOConfigHandlerFBMultiSlave` and extend `IOConfigFBMultiSlave` instead (e.g. add `createSlaveHandler()` there, or refactor the shared init path). This would affect all users of `IOConfigFBMultiSlave` and needs broader regression testing (Embrick, etc.).

Could you let us know which option you prefer? If Option A is acceptable, I will proceed with the rename to `IOConfigHandlerFBMultiDevice`.

### 2. `EsiFileParser` singleton

Understood — I will refactor `EsiFileParser` to a **non-singleton** design. The parser currently caches ESI documents and device/module lookups globally, which is indeed a concern. I plan to bind parser lifetime to the bus/controller context instead of using `getInstance()`.

I will push follow-up commits for both items once we align on Option A vs B for the handler base class.

Thanks again!

wzj7531:
Hi @azoitl,

A quick follow-up on **Option B** from my previous comment — there is a way to extend `IOConfigFBMultiSlave` without breaking existing modules such as Embrick:

- Add `createSlaveHandler()` as a **non-pure** virtual member with an **empty default implementation** in `IOConfigFBMultiSlave`.
- Call it from the shared init path (same place as in our separate handler class today).
- Existing subclasses that do not need bus-level handler creation simply keep the default no-op.
- EtherCAT `ECDevice` / `ECModule` override `createSlaveHandler()` to create their `ECBusDeviceHandler` / `ECModuleHandler`.

That would allow us to **remove the separate `IOConfigHandlerFBMultiSlave` class**, rename `IOConfigFBMultiSlave` → `IOConfigHandlerFBMultiDevice` (or evolve the existing class in place toward device terminology), and avoid duplicating most of `io_slave_multi`.

Would this variant of Option B work for you, or do you still prefer **Option A** (keep a separate EtherCAT-specific base class, renamed to `IOConfigHandlerFBMultiDevice`)?

Happy to proceed either way once you confirm.

Thanks!

azoitl:
@wzj7531 thx for your detailed investigation and the options. I think your second option B would all in all be the better option for 4diac FORTE in the long run. It has less duplication and a single place. Also if in the future similar requirements as for Ethercat emerge we have a solution in place.

wzj7531:
Hi @azoitl,

Thank you for confirming **Option B** — both review points are now addressed in the latest push (`be31df5`):

### 1. Option B — merge into `IOConfigFBMultiSlave`

- Removed the separate `IOConfigHandlerFBMultiSlave` class
- Extended `IOConfigFBMultiSlave` with:
  - a default no-op `createSlaveHandler()` hook
  - the generic FB constructor needed by `CGenFunctionBlock`
- `handleInitEvent()` now calls `createSlaveHandler()` before `isSlaveAvailable()` / `checkSlaveType()`
- EtherCAT `ECDevice` / `ECModule` / `ECCoupler` override `createSlaveHandler()`; existing modules (Embrick, RevPi, …) keep the default no-op

### 2. `EsiFileParser` — non-singleton design

- Removed the singleton and all static caches
- Split into:
  - **`EsiCatalog`** — ESI index, lazy XML loading, separate device/module key maps
  - **`EsiIoConfigurator`** — PDO parsing, buffer sizing, handle registration
- Both are owned by `ECBusHandler` (created with the bus, released in `deInit()`)
- FBs access ESI via `bus.esiConfigurator()` instead of `getInstance()`

Build verified locally with `./compile.cmd ethercat` (including Embrick and RevPi compilation).

Let us know if anything else should be adjusted before merge.

Thanks!

core/include/forte/io/configFB/CMakeLists.txt
azoitl: when you change an existing file please add your company name after a comma to the copyright header.

e.g., copyright (c) 2025 Martin Erich Jobst, Sichuan Qunyuan Technology Co., Ltd.

core/src/io/configFB/io_slave_multi.cpp
azoitl: If you find a file with the old copyright header format please change it such that only the first your is there before you add your company name at the end.

core/src/io/configFB/io_slave_multi.cpp
```diff
@@ -9,6 +9,7 @@
  * Contributors:
  *   Johannes Messmer - initial API and implementation and/or initial documentation
  *   Jose Cabral - Cleaning of namespaces
+ *   Sichuan Qunyuan Technology Co., Ltd. - add createSlaveHandler() hook and generic FB constructor
```
azoitl: I noticed that I made a mistake in my last review. In this list your name has to be put as the person who did the changes. May I kindly ask you to also fix the other files.

azoitl (commented):
thx for the update this looks much better. I noticed that I made a mistake in providing feedback with the copyright headers (see below). Can you please fix them and clean the commit history.

wzj7531:
Hi @azoitl,

Thank you for the clarification on the copyright headers. I have pushed an update to this PR:

### Copyright / Contributors (this PR, #982)

- **Modified upstream files** (`io_slave_multi.h/.cpp`, core `CMakeLists.txt`):
  - Copyright: append `, Sichuan Qunyuan Technology Co., Ltd.`; use first year only where applicable (`2017 fortiss GmbH, …`)
  - Contributors: **Zijun Tang** (Eclipse account: tangzijun)
- **New EtherCAT IO files**: Contributors updated to **Zijun Tang**; Copyright remains `Sichuan Qunyuan Technology Co., Ltd.` as new-file contributions
- **PR #968 core files** touched for the same Contributors correction (header-only)

### Commit history

- Rebased onto current `develop` and **squashed to a single commit** (`8dc9152`)

### Already merged (#968)

- `modules/CMakeLists.txt` is fixed in a **separate follow-up PR** to `develop` (not rewriting #968 history): see the new PR from branch `fix-ethercat-copyright-headers`

Build verified locally with `./compile.cmd ethercat`.

Thanks!

azoitl (approved):

PR #983, opened 2026-07-08
Add Support for Monitoring Struct Members

m-meingast:


core/src/monitoring.cpp
```diff
@@ -130,6 +147,20 @@ namespace forte {
...
+                          TNameIdentifier &paPortNameList) {
+        for (auto nameListIt = paNameList.cend(); nameListIt != paNameList.cbegin(); nameListIt--) {
+          if (*nameListIt == paFB->getInstanceNameId()) {
+            paFBNameList.insert(paFBNameList.begin(), paNameList.begin(), nameListIt + 1);
+            paPortNameList.insert(paPortNameList.begin(), nameListIt + 1, paNameList.end());
+            return true;
+          }
+        }
+        return false;
+      }
```
mx990: I assume this is supposed to separate the path of the FB instance from the path inside of the FB? What if the FB instance has the same name as one of its containers (e.g., `Instance1.Instance1.DI1`)?

I would suggest to use `CFBContainer::getFB`, which provides the remaining path after looking up the FB instance, instead of the wrapper `CMonitoringHandler::getFB`.

core/src/monitoring.cpp
```diff
@@ -38,16 +38,41 @@ namespace forte {
...
+          paFullName.append(".");
+        }
+        paFullName.pop_back();
+      }
+
+      StringId getFullPortName(TNameIdentifier &paPortId) {
+        std::string name;
+        createFullFBName(name, paPortId);
+        return StringId::insert(name);
+      }
```
mx990: I believe we should change `CWatchEntry::mPortId` to `TNameIdentifier` instead of creating a combined string id here. That should allow for simpler lookups and also avoid bloating the string ids with potentially lots of permutations, particularly if we add support for arrays in the future.

mx990 (changes requested):

core/src/monitoring.cpp
m-meingast: Thanks for pointing this out. I had that already in mind while implementing this. I created an overload that uses the runner for splitting FB name an Port name

core/src/monitoring.cpp
m-meingast: This made some overall changes necessary, but looking at the issue with combining StringIds, this is the better solution.

core/include/forte/util/inplace_vector.h
```diff
@@ -211,6 +211,12 @@ namespace forte::util {
   bool operator!=(const inplace_vector<T, U> &paFirst, const inplace_vector<T, V> &paSecond) {
     return !(paFirst == paSecond);
   }
+
+  template<typename T, std::size_t U, std::size_t V>
+  [[nodiscard]]
+  bool operator<(const inplace_vector<T, U> &paFirst, const inplace_vector<T, V> &paSecond) {
```
mx990: I would recommend overloading [`operator<=>`](https://en.cppreference.com/cpp/language/operator_comparison#Three-way_comparison) instead, which will provide the remaining comparison operators.

core/src/monitoring.cpp
```diff
@@ -29,16 +31,25 @@ using namespace std::string_literals;
...
           [](const SFBMonitoringEntry &paItem, const CFunctionBlock *const paFB) { return &paItem.getFB() < paFB; };
 
-      void addDataWatch(SFBMonitoringEntry &paFBMonitoringEntry, StringId paPortId, CIEC_ANY &paDataVal) {
+      void createFullFBName(std::string &paFullName, TNameIdentifier &paNameList) {
+        for (const auto &runner : paNameList) {
+          paFullName.append(runner);
+          paFullName.append(cgNameDelimiter);
+        }
+        paFullName.pop_back();
+      }
```
mx990: I would recommend using `forte::util::join` from `string_utils.h` instead, for an example see:
https://github.com/eclipse-4diac/4diac-forte/blob/7406b67fe256618464e63f66c235302821d5d3c5/core/src/resource.cpp#L88

core/src/monitoring.cpp
```diff
@@ -84,9 +96,16 @@ namespace forte {
...
         paResponse += "<Port name=\""s;
-        paResponse += paPortId;
+        bool first = true;
+        for (const auto &id : paPortId) {
+          if (!first) {
+            paResponse += cgNameDelimiter;
+          }
+          paResponse += id;
+          first = false;
+        }
```
mx990: I would again recommend using `forte::util::join` from `string_utils.h` instead.

core/src/monitoring.cpp
```diff
@@ -168,29 +178,40 @@ namespace forte {
     return mResource.getFB(runner, paNameList.cend());
   }
 
+  CFunctionBlock *CMonitoringHandler::getFB(TNameIdentifier &paNameList,
+                                            TNameIdentifier &paFBNameList,
+                                            TNameIdentifier &paPortNameList) {
```
mx990: Can we use `std::span<const StringId>` for the split lists, so that we do not need two full separate lists for performing lookups?

mx990 (changes requested):

core/include/forte/util/inplace_vector.h
m-meingast: Done

core/src/monitoring.cpp
m-meingast: Done

core/src/monitoring.cpp
m-meingast: Done

core/src/monitoring.cpp
m-meingast: I changed the parameters to `std::span<const StringId>&`, but since `TNameIdentifier` is needed in most of the following functions, I need to convert the spans back as soon as they are used.

mx990: I would suggest to also change the `{add,remove}{Event,Data}Watch` to accept a `std::span<const StringId>` and add an `operator<=>` for `TNameIdentifier` and `std::span<const StringId>` to perform the lookups using spans. That should only leave the constructors of the watches where you need to convert from spans to `TNameIdentifier`.

mx990 (approved):

azoitl (approved):

PR #1008, opened 2026-08-06
Ensure unique IDs for adapter instances to prevent event overlap

azoitl:
Independent of plug or socket adapter instances need to be given a FB unique id. Otherwise input or output events of different adpaters will overlap and lead to wrong behavior. This was a problem in several IO modules.

FYI @wzj7531.

### DIFF
--- a/modules/embrick/types/Slave.cpp
+++ b/modules/embrick/types/Slave.cpp
@@ -33,7 +33,7 @@ namespace forte::eclipse4diac::io::embrick {
       IOConfigFBMultiSlave(
           paSlaveConfigurationIO, paSlaveConfigurationIO_num, paType, paContainer, paInterfaceSpec, paInstanceNameId),
       var_BusAdapterIn("BusAdapterIn"_STRID, *this, 0),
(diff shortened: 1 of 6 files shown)


moniwe (approved):
I had to remove members within the WagoBusAdapter that it worked again, possibly also needed for other BusAdapters, as RevolutionPi, ...

```
static const TEventID scmEventINITOID = 0;
static const TEventID scmEventINITID = 0;

void setInitialValues() override;
 
CIEC_BOOL var_QO;
CIEC_BOOL var_QI;
CIEC_UINT var_MasterId;
CIEC_UINT var_Index;

TEventID evt_INITO() {
  return getParentAdapterListEventID() + scmEventINITOID;
}

TEventID evt_INIT() {
  return getParentAdapterListEventID() + scmEventINITID;
}
```

azoitl:
there seems to be an issue with the codeql service. As the rest is working I would merge this.

@moniwe can you please provide a PR for your other changes to Wago KBus IOs?

PR #1014, opened 2026-08-13
Add optional OPC UA over WebSocket (opc.ws://, opc.wss://) server endpoints

franz-hoepfinger-4diac:
See also:

- https://github.com/franz-hoepfinger-4diac-org/4diac-forte/pull/33
- https://github.com/eclipse-4diac/4diac-forte/discussions/1002

Introduces FORTE_COM_OPC_UA_WEBSOCKET (default OFF), which registers
opc.ws:// and/or opc.wss:// endpoints alongside the plain opc.tcp:// one,
one and two ports above it respectively. On platforms built with
UA_ENABLE_LWS (libwebsockets, POSIX-only), the endpoints are backed by
open62541's LWS-based WebSocket ConnectionManager
(UA_ConnectionManager_new_LWS_WebSocket) and support both opc.ws:// and
opc.wss://. On RTOS targets without libwebsockets (e.g. ESP-IDF), a
platform-provided ConnectionManager can be plugged in via
FORTE_COM_OPC_UA_WEBSOCKET_CONNECTIONMANAGER_DIR instead, supporting
opc.ws:// only. FORTE_COM_OPC_UA_WEBSOCKET_INSECURE/_SECURE gate the two
transports independently; INSECURE defaults to OFF so enabling WebSocket
support doesn't implicitly expose an unencrypted endpoint.

The WebSocket support (UA_ServerConfig::webSocketEnabled and friends,
UA_ENABLE_LWS) is a patch on top of upstream open62541, not present in
stock releases such as the "1.5.4" pulled in by the find_package()
fallback. Guard this with a configure-time capability check instead of
letting it fail with a confusing compiler error deep in
opcua_local_handler.cpp: check the relevant UA_ServerConfig members via
CMAKE_REQUIRED_INCLUDES rather than linking against open62541::open62541,
since FORTE_COM_OPC_UA_INCLUDE_DIR mode is headers-only by design (the
library is compiled and linked one level up, e.g. as its own ESP-IDF
component) and a link-based check would always spuriously fail there.
Also reject the default SECURE=ON/INSECURE=OFF combination when
UA_ENABLE_LWS isn't available, since opc.wss:// has no non-LWS backend and
that combination would otherwise silently register no WebSocket endpoint
at all.

The server startup log now reflects the endpoints configureUAServer()
actually registered (read back from UA_ServerConfig::serverUrls after
configuration, e.g. dropping opc.ws:// if the WebSocket ConnectionManager
failed to register, or opc.wss:// if no certificate/private key were
configured) and is only emitted once the server has actually started
listening, rather than being derived from the configured port and
compile-time flags before any of that runs.

opcua_local_handler.h now includes opcua_defaults.h directly: that header
is also reached via stdfblib/system/src/OPCUA_MGR.h in a different CMake
target than forte-com-opc_ua, and without it that target would see
FORTE_COM_OPC_UA_WEBSOCKET as undefined while opcua_local_handler.cpp sees
it defined -- an ODR violation (COPC_UA_Local_Handler with and without
mWsConnectionManager) that had the constructor writing past the end of an
object allocated using the smaller, member-less layout.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/com/opc_ua/CMakeLists.txt
+++ b/com/opc_ua/CMakeLists.txt
@@ -1,16 +1,17 @@
 # *******************************************************************************************
 
-# Copyright (c) 2015-2016 Florian Froschermeier
-# <florian.froschermeier@tum.de>, fortiss GmbH This program and the
-# accompanying materials are made available under the terms of the Eclipse
-# Public License 2.0 which is available at
+# Copyright (c) 2015 Florian Froschermeier
+# <florian.froschermeier@tum.de>, fortiss GmbH, HR Agrartechnik GmbH. This
+# program and the accompanying materials are made available under the terms
+# of the Eclipse Public License 2.0 which is available at
 # http://www.eclipse.org/legal/epl-2.0.
 #
 # SPDX-License-Identifier: EPL-2.0
 #
 # Contributors: Florian Froschermeier - initial integration of OPC Unified
-# Architecture into FORTE Stefan Profanter - refactoring and adaption to new
-# concept
+# Architecture into FORTE Stefan Profanter - refactoring and adaptation to new concept;
+# Franz Höpfinger - Add optional OPC UA over WebSocket (insecure and secure
+# transports) server endpoint
 # ******************************************************************************************
 
 # ############################################################################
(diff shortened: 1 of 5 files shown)


franz-hoepfinger-4diac:
Ready for Review.

azoitl:
Can you please not submit half backed ideas. To reduce noise for other Eclipse 4diac members I'm closing this PR for now. Feel free to resubmit when you and open62541 are ready that this can be evaluated by a broader audience.

franz-ms-muc:
I want to point out that i using this already in prodution, 
so this is not half-baked. 
you will need open62541 v1.6, that is clear.

azoitl:
> I want to point out that i using this already in prodution, so this is not half-baked. you will need open62541 v1.6, that is clear.

If you constantly update a PR it is a clear sign for not being ready!

franz-ms-muc:
> > I want to point out that i using this already in prodution, so this is not half-baked. you will need open62541 v1.6, that is clear.
> 
> If you constantly update a PR it is a clear sign for not being ready!

https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/33 answering the Review Comments. not more.  
lets wait for open62541 v1.6 then i re-submit this PR.

PR #1024, opened 2026-08-28
fix(opc_ua): fix UA_ServerConfig aggregate-init clobbering context

franz-ms-muc:
UA_ServerConfig config{ config.logging = &getLogger() } is not a
designated initializer (missing the leading dot): it evaluates the
assignment expression config.logging = &getLogger() and uses its
result as the positional initializer for UA_ServerConfig's first
member, which is context, not logging. The assignment itself does set
config.logging correctly as a side effect, but config.context -- which
should stay null -- ends up aliased to the logger pointer instead.

Use a proper designated initializer for .logging so context is left at
its zero-initialized default, as intended.


mx990 (changes requested):
Please remove the conventional commit prefix `fix(opc_ua):`.

mx990 (approved):

PR #1026, opened 2026-08-28
Fix CommandParser::parseIdentifier() dropping a Connection Source's trailing empty port name

franz-hoepfinger-4diac:
A Connection's Source ending in '.' means its final segment (the port
name) is itself the empty string. That's valid: IEC 61131-3 Functions
use the empty string as the name of their return-value output port
(e.g. ASSEMBLE_BYTE_FROM_BOOLS), so a Connection sourced from such an
instance's return value legitimately looks like
"App.functions.RotaCut.ASSEMBLE_BYTE_FROM_BOOLS." on the wire.

parseIdentifier()'s loop tested "while (!paIdentifierString.empty())"
*before* checking for a final dot-less segment, so consuming the
trailing '.' left an empty string that exited the loop before that
last (empty) segment was ever pushed, falling through to a bare
"return BadParams". Observed on real hardware as "Boot file error.
DEV_MGR says error is BAD_PARAMS" for a CREATE Connection whose source
ends at such a pin.

Adds an opt-in paAllowEmptyLastSegment parameter (default false, so
FB instance names and Connection Destinations -- which must always
have a real final segment -- keep their existing behavior unchanged)
and passes true only for a Connection's Source in
parseConnectionContent, which is shared by both CREATE and DELETE.

Updates the existing createConnectionDotAtEndOfSource unit test
(renamed createConnectionSourceEndsWithEmptyPortName) to reflect the
corrected behavior, and adds a symmetric
deleteConnectionSourceEndsWithEmptyPortName test. createFBDotAtEnd and
createConnectionDotAtEndOfDestination are intentionally unchanged.

Also adds negative tests covering every other place parseIdentifier()
is used with a trailing '.', confirming it is still rejected as
BadParams there: DELETE's FB Name and Connection Destination, Watch
Destination (CREATE and DELETE, which share parseConnectionContent),
and WRITE's Destination.

Verified: full forte_test suite passes (posix_ws build,
FORTE_COM_OPC_UA=OFF).

Reference:

- https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/44

### DIFF
--- a/stdfblib/system/src/CommandParser.cpp
+++ b/stdfblib/system/src/CommandParser.cpp
@@ -9,6 +9,8 @@
  * Contributors:
  *   Jose Cabral
  *    - initial API and implementation and/or initial documentation
+ *   Franz Höpfinger
+ *    - fix parseIdentifier() dropping a trailing empty name segment
  *******************************************************************************/
 
 #include "CommandParser.h"
@@ -272,7 +274,9 @@ namespace forte::iec61499::system {
     return EMGMResponse::Ready;
   }
 
-  EMGMResponse CommandParser::parseIdentifier(std::string_view paIdentifierString, TNameIdentifier &paIdentifier) {
+  EMGMResponse CommandParser::parseIdentifier(std::string_view paIdentifierString,
+                                              TNameIdentifier &paIdentifier,
+                                              bool paAllowEmptyLastSegment) {
     while (!paIdentifierString.empty()) {
       auto pos = paIdentifierString.find('.');
       if (pos == std::string_view::npos) {
@@ -288,6 +292,18 @@ namespace forte::iec61499::system {
         return EMGMResponse::Overflow;
       }
       paIdentifierString = paIdentifierString.substr(pos + 1);
+      if (paIdentifierString.empty()) {
+        // Trailing '.': the final segment is the empty string. Only valid where explicitly
+        // allowed (a Connection's Source -- see the parameter doc in the header); an FB
+        // instance name or a Connection's Destination must always have a real final segment.
+        if (!paAllowEmptyLastSegment) {
+          return EMGMResponse::BadParams;
+        }
+        if (!paIdentifier.try_push_back(StringId::insert(std::string_view{}))) {
+          return EMGMResponse::Overflow;
+        }
+        return EMGMResponse::Ready;
+      }
     }
     return EMGMResponse::BadParams;
   }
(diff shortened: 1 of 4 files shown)


stdfblib/system/src/CommandParser.cpp
```diff
@@ -272,7 +274,12 @@ namespace forte::iec61499::system {
     return EMGMResponse::Ready;
   }
 
-  EMGMResponse CommandParser::parseIdentifier(std::string_view paIdentifierString, TNameIdentifier &paIdentifier) {
+  EMGMResponse CommandParser::parseIdentifier(std::string_view paIdentifierString,
+                                              TNameIdentifier &paIdentifier,
+                                              bool paAllowEmptyLastSegment) {
+    if (paIdentifierString.empty()) {
```
azoitl: please remove this if. This does not add value and complicates the method.

stdfblib/system/src/CommandParser.cpp
```diff
@@ -612,12 +631,13 @@ namespace forte::iec61499::system {
   }
 
   void CommandParser::appendIdentifierName(CIEC_STRING &paDest, TNameIdentifier &paIdentifier) {
-    if (!paIdentifier.empty()) {
-      for (const auto &runner : paIdentifier) {
-        paDest.append(runner.data());
+    bool first = true;
```
azoitl: I don't see why the change in this method is needed for the problem described in the PR.

azoitl (changes requested):
As we are in feature freeze phase and this is a bug report please move this commit and PR to the 3.3.x branch.

The commit message is way to long. The current description is suitable for a PR where there was no issue created first but not for a commit message. Please shorten to the core what is changed. 

Please add also negative tests showing that other places where identifiers may be parsed do not allow a `.` at the end.

stdfblib/system/src/CommandParser.cpp
franz-hoepfinger-4diac: Good point, removed — the while loop already falls through to "return EMGMResponse::BadParams;" for an empty string, so the guard was redundant. Fixed in 78eb9d555.

stdfblib/system/src/CommandParser.cpp
franz-hoepfinger-4diac: Fair point — this wasn't needed for the BAD_PARAMS bug this PR fixes. Split it out into #1031.

franz-hoepfinger-4diac:
Thanks for the review, @azoitl — all three addressed:

- **Base branch**: this PR already targets `3.3.x`.
- **Commit message**: shortened to just the core parseIdentifier() fix (`78eb9d555`).
- **Negative tests**: added for every other place `parseIdentifier()` is called with a trailing `.`, confirming it's still rejected as `BadParams`: DELETE's FB Name and Connection Destination, Watch Destination (CREATE and DELETE, which share `parseConnectionContent`), and WRITE's Destination.

Also, per your other two inline comments: removed the redundant empty-string check, and split the `appendIdentifierName()` fix out into #1031 since it wasn't needed for the bug this PR fixes.

azoitl (approved):

PR #1031, opened 2026-08-31
Fix CommandParser::appendIdentifierName() emitting a duplicated final segment

franz-hoepfinger-4diac:
Split off from #1026 per review feedback (the fix isn't needed for the
BAD_PARAMS bug that PR addresses).

appendIdentifierName() appended a separator after every segment,
including the last, then appended the last segment a second time. For
an ordinary hierarchical name this duplicated the final segment in
generated READ responses, e.g. Source="SubApp1.Trigger.DT" was echoed
back as "SubApp1.Trigger.DT.DT".

Join segments with a separator only between them instead.

Verified: full forte_test suite passes (posix_ws build,
FORTE_COM_OPC_UA=OFF).

Reference: 

- https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/48

### DIFF
--- a/stdfblib/system/src/CommandParser.cpp
+++ b/stdfblib/system/src/CommandParser.cpp
@@ -10,6 +10,8 @@
  *   Jose Cabral
  *    - initial API and implementation and/or initial documentation
  *   Franz Höpfinger
+ *    - fix appendIdentifierName() emitting a duplicated/extra final segment
+ *   Franz Höpfinger
  *    - fix parseIdentifier() dropping a trailing empty name segment
  *******************************************************************************/
 
@@ -628,12 +630,15 @@ namespace forte::iec61499::system {
   }
 
   void CommandParser::appendIdentifierName(CIEC_STRING &paDest, TNameIdentifier &paIdentifier) {
-    if (!paIdentifier.empty()) {
-      for (const auto &runner : paIdentifier) {
-        paDest.append(runner.data());
-        paDest.append(".");
-      }
-      paDest.append(paIdentifier.back().data());
+    auto it = paIdentifier.begin();
+    if (it == paIdentifier.end()) {
+      return;
+    }
+
+    paDest.append(it->data());
+    for (++it; it != paIdentifier.end(); ++it) {
+      paDest.append(".");
+      paDest.append(it->data());
     }
   }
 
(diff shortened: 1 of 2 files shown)


tests/stdfblib/system/commandparser_test.cpp
```diff
@@ -736,6 +736,28 @@ namespace forte::iec61499::system::test {
...
+  // the last) and then appended the last segment a second time, duplicating it in READ responses
+  // (e.g. "SubApp1.Trigger.DT" became "SubApp1.Trigger.DT.DT").
+  BOOST_AUTO_TEST_CASE(readConnectionResponseDoesNotDuplicateLastSourceSegment) {
+    SManagementCMD cmd;
+    CommandParser parser(cmd);
+
+    auto response = parser.parseMGMCommand(
+        "", R"(<Request ID="1" Action="READ"><Connection Source="SubApp1.Trigger.DT" Destination="*" /></Request>)");
+
+    BOOST_TEST(response == EMGMResponse::Ready);
```
azoitl: this is tested somewhere else. Please remove the checks for parsing.

stdfblib/system/src/CommandParser.cpp
```diff
@@ -612,12 +614,13 @@ namespace forte::iec61499::system {
   }
 
   void CommandParser::appendIdentifierName(CIEC_STRING &paDest, TNameIdentifier &paIdentifier) {
-    if (!paIdentifier.empty()) {
-      for (const auto &runner : paIdentifier) {
-        paDest.append(runner.data());
+    bool first = true;
```
azoitl: I find this first helper variable very ugly and it leads to hard to read code. I like something like the following more. WDYT?

 auto it = paIdentifier.begin();
  if (it == paIdentifier.end()) return;

  paDest.append(it->data());
  for (++it; it != paIdentifier.end(); ++it) {
    paDest.append(".");
    paDest.append(it->data());
  }

azoitl (changes requested):

tests/stdfblib/system/commandparser_test.cpp
franz-hoepfinger-4diac: Good point, removed — the response==Ready and nameIdentifierEquals checks duplicated existing parsing coverage (e.g. readConnectionWithWildcardDestination). Kept only the serialization check this test exists for. Fixed in d91e99537.

stdfblib/system/src/CommandParser.cpp
franz-hoepfinger-4diac: Agreed, that's cleaner — applied as suggested. Fixed in e31ea6450.

azoitl (approved):

Issue #1038, opened 2026-09-07
OPC UA Client: createSubscription() creates a new subscription per action instead of reusing one, leaking subscriptions and causing BadMonitoredItemIdInvalid on teardown


franz-hoepfinger-4diac:
**Component:** `com/opc_ua/src/opcua_client_information.cpp`

**Summary**

`CUA_ClientInformation::createSubscription()` has no guard against an already-existing subscription and unconditionally creates a brand new OPC UA subscription on the server every time it is called:

```cpp
bool CUA_ClientInformation::createSubscription() {
  UA_CreateSubscriptionRequest request = UA_CreateSubscriptionRequest_default();
  request.requestedPublishingInterval = FORTE_COM_OPC_UA_CLIENT_PUB_INTERVAL;
  UA_CreateSubscriptionResponse response = UA_Client_Subscriptions_create(
      mClient, request, this, nullptr, CUA_RemoteCallbackFunctions::deleteSubscriptionCallback);
  if (UA_STATUSCODE_GOOD == response.responseHeader.serviceResult) {
    ...
    mSubscriptionInfo.mSubscriptionId = response.subscriptionId;
    return true;
  }
  ...
}
```

It is called from `initializeSubscription()` **once per action/FB** (`opcua_client_information.cpp:442`), but `CUA_ClientInformation` represents a single client connection **per remote endpoint** and accumulates monitored items for *all* subscribe actions on that endpoint into one shared `mSubscriptionInfo.mMonitoredItems` vector.

**Effect:** for N `SUBSCRIBE` FBs pointed at the same remote endpoint, N separate subscriptions get created on the server, but `mSubscriptionInfo.mSubscriptionId` is overwritten to the newest one each time. The N-1 earlier subscriptions:
- are never explicitly deleted (leaked on the server), and
- their monitored items remain tracked client-side under the *current* (mismatched, newer) subscription id.

When the resource is killed/redeployed, `uninitializeSubscribeAction()` calls `UA_Client_MonitoredItems_deleteSingle(mClient, mSubscriptionInfo.mSubscriptionId, item->mMonitoringItemId)` for every tracked item, but `mSubscriptionInfo.mSubscriptionId` only ever refers to the last-created subscription - so for every item that actually belongs to an earlier (already-orphaned) subscription, the server correctly reports `BadMonitoredItemIdInvalid` since that item id doesn't exist under that subscription.

**Observed in practice**

[...]

PR #1040, opened 2026-09-07
Add regression test reproducing OPC UA client subscription leak

franz-hoepfinger-4diac:
## Summary

Adds a failing regression test for the bug reported in #1038: `CUA_ClientInformation::createSubscription()` always creates a new OPC UA subscription instead of reusing an existing one when several `SUBSCRIBE_1` actions target the same remote endpoint, leaking every earlier subscription on the server and breaking teardown of their monitored items.

This PR is test-first and deliberately stays red - the fix is in the follow-up PR (to be linked once opened), fixing #1038.

Already staged and verified on our own fork: [Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte#51](https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/51) (red) / [#52](https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/52) (green), including CI runs on that fork.

## Test plan

- [x] `OPCUA_KillMultiSubscribe` systemtest added: two `SUBSCRIBE_1` FBs against one remote endpoint, each explicitly de-initialized after connecting.
- [x] Verified locally (Linux, open62541 v1.5.4, same flags as CI) and in our fork's CI: fails with `BadMonitoredItemIdInvalid`, reproducing the bug.


azoitl:
I would prefer to close this pr as alone it makes no sense and the commit is anyhow also part of #1041.

franz-ms-muc:
> I would prefer to close this pr as alone it makes no sense and the commit is anyhow also part of #1041.

YES, i wanted to show: the TEST is  working: it beomes RED, and after #1041 it gets green again. I guess this is always the best way to proof the Test is actually testing the Problem which is fixed after, right ? 

#1040 can be closed unmerged. 


BTW: the WINDOWS Build "[CMake on multiple platforms / build (windows-latest, Debug, cl) (pull_request)](https://github.com/eclipse-4diac/4diac-forte/actions/runs/34124527994/job/101749954002?pr=1040)Successful in 16m" does pass, as it has no OPC-UA inside enabled at all.

Issue #1039, opened 2026-09-07
OPC UA client: actions are never re-initialized after an unsolicited (server-side) subscription deletion


franz-hoepfinger-4diac:
**Component:** `com/opc_ua/src/opcua_client_information.cpp`

**Summary**

`CUA_ClientInformation::deleteSubscriptionCallback()` fires whenever open62541 removes its internal representation of a subscription - both for our own explicit deletion request and for an *unsolicited* server-side deletion (e.g. the subscription's keepalive/lifetime expires, or the server restarts/loses the subscription for its own reasons) while the client-side actions are still active and expected to keep receiving data.

That callback calls `resetSubscription(false)`, which clears `mSubscriptionInfo`'s tracked subscription id and monitored items (see #52 for that part), but does not requeue the affected, already-initialized `CActionInfo` objects into `mActionsToBeInitialized`. `handleClientState()` only (re-)initializes actions present in that list, and nothing else calls it in response to this callback. Consequently, after an unsolicited server-side subscription deletion, every action that was subscribed through it stays silently marked as "initialized" from the client's perspective, but has no subscription or monitored items on the server anymore - it will never receive another update unless some unrelated event (e.g. a full client disconnect/reconnect cycle) happens to reinitialize it.

**This predates #52** - in the code before that PR, `resetSubscription(false)` did nothing at all, so the same actions were equally stuck, just with additionally-stale (non-zero) subscription-id bookkeeping. #52 makes the internal state consistent again but does not address the missing requeue, since it targets a different bug (duplicate subscription creation on kill/redeploy) than this one (recovery after unsolicited deletion).

**Suggested fix**

In the `resetSubscription(false)` path (or in `deleteSubscriptionCallback()` itself), requeue the actions that were tracked against the deleted subscription back into `mActionsToBeInitialized` so `handleClientState()` recreates their subscription and monitored items on its next iteration.

**Test plan**

A regression test for this needs a way to make the *server* delete the subscription out from under an otherwise-idle client (e.g. a server that closes/kills a subscription mid-test, or a short subscription lifetime configured to expire), then assert the client eventually re-establishes it and resumes receiving updates - this is more involved than a typical systemtest boot-file scenario and will need some thought on how to trigger the unsolicited deletion deterministically.

PR #1041, opened 2026-09-07
Fix OPC UA client subscription reuse across actions

### LINKED ISSUE #1038: OPC UA Client: createSubscription() creates a new subscription per action instead of reusing one, leaking subscriptions and causing BadMonitoredItemIdInvalid on teardown

franz-hoepfinger-4diac:
## Summary

Fixes #1038. Stacked on #1040 (which stays red on its own - the regression test); this PR adds the actual fix and turns `OPCUA_KillMultiSubscribe` green.

`createSubscription()` never checked whether `CUA_ClientInformation` already had a subscription, so every `SUBSCRIBE_1` action sharing one client/endpoint created its own. Only the latest subscription id stayed tracked in `mSubscriptionInfo.mSubscriptionId`, leaking every earlier subscription on the server and breaking teardown of their monitored items (`BadMonitoredItemIdInvalid`).

## Changes

- `createSubscription()`: return early if a subscription id is already tracked, reusing it instead of creating a new one.
- `resetSubscription()`: only clear the tracked subscription id/monitored items from the `deleteSubscriptionCallback()` path (i.e. once open62541 has actually confirmed the subscription's removal), not unconditionally on every delete attempt - a failed delete request no longer discards a subscription id that may still be valid server-side.

## Test plan

- [x] `OPCUA_KillMultiSubscribe` (from #1040) now passes locally (Linux, open62541 v1.5.4, same flags as CI).
- [x] Also staged and verified green on our own fork: [Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte#52](https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/52), including CI.

Once #1040 merges, this PR's base will need to move to `develop` (already targets it directly since the branch is stacked on the fork side).

A related but distinct gap (actions are never re-initialized after an *unsolicited* server-side subscription deletion) is out of scope here and tracked separately in #1039.


com/opc_ua/src/opcua_client_information.cpp
```diff
@@ -560,7 +565,20 @@ namespace forte::com_infra::opc_ua {
...
+        // let a future createSubscription() create a duplicate for a subscription that is
+        // still alive. On success, open62541 invokes our deleteCallback synchronously from
+        // within UA_Client_Subscriptions_deleteSingle() above, which already reached the
+        // paDeleteSubscription == false branch below and cleared the tracked state.
       }
+    } else {
+      // Reached only from deleteSubscriptionCallback(): the client library has already
+      // confirmed and processed the subscription's removal (our own explicit delete above,
+      // or an unsolicited server-side deletion), so anything we were tracking for it is
+      // stale now.
```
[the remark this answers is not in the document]
franz-hoepfinger-4diac: Fixed: `resetSubscription()` no longer relies on the deletion callback firing synchronously. On a successful `UA_Client_Subscriptions_deleteSingle()` call it now clears the tracked state and accounts for the completed async call directly, and both that path and the `deleteSubscriptionCallback()` path are guarded on `mSubscriptionInfo.mSubscriptionId != 0`, so whichever runs first (sync or async callback) does the work exactly once and the other becomes a no-op. This also fixes a latent double-decrement of `mMissingAsyncCalls` that existed when the callback fired synchronously. See commit ab3b132a7.

systemtests/src/modules/opc_ua/killMultiSubscribeClient.fboot
```diff
@@ -0,0 +1,32 @@
+;<Request ID="1" Action="CREATE"><FB Name="killMultiSubscribe" Type="iec61499::system::EMB_RES" /></Request>
+killMultiSubscribe;<Request ID="2" Action="CREATE"><FB Name="Subscribe_A" Type="iec61499::net::SUBSCRIBE_1" /></Request>
+killMultiSubscribe;<Request ID="3" Action="CREATE"><FB Name="Subscribe_B" Type="iec61499::net::SUBSCRIBE_1" /></Request>
+killMultiSubscribe;<Request ID="4" Action="CREATE"><FB Name="SR_A" Type="iec61499::events::E_SR" /></Request>
+killMultiSubscribe;<Request ID="5" Action="CREATE"><FB Name="SR_B" Type="iec61499::events::E_SR" /></Request>
+killMultiSubscribe;<Request ID="6" Action="CREATE"><FB Name="E_DELAY_KILL_A" Type="iec61499::events::E_DELAY" /></Request>
+killMultiSubscribe;<Request ID="7" Action="WRITE"><Connection Source="T#10s" Destination="E_DELAY_KILL_A.DT" /></Request>
+killMultiSubscribe;<Request ID="8" Action="CREATE"><FB Name="E_DELAY_KILL_B" Type="iec61499::events::E_DELAY" /></Request>
+killMultiSubscribe;<Request ID="9" Action="WRITE"><Connection Source="T#10s" Destination="E_DELAY_KILL_B.DT" /></Request>
```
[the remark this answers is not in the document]
franz-hoepfinger-4diac: Fixed: staggered the two teardown timers so `E_DELAY_KILL_A.DT` fires at `T#10s` and `E_DELAY_KILL_B.DT` at `T#11s`, avoiding the simultaneous-firing race while still exercising two independent `SUBSCRIBE_1` teardown lifecycles.

com/opc_ua/src/opcua_client_information.cpp
```diff
@@ -560,7 +565,20 @@ namespace forte::com_infra::opc_ua {
         DEVLOG_ERROR("[OPC UA CLIENT]: Couldn't delete subscription %u. Failed with error %s. No further actions will "
                      "be taken\n",
                      mSubscriptionInfo.mSubscriptionId, UA_StatusCode_name(retval));
+        // The subscription may still exist server-side (e.g. the request itself failed),
+        // so mSubscriptionInfo is intentionally left untouched here - clearing it now would
+        // let a future createSubscription() create a duplicate for a subscription that is
+        // still alive. On success, open62541 invokes our deleteCallback synchronously from
+        // within UA_Client_Subscriptions_deleteSingle() above, which already reached the
+        // paDeleteSubscription == false branch below and cleared the tracked state.
```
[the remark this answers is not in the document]
franz-hoepfinger-4diac: Fixed: the comment now refers to the actual function name, `deleteSubscriptionCallback()`, instead of the shorthand `deleteCallback`.

franz-ms-muc:
@azoitl many times sorry for pushing again, but: the downstream Review found another, valid issue wich i fixed: 
https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/52#discussion_r3959244461

azoitl:
@franz-ms-muc as you are on it. Please tripple check the commit messages and headlines. I think both overlap to much. Also I find the comments in the code a bit strange.

com/opc_ua/src/opcua_client_information.cpp
```diff
@@ -553,13 +562,39 @@ namespace forte::com_infra::opc_ua {
...
+      }
+      // The delete request succeeded: account for the completed async call and clear our
+      // tracked state here explicitly, instead of depending on deleteSubscriptionCallback()
+      // being invoked synchronously from within UA_Client_Subscriptions_deleteSingle() above.
+      // Guarded so nothing is double-counted if that callback already ran synchronously (see
+      // the matching guard in the else branch below).
+      if (0 != mSubscriptionInfo.mSubscriptionId) {
+        removeAsyncCall();
+        mSubscriptionInfo.mSubscriptionId = 0;
+        mSubscriptionInfo.mMonitoredItems.clear();
```
franz-hoepfinger-4diac: we have here a 3 line Block which is IDENTICAL to the next

com/opc_ua/src/opcua_client_information.cpp
```diff
@@ -553,13 +562,39 @@ namespace forte::com_infra::opc_ua {
...
+    } else {
+      // Reached from deleteSubscriptionCallback(): the client library has confirmed the
+      // subscription's removal (our own explicit delete above, or an unsolicited server-side
+      // deletion). Guarded so this stays correct and idempotent whether open62541 invokes the
+      // callback synchronously (in which case the branch above already handled it) or
+      // asynchronously (in which case this is the first and only place that handles it).
+      if (0 != mSubscriptionInfo.mSubscriptionId) {
+        removeAsyncCall();
+        mSubscriptionInfo.mSubscriptionId = 0;
+        mSubscriptionInfo.mMonitoredItems.clear();
```
franz-hoepfinger-4diac: to this block. I will merge them.

franz-hoepfinger-4diac: done

com/opc_ua/src/opcua_client_information.cpp
```diff
@@ -474,6 +474,10 @@ namespace forte::com_infra::opc_ua {
   }
 
   bool CUA_ClientInformation::createSubscription() {
+    if (0 != mSubscriptionInfo.mSubscriptionId) { // already have a subscription, all actions share it
```
gravemalte: Please test the member again 0 an not the other way around. We want to avoid Yoda expressions

com/opc_ua/src/opcua_client_information.cpp
```diff
@@ -553,15 +558,27 @@ namespace forte::com_infra::opc_ua {
...
+        // clearing it now could let a future createSubscription() create a duplicate.
+        return;
       }
     }
+
+    // Reached either on a successful delete above, or from deleteSubscriptionCallback() (our own
+    // delete's confirmation, or an unsolicited server-side deletion). Guarded so this runs
+    // exactly once even if open62541 invokes that callback synchronously from within
+    // UA_Client_Subscriptions_deleteSingle() above.
+    if (0 != mSubscriptionInfo.mSubscriptionId) {
```
gravemalte: Same here.

gravemalte (changes requested):

PR #1045, opened 2026-09-09
Add missing find_dependency for open62541 in cmake config file

mx990:
The CMake config file was missing a `find_dependency` macro to provide the required `open62541` dependency for users of forte when OPC UA is enabled.

This adds the missing `find_dependency` call for `open62541` to `forte-config.cmake.in`.


com/opc_ua/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ if (NOT FORTE_COM_OPC_UA)
     return()
 endif ()
 
+# keep in sync with forte-config.cmake.in
```
ernstblechaPT: The fact that this comment is needed feels wrong; but if this is the only way it should be in all three places that need to be synced:
 - com/opc_ua/CMakeLists.txt
 - forte-config.cmake.in
 - dependencies.spdx

mx990: This seems to be the idiomatic way to handle this in CMake. I have added respective comments in the other locations.

azoitl (approved):

PR #1046, opened 2026-09-13
Wago 750-636 register communication switch does not always work

moniwe:
Sometimes switching to register communication does not work. This workaround allows the user to trigger it manually during monitoring mode


modules/wagokbus/WagoDeviceController.cpp
```diff
@@ -250,11 +255,21 @@ namespace forte::eclipse4diac::io::wago {
     }
   }
 
+  bool WagoDeviceController::regComInitFailed(){
+    util::CCriticalRegion criticalRegion(mRegComMutex);
+    if (mReadCounter == -1) {
+      return true;
+    } else {
+      return false;
+    }
```
MartinMelikMerkumians, line 264: I guess you could directly return the result of the if condition.
However I would recommend to add a name for the -1 magic number as a constexpr so the semantic meaning of mReadCounter = -1 is clear

modules/wagokbus/WagoDeviceController.cpp
```diff
@@ -225,13 +225,18 @@ namespace forte::eclipse4diac::io::wago {
...
       mAppDevInterface->ReadStart(mKBusDeviceId, mTaskId);
+      mReadCounter += 1;
       mAppDevInterface->ReadBytes(mKBusDeviceId, mTaskId, mRegComDevice->getOffset_REG_S0() / 8, 1, &lREG_S);
       mAppDevInterface->ReadEnd(mKBusDeviceId, mTaskId);
       lRegComDevice = mRegComDevice;
       if (!isRegComOn && ((lREG_S & 0x80) == 0x80)) {
         isRegComOn = true;
         triggerEvent = true;
         DEVLOG_DEBUG("[WagoDeviceController] Register communication on.\n");
+      } else if (mReadCounter > 50) { // abort in case of C0.7 fail
```
MartinMelikMerkumians, line 236: what ia a C0.7 fail? 😅

MartinMelikMerkumians (changes requested):

modules/wagokbus/WagoDeviceController.cpp
moniwe: it means writing on Control Byte 0 Bit 7 failed ... not the best comment probably 🐱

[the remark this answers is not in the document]
moniwe: probably true 🙈

Issue #1054, opened 2026-09-17
Memory leak in OPCUA_MGR::initialize() (34 allocations, 385 bytes, every -d OPCUA_DEV startup)


franz-hoepfinger-4diac:
## Summary
Starting FORTE with the `-d OPCUA_DEV` device (see `doc/tutorials/opcuadeployment.adoc`) leaks 385 bytes across 34 allocations, every single time, confirmed via AddressSanitizer/LeakSanitizer. The leak originates from `OPCUA_MGR::initialize()` (called once, from `OPCUA_DEV::startDevice()`), via two sub-paths that both end up copying a `UA_NodeId`/`UA_String` into open62541's node store without that copy ever being freed:

- `OPCUA_MGR::addMethodNode()` — called once per management method OPCUA_MGR registers (`writeDevice`, `startDevice`, `stopDevice`, `resetDevice`, `killDevice`, per-resource/FB CRUD methods, etc.) — 31 of the 34 leaked allocations.
- `OPCUA_MGR::addMgmtObjectInstance()` — 3 of the 34 leaked allocations, for the management object instance node itself.

## Root cause (as far as traced)
Both paths eventually call into open62541's `addNode_raw()`/`Operation_addNode_begin()`, which internally does `UA_copy()` → `NodeId_copy()` → `String_copy()` → `UA_Array_copy()` to store its own copy of the requested `UA_NodeId`'s string component. That copy is what LeakSanitizer reports as never freed. I have not tracked down why this particular copy escapes cleanup (whether it's an OPCUA_MGR-side issue, e.g. never deleting these specific nodes before the server itself is torn down, or something open62541-side) -- flagging with the confirmed call sites and stack traces rather than guessing further.

## Reproduction
Build with `-DFORTE_COM_OPC_UA=ON -DFORTE_SYSTEM_TESTS=ON` and Address/LeakSanitizer enabled (e.g. via `-DCMAKE_CXX_FLAGS=-fsanitize=address`), then run any bootfile with `-d OPCUA_DEV`:

```
./forte_systemtest -d OPCUA_DEV -f <any .fboot>
```

LeakSanitizer reports the leak at process exit regardless of which bootfile is used, since the leak happens during `OPCUA_DEV::startDevice()`, before any bootfile is even loaded.

## Evidence
[...]

PR #1055, opened 2026-09-18
Fix OPCUA_MGR NodeId memory leak

franz-hoepfinger-4diac:
## Summary

`UA_Server_addMethodNode()`/`addObjectNode()`/`addObjectTypeNode()` write the newly assigned NodeId into an `outNewNodeId` output parameter that the caller owns and must release; `OPCUA_MGR` never did, leaking the heap-allocated string inside each returned NodeId. This hit `addMethodNode()` once per management method `OPCUA_MGR` registers (exercised via the `-d OPCUA_DEV` device), plus the `mMgmtNodeId` member.

Fixes #1054.

## Fix

- Clears the local NodeId in `addMethodNode()` right after use, and clears `mMgmtTypeId`/`mMgmtNodeId`/`mResourceTypeId` (now also NULL-initialized in the constructor) in the destructor.
- The two type-definition functions pass a separate, non-owning local as the requested NodeId instead of the member itself, so a failed add leaves the member at its safe NULL default rather than a borrowed pointer into static storage that the destructor would then try to free.

Confirmed via LeakSanitizer: 385 bytes across 34 allocations before this change, zero after.

## Test plan

- [x] Existing opc_ua systemtest suite passes unchanged
- [x] LeakSanitizer run with `-d OPCUA_DEV` confirms zero leaks after the fix (385 bytes / 34 allocations before)


franz-hoepfinger-4diac:
Code Reviews in Downstream Repo: 

- https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/59

Issue #1053, opened 2026-09-17
COPC_UA_Local_Handler::startServer() uses isAlive() as a readiness signal, causing permanent NO_SOCKET for local OPC UA actions that INIT during server startup


franz-hoepfinger-4diac:
## Summary
`COPC_UA_Local_Handler::startServer()` (`com/opc_ua/src/opcua_local_handler.cpp`) has a check-then-act race: function blocks whose local OPC UA action `INIT`s concurrently with the very first call that spawns the server thread can permanently fail to ever create/access their OPC UA node — with no retry, for the entire runtime session.

## Root cause
```cpp
void COPC_UA_Local_Handler::startServer() {
  if (!isAlive()) {
    start();
    mServerStarted.waitIndefinitely();
    mServerStarted.inc();
  }
}
```
`isAlive()` becomes `true` as soon as the background thread (`run()`) is spawned — well before `run()` reaches `mUaServer = UA_Server_newWithConfig(&config);` and eventually signals readiness via `mServerStarted.inc()`.

Any concurrent caller of `startServer()` (via `initializeAction()`, called from a different FB's `INIT`) that arrives while the thread is already alive but hasn't finished constructing `mUaServer` sees `isAlive() == true`, skips the entire `if` block — including the wait — and returns immediately. `initializeAction()` then checks `if (mUaServer)` (still `nullptr`) and returns `UA_STATUSCODE_BADINTERNALERROR`. This propagates through `COPC_UA_Layer::openConnection()` as `e_InitTerminated`, and the owning communication FB (e.g. `PUBLISH_1`) permanently reports `STATUS = "NO_SOCKET"` — `openConnection()` is only attempted once per FB lifetime, so nothing ever retries it.

The concurrency itself is architecturally real: each unconnected `EInit` input auto-fires once at resource start as an independent event chain (per FORTE semantics), and independent event chains are dispatched to separate `CEventChainExecutionThread` instances (`core/include/forte/ecet.h`), i.e. genuine separate OS threads (`arch::CThread`). A resource containing several function blocks that each perform a local OPC UA WRITE/PUBLISH action with an unconnected `INIT` will therefore call `initializeAction()`/`startServer()` from multiple real threads concurrently at boot.

## Impact
[...]

PR #1056, opened 2026-09-18
Add regression test for OPC UA startServer() race

franz-hoepfinger-4diac:
## Summary

`COPC_UA_Local_Handler::startServer()` uses `isAlive()` as its readiness signal, but `isAlive()` turns true the instant the background `run()` thread is scheduled, well before the real server startup finishes. Any caller that arrives during that window gets a permanently null `mUaServer`, with no retry.

Relates to #1053.

## Test

Adds `OPCUA_LocalHandlerStartServerRace`: starts 16 local `PUBLISH_1` instances under the `-d OPCUA_DEV` device and checks that all end up with `STATUS=="OK"`. All 16 fire off the same delayed start event, so none of them can race ahead of the device thread's own eager `startServer()` call and accidentally pass on an unfixed implementation. A test-only hook (compiled out of non-test builds) lets `run()` wait on a deterministic attempt counter instead of a fixed delay, so the outcome does not depend on host speed.

This PR is deliberately test-only and currently **red** against `develop`. The fix follows as a separate, stacked PR (opened shortly, will be linked here once it exists) so the two can be reviewed independently: this one proves the bug, the other proves the fix.

## Test plan

- [x] `ctest -R OPCUA_LocalHandlerStartServerRace` fails reliably (16/16) against current `develop`
- [x] Rest of the opc_ua systemtest suite unaffected


franz-hoepfinger-4diac:
code review in downstream Repo: 

- https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/57

systemtests/src/modules/opc_ua/CMakeLists.txt
```diff
@@ -19,7 +19,59 @@ SET_TESTS_PROPERTIES(OPCUA_LocalFirstRead PROPERTIES FAIL_REGULAR_EXPRESSION "ER
...
+                      ENVIRONMENT "FORTE_COM_OPC_UA_TEST_STARTUP_DELAY_MS=500;FORTE_COM_OPC_UA_TEST_STARTUP_WAIT_FOR_ATTEMPTS=16;LSAN_OPTIONS=suppressions=${CMAKE_CURRENT_SOURCE_DIR}/opcua_dev_known_leak.lsan-suppressions;ASAN_OPTIONS=fast_unwind_on_malloc=0:malloc_context_size=30")
+# Deliberately RED on this branch alone (see the fix in the stacked PR). WILL_FAIL (the pattern
+# Test_Hard_Fail_Alaways_Inverted/Test_Soft_Fail_Alaways_Inverted use above) only inverts a
+# nonzero *exit code* -- it has no effect here, since this failure is detected via
+# FAIL_REGULAR_EXPRESSION matching TEST_CONDITION_FAILED in the output while the process itself
+# still exits 0, and CTest treats a FAIL_REGULAR_EXPRESSION match as a failure unconditionally.
+# DISABLED keeps develop's CI green if this PR is ever merged before its stacked fix PR instead:
+# the test is skipped rather than run-and-failed. Removed by the stacked fix PR, where this test
+# is expected to actually run and pass.
+SET_TESTS_PROPERTIES(OPCUA_LocalHandlerStartServerRace PROPERTIES DISABLED TRUE)
```
franz-hoepfinger-4diac, line 74: the PR itself is Green, because this Test is DISABLED TRUE

azoitl:
Please do not submitt regression tests as own PRs. This is just generates stupid noise and adds no value.

franz-ms-muc:
> Please do not submitt regression tests as own PRs. This is just generates stupid noise and adds no value.

I will not do again. I can post them on my repo and not push here.

PR #1057, opened 2026-09-18
Fix OPC UA startServer() readiness check

franz-hoepfinger-4diac:
## Summary

`isAlive()` becomes true as soon as the background server thread is scheduled, not once the server is actually ready, so a caller arriving during real startup skipped the intended wait and saw a permanently null `mUaServer`.

Fixes #1053. Stacked on #1056 (the regression test) — until #1056 merges, this PR's diff will show both commits; only the second one is the actual fix.

## Fix

- Tracks readiness explicitly with a single `mServerReady` flag, set only once `run()` actually finishes successfully, and a dedicated mutex serializing the whole start attempt end to end so concurrent callers can't race on `isAlive()` at all.
- Left `false` on a failed attempt so the next caller retries, and reset by `stopServer()` so a later restart doesn't skip `start()` and reuse an already-deleted `mUaServer`.
- A failed attempt's thread signals readiness from inside `run()`, before the thread base clears its handle, so an immediate retry could see the old thread as still running and skip spawning a new one; fixed by joining the failed thread before returning.
- A successful run signals readiness twice, once when ready and again, unconsumed, on its own shutdown, leaving that second signal stuck for the next `startServer()` call to consume instead of waiting for its own thread; fixed by only signaling once per `run()`.
- `CThreadBase::join()`'s own "allow many joins" re-post means `join()` can return immediately on a retry, consuming a stale post from an earlier failed attempt instead of actually waiting for the current thread to unwind; fixed by waiting on the thread handle directly instead of trusting `join()` across repeated failure cycles on the same object.

## Test plan

- [x] Turns #1056's regression test green
- [x] Rest of the opc_ua systemtest suite (18/18) still passes


franz-hoepfinger-4diac:
Code-Reviews: in downstream Repo: 

- https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/4diac-forte/pull/58

systemtests/src/modules/opc_ua/CMakeLists.txt
```diff
@@ -19,7 +19,50 @@ SET_TESTS_PROPERTIES(OPCUA_LocalFirstRead PROPERTIES FAIL_REGULAR_EXPRESSION "ER
...
+# leak detection for the process, so a genuinely new leak is still caught. ASAN_OPTIONS below
+# forces slow/precise unwinding (the default fast unwinder only resolves 2 stack frames on this
+# build, too shallow for the suppression to match its FORTE-side frame).
+file(TO_NATIVE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/opcuaLocalHandlerStartServerRace.fboot" opcua_race_bootfile)
+add_test(NAME OPCUA_LocalHandlerStartServerRace
+         COMMAND $<TARGET_FILE:forte_systemtest> -d OPCUA_DEV -f "${opcua_race_bootfile}")
+set_tests_properties(OPCUA_LocalHandlerStartServerRace PROPERTIES
+                      TIMEOUT 10
+                      FAIL_REGULAR_EXPRESSION "TEST_CONDITION_FAILED|ERROR: T|==ERROR"
+                      ENVIRONMENT "FORTE_COM_OPC_UA_TEST_STARTUP_DELAY_MS=500;FORTE_COM_OPC_UA_TEST_STARTUP_WAIT_FOR_ATTEMPTS=16;LSAN_OPTIONS=suppressions=${CMAKE_CURRENT_SOURCE_DIR}/opcua_dev_known_leak.lsan-suppressions;ASAN_OPTIONS=fast_unwind_on_malloc=0:malloc_context_size=30")
```
franz-hoepfinger-4diac, line 28: see: 
https://github.com/eclipse-4diac/4diac-forte/pull/1056/changes#r4044761034 

Disabled true: the Test here runs really, and is green.

com/opc_ua/src/opcua_local_handler.h
```diff
@@ -187,6 +190,37 @@ namespace forte::com_infra::opc_ua {
...
 
+#ifdef FORTE_COM_OPC_UA_TEST_HOOKS
+      /**
+       * Test-only. Incremented once per initializeAction() entry, before enableHandler() is
+       * called. Used by run()'s test-only startup delay hook to deterministically wait until a
+       * configured number of attempts have been observed, instead of guessing a fixed
+       * wall-clock delay. See FORTE_COM_OPC_UA_TEST_STARTUP_DELAY_MS /
+       * FORTE_COM_OPC_UA_TEST_STARTUP_WAIT_FOR_ATTEMPTS in run().
+       */
+      std::atomic<unsigned int> mTestInitializeActionAttempts{0};
```
mx990, line 194: Does this really need to live inside the class? How about using a global or thread-local variable instead?

com/opc_ua/CMakeLists.txt
```diff
@@ -63,6 +63,17 @@ add_library(forte-com-opc_ua)
...
+# Only compiled in for systemtest builds -- gates test-only instrumentation (e.g. the artificial
+# startup delay hook in COPC_UA_Local_Handler::run(), used by OPCUA_LocalHandlerStartServerRace)
+# out of production/release builds entirely, rather than relying on the hook being a no-op at
+# runtime. Must be PUBLIC, not PRIVATE: opcua_local_handler.h is also included by other targets
+# that link against forte-com-opc_ua (e.g. forte-system's OPCUA_MGR.cpp, which needs the full
+# COPC_UA_Local_Handler definition). If those consumers were compiled without this definition
+# while forte-com-opc_ua's own sources have it, COPC_UA_Local_Handler would have two different
+# sizes/layouts across translation units -- a silent ODR violation that corrupts memory instead
+# of failing to build.
+target_compile_definitions(forte-com-opc_ua PUBLIC $<$<BOOL:${FORTE_SYSTEM_TESTS}>:FORTE_COM_OPC_UA_TEST_HOOKS>)
```
mx990, line 67: As already noted in the comment, definitions that influence memory layout can be quite problematic, often producing failures that are very difficult to diagnose. However, making it PUBLIC will propagate that definition beyond forte to user code that links against forte. If we must have this definition propagate to code depending on the OPC UA comlayer, then we should make a configure header instead to limit its scope as much as possible.

systemtests/src/modules/opc_ua/CMakeLists.txt
```diff
@@ -19,7 +19,50 @@ SET_TESTS_PROPERTIES(OPCUA_LocalFirstRead PROPERTIES FAIL_REGULAR_EXPRESSION "ER
...
+#
+# NOTE: using "-d OPCUA_DEV" here exposes a separate, pre-existing memory leak in
+# OPCUA_MGR::initialize() (confirmed present with or without this fix, see
+# https://github.com/eclipse-4diac/4diac-forte/issues/1054) that no other systemtest hits
+# because none of them use the OPCUA_DEV device. It is unrelated to this test's actual purpose
+# (verifying startServer()'s readiness handling), so it is suppressed narrowly, by function name,
+# via opcua_dev_known_leak.lsan-suppressions -- this only silences that specific known leak, not
+# leak detection for the process, so a genuinely new leak is still caught. ASAN_OPTIONS below
+# forces slow/precise unwinding (the default fast unwinder only resolves 2 stack frames on this
+# build, too shallow for the suppression to match its FORTE-side frame).
```
mx990: Please avoid adding walls of text as comments. If one really wants the full story, they can get the commit via git blame and read the commit message.

mx990 (changes requested):
I appreciate the added regression test, but this is currently very intrusive into the tested code. I believe we should try harder to find a better way to test this before going forward with the current solution.

franz-hoepfinger-4diac:
> I appreciate the added regression test, but this is currently very intrusive into the tested code. I believe we should try harder to find a better way to test this before going forward with the current solution.

my goal was to find a way to simulate the behaviour in linux, to have a failing test first, and a non-failing after the test. my main goal was not to add a Test for the test suite. So i totally agree: this can be improved.

systemtests/src/modules/opc_ua/CMakeLists.txt
franz-hoepfinger-4diac: already deleted.

franz-hoepfinger-4diac:
@mx990 BTW the Fix can live without the Test. should we split this (again) into 2 PRs for easier Discussion ???

azoitl:
I'm here with @mx990 there shouldn't be test specific ifdefs or code in the normal code. That sounds not only like a maintainance nightmare it also violates all test principles. However we should have some tests as this seems to be a critical issue and we want to avoid that its coming back. Especially as a major rework of OPC UA is on our roadmap.

franz-hoepfinger-4diac:
> I'm here with @mx990 there shouldn't be test specific ifdefs or code in the normal code. That sounds not only like a maintainance nightmare it also violates all test principles. However we should have some tests as this seems to be a critical issue and we want to avoid that its coming back. Especially as a major rework of OPC UA is on our roadmap.

as i have absolutely no idea how to make such a Test, i suggest, i split the Test and the fix again, so we can merge the Fix, close the issue.

To come forward with the Test, i would ask the simulation guys at your university. in the CI/CD Pipeline it is possible to run a QEMU Simulation, and this can be -- according to a fast googling -- reduce the CPU Speed of that target and so avoid hacking the code with a injected delay -- which is also not really the same as running on a slower CPU. 

I guess such a Simulaiton would bring other benefits too, isn't it ?

azoitl:
> as i have absolutely no idea how to make such a Test, i suggest, i split the Test and the fix again, so we can merge the Fix, close the issue.
> 
> To come forward with the Test, i would ask the simulation guys at your university. in the CI/CD Pipeline it is possible to run a QEMU Simulation, and this can be -- according to a fast googling -- reduce the CPU Speed of that target and so avoid hacking the code with a injected delay -- which is also not really the same as running on a slower CPU.
> 
> I guess such a Simulaiton would bring other benefits too, isn't it ?

This sounds like an extremely fragile and heavy maintenance setup. Yes it may be interesting.  But it does not sound like something I would like to have in our day to day CI/Cd setup.

franz-ms-muc:
I guess such thing must be widely researched. How do other projects do this. And then tested on a Test repo; not on Forte.

mx990:
If I understand your test correctly, you just need to delay the startup of the OPC UA server to emulate a slower system? 

Since `COPC_UA_Local_Handler::run()` initializes the nodesets before starting the server, you can register your own dummy nodeset hook that simply waits for the desired time before returning, see below for an example on how to register a nodeset hook:
https://github.com/eclipse-4diac/4diac-forte/blob/35710446753e4b0a6df1a55f58a6ee4c9b7de346/systemtests/src/modules/opc_ua/namespace/fordiacNamespace.cpp#L19

You then need to include this dummy nodeset hook in the systemtest executable and set the desired time, possibly via a command line argument, see below for an example on how to register a command line option:
https://github.com/eclipse-4diac/4diac-forte/blob/35710446753e4b0a6df1a55f58a6ee4c9b7de346/com/opc_ua/src/opcua_local_handler.cpp#L52-L71

